PR #10, opened 2022-03-22
Use on-disk recording ID to create recordings if it is a UUID

bhackett1024:
For https://github.com/RecordReplay/backend/issues/5028, only use a single ID if the one created by the driver is a UUID.

### DIFF
--- a/packages/replay/package.json
+++ b/packages/replay/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@replayio/replay",
-  "version": "0.1.0-alpha.7",
+  "version": "0.1.0-alpha.8",
   "description": "CLI tool for uploading and managing recordings",
   "bin": {
     "replay": "bin/replay"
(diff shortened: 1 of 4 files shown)


packages/replay/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@replayio/replay",
-  "version": "0.1.0-alpha.7",
+  "version": "0.1.0-alpha.8",
```
jcmorrow, line 3: If you incremented this by hand it might be better to leave it and let `lerna` do its thing when it publishes (but I could be wrong).

jcmorrow (approved):
One comment about package.json, but that's not a crucial point. I think the worst possible outcome there is we skip a pre-release number.

bhackett1024:
I don't know what lerna is.

jcmorrow:
Yeah, you can probably just leave the version number alone, we're using this [lerna](https://lerna.js.org/) tool to help with monorepo management and npm packaging and it will do some version bumping when we tell it to publish.

bhackett1024:
I don't know how to release the new version of the CLI, @ryanjduffy or @jcmorrow can you take care of this?

PR #12, opened 2022-04-01
Add bin to support installing browsers

ryanjduffy:
Adds support for `PLAYWRIGHT_SKIP_BROWSER_INSTALL` env variable and `npx @replayio/playwright install` command to install browsers


ryanjduffy:
I've published this on npm as `0.1.1-alpha.2` if you want to test it

jcmorrow (approved):
OK, so I'm all good with this. I just want to recap what I just chatted through with @ryanjduffy . I was a little confused because in past I have used `PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD` to mean "don't install the playwright version of FF, I want to install the Replay version instead". Now, setting `PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD` will mean "don't install the playwright version of any browsers, and also don't automatically install the Replay version of any browsers". To Ryan's point, this makes sense because a lot of CI configs for OSS projects do this exact thing (and then manually install the browsers they want to use), so we are adding support for this env var to make it feel like `replay-cli` "just works" like playwright as much as possible. 👍

PR #13, opened 2022-04-05
Fix installing firefox for playwright

ryanjduffy:
* Adds logging to the install process (unlocked by `opts.verbose` as elsewhere)
* Renames the key for installing the replay firefox browser from "gecko" to "firefox"


jcmorrow (approved):
LGTM

PR #22, opened 2022-04-27
Add puppeteer package

ryanjduffy:
Migrates `puppeteer-config` to the cli and new npm namespace

Fixes https://github.com/RecordReplay/ops/issues/774


ryanjduffy:
@jcmorrow - tagged you on review for this to get some extra eyes. I know you haven't been in this code much so feel free to reach out with any questions!

jcmorrow (approved):
Well, you are right that I don't have a ton of context, and I've never used puppeteer, but it looks like a lot of familiar patterns came over from the playwright package, and that seems like a good approach - trying to keep these as consistent as possible for maintenance reasons. Nothing stands out to me as wrong, if there's anything particularly tricky you want to talk through I'm happy to chat, otherwise, merge away!

ryanjduffy:
It's pretty basic at this point by design. :)

PR #24, opened 2022-05-04
Upload extended recording metadata

ryanjduffy:
* Adds `@replayio/replay/metadata` submodule which exposes a method for each known metadata type to validate that the structure meets the minimum requirements for the current version.
  * For `test` version 1, this means that if the metadata is set, it must include a `result` and a `title`
* Validate the metadata with `superstruct` before starting the upload to avoid creating a recording and then failing on setting the metadata

I didn't document these methods as public API just yet because I want to limit the adoption until we've had a chance to use it internally (e.g. with #26)


packages/replay/metadata/test.js
```diff
@@ -0,0 +1,43 @@
...
+    title: string(),
+    result: enums(["passed", "failed", "timedOut"]),
+    path: optional(array(string())),
+    run: optional(string()),
+    file: optional(string()),
+  }),
+};
+
+function validate(metadata) {
+  return init(metadata && metadata.test);
```
jcmorrow, line 26: Maybe we should put a guard here rather than doing the `&&`? I could see that throwing some cryptic error on line 30 like `false does not have .version` or whatever, when the real problem is that `metadata` was falsey.

packages/replay/metadata/test.js
```diff
@@ -0,0 +1,43 @@
...
+const VERSION = 1;
+
+const versions = {
+  1: object({
+    version: defaulted(number(), () => 1),
+    title: string(),
+    result: enums(["passed", "failed", "timedOut"]),
+    path: optional(array(string())),
+    run: optional(string()),
+    file: optional(string()),
```
jcmorrow, line 21: This is a place where I actually think sorted keys might be nice, but I don' feel strongly about it.

packages/replay/metadata/index.js
```diff
@@ -0,0 +1,42 @@
...
+// Sanitizing arbitrary recording metadata before uploading by removing any
+// non-object values (allowing null) and limiting object values to known keys or
+// userspace keys prefixed by `x-`.
+function sanitize(metadata, opts = {}) {
+  const updated = {};
+  Object.keys(metadata).forEach((key) => {
+    const value = metadata[key];
+    // intentionally allowing `null` with the `typeof` check here
+    if (typeof value === "object") {
+      if (value === null || key.startsWith("x-")) {
```
jcmorrow, line 20: I'm not sure how I feel about the `x-` scheme. I have two questions: 
1) what are the trade-offs vs. allowing a particular key like `extras` and allowing whatever to be nested under that?
2) So people will be able to nest as much as they want under any key that starts with `x-`? What kind of use cases motivate us to have this right now? I'm all for making sure that we have custom metadata eventually, I just want to make sure we don't paint ourselves into a corner by choosing a particular schema before we know how people will actually use it and then having to support that for the rest of time.

packages/replay/metadata/index.js
```diff
@@ -0,0 +1,42 @@
...
+        updated[key] = value;
+      } else if (ALLOWED_KEYS.includes(key)) {
+        // validate known types
+        Object.assign(updated, handlers[key](metadata));
+      } else {
+        // and warn when dropping all other types
+        maybeLog(opts.verbose, `Ignoring metadata key "${key}". Custom metadata blocks must be prefixed by "x-". Try "x-${key}" instead.`);
+      }
+    } else {
+      maybeLog(opts.verbose, `Ignoring metadata key "${key}". Expected an object but received ${typeof value}`);
```
jcmorrow, line 31: I think for readability I would probably make this a guard at the top, but don't feel too strongly about it.

jcmorrow (commented):
Couple of small thoughts, but overall looks good!

jcmorrow (approved):

packages/replay/metadata/index.js
ryanjduffy: 1. The idea here was to reserve space for us to grow custom metadata in the future without stepping on userspace fields. Since this isn't public API yet, we don't expect to see much if any of this but it seems useful to carve out a space for custom metadata now.
2. That's fair. I'm not sure how people will use it but I'm also not sure how we'll figure it out if we don't provide a place. Again, this isn't public yet so we can walk it back a bit if we find reasons to do so.

PR #26, opened 2022-05-04
Use extended metadata for playwright/test

ryanjduffy:
Adopts the changes from #24 to add the proper extended metadata for tests for recordings.

* Use `@replayio/replay/metadata` to create the test metadata structure
* Move `testId` into an `x-playwright` custom block to avoid the CLI complaining about invalid custom metadata fields
* Updates the `test` metadata structure to use `run` as an object with an `id` that is a unique UUID and `title` that is a optionally user-defined string.
* Adds support for passing additional metadata (e.g. the git source for the test execution) via the existing `RECORD_REPLAY_METADATA` env variable or the new `PLAYWRIGHT_REPLAY_METADATA` or via `playwright.config.ts` config
* Adds support for configuring the run title either via `PLAYWRIGHT_REPLAY_RUN_TITLE` or via the config

### Adds some git metadata to every recording
```
PLAYWRIGHT_REPLAY_METADATA='{"git": {"repo": "https://github.com/RecordReplay/replay-cli"}}' npx playwright test --project replay-firefox --reporter @replayio/playwright/reporter,line
```
Using our "standard" env variable
```
RECORD_REPLAY_METADATA='{"git": {"repo": "https://github.com/RecordReplay/replay-cli"}}' npx playwright test --project replay-firefox --reporter @replayio/playwright/reporter,line
```
Using both variables. This'll warn and only use `PLAYWRIGHT_REPLAY_METADATA` as the most specific field
```
PLAYWRIGHT_REPLAY_METADATA='{"git": {"repo": "https://github.com/RecordReplay/replay-cli"}}' RECORD_REPLAY_METADATA='{"not-used": {"repo": "https://github.com/RecordReplay/replay-cli"}}' npx playwright test --project replay-firefox --reporter @replayio/playwright/reporter,line
```

Using the config

```ts
// playwright.config.ts
import { PlaywrightTestConfig } from "@playwright/test";
import { devices as replayDevices } from "@replayio/playwright";

const config: PlaywrightTestConfig = {
  reporter: [
    ["line"],
    ["@replayio/playwright/reporter", { runTitle: "hey", metadata: { git: { repo: "okay" } } }],
  ],
 // ... etc ...
}
```

### Set the run title

Using env variable
```
PLAYWRIGHT_REPLAY_RUN_TITLE="my custom run" npx playwright test --project replay-firefox --reporter @replayio/playwright/reporter,line
```

Using the config

```ts
// playwright.config.ts
import { PlaywrightTestConfig } from "@playwright/test";
import { devices as replayDevices } from "@replayio/playwright";

const config: PlaywrightTestConfig = {
  reporter: [
    ["line"],
    ["@replayio/playwright/reporter", { runTitle: "hey" }],
  ],
 // ... etc ...
}
```


packages/playwright/types/replayio__replay__metadata.d.ts
```diff
@@ -0,0 +1,26 @@
...
+  export interface TestMetadata {
+    version: 1,
+    result: "passed" | "failed" | "timedOut";
+    // the "path" to the test as a list of the branches of the describe/it tree
+    path?: string[];
+    // title of the current test
+    title: string;
+    // the file containing the test ran
+    file?: string;
+    // a unique identifier for the test run (e.g. the commit + the current time?)
```
jcmorrow, line 20: If running from GH actions probably the `github.run_id` would be best https://docs.github.com/en/actions/learn-github-actions/contexts#github-context

packages/playwright/src/reporter.ts
```diff
@@ -44,19 +109,39 @@ class ReplayReporter implements Reporter {
...
+              title: test.title,
+              result: status,
+              path: test.titlePath(),
+              run: {
+                id: this.baseId,
+                title: this.runTitle
+              },
+              // extract the relative path from titlePath() but fall back to the
+              // full path
+              file: test.titlePath()[2] || test.location.file,
```
jcmorrow, line 143: what is the shape of `titlePath()` here usually? Is it possible that we'll blow up trying to pull [2] from `undefined` or anything like that?

packages/playwright/src/reporter.ts
```diff
@@ -44,19 +109,39 @@ class ReplayReporter implements Reporter {
   }
 
   onTestEnd(test: TestCase, result: TestResult) {
-    if (!["passed", "failed"].includes(result.status)) return;
+    const status = result.status;
+    if (status !== "passed" && status !== "failed") return;
```
jcmorrow, line 113: What would this indicate?

jcmorrow (approved):
I'm not going to pretend to understand what every line of this does, but I called out the things that seemed slightly 🐟 y to me. I leave it to your discretion as to what should actually block merge. Also - your PR body is great in terms of examples. How would you feel about also adding that somewhere in a README (not sure if it should be in this repo or some related repo).

packages/playwright/src/reporter.ts
ryanjduffy: This reporter would still be called for a skipped test but there wouldn't be a recording so there's no work to do so we're bailing early.

ryanjduffy: Thinking more though, we do support the `timedOut` status so this should be `if (status === "skipped") return` which  is more clear anyway.

ryanjduffy:
I'm holding off on documenting all of this stuff in the readme and the docs until it's out the door and I've had a chance to try it out with a couple test projects. I did document them here as a shortcut to future doc updates :)

packages/playwright/src/reporter.ts
jcmorrow: that is much clearer 🏅

PR #27, opened 2022-05-04
Convert replay CLI to typescript

ryanjduffy:


packages/replay/metadata/index.ts
```diff
@@ -1,18 +1,28 @@
...
 // Each known metadata block should have a sanitizer that will check the contents before the upload
 const handlers = {
   test: test.validate
 };
 
+type AllowedKey = keyof typeof handlers;
 const ALLOWED_KEYS = Object.keys(handlers);
 
+function isAllowedKey(key: string): key is AllowedKey {
+  if (ALLOWED_KEYS.includes(key)) {
```
markerikson, line 14: :hand: tiny nitpick: could just be `return ALLOWED_KEYS.includes(key)` :)

packages/replay/metadata/index.ts
```diff
@@ -24,9 +34,9 @@ function sanitize(metadata, opts = {}) {
     if (value === null || key.startsWith("x-")) {
       // passthrough null or userspace types
       updated[key] = value;
-    } else if (ALLOWED_KEYS.includes(key)) {
+    } else if (isAllowedKey(key)) {
       // validate known types
-      Object.assign(updated, handlers[key](metadata));
+      Object.assign(updated, handlers[key](metadata as any));
```
markerikson, line 39: :speech_balloon: Probably a better way to do it here, but this is not safety-critical code, so YOLO

packages/replay/metadata/test.ts
```diff
@@ -1,3 +1,5 @@
+import { Struct } from "superstruct";
```
markerikson, line 1: :hand: can probably use `import type { Struct }` here. Not critical, but nice to delineate what's a type vs what's a value

packages/replay/metadata/test.ts
```diff
@@ -13,30 +15,30 @@ const isUuid = require("is-uuid");
...
   if (!metadata || !metadata.test) {
     throw new Error('Test metadata does not exist');
   }
 
   return init(metadata.test);
 }
 
-function init(data) {
-  const version = data.version || VERSION;
+function init(data: Record<string, unknown>) {
```
markerikson, line 40: :hand: given that we have `Record<string, unknown>` popping up quite a bit, I'll often extract that into its own type just so I don't repeat myself

packages/replay/src/upload.ts
```diff
@@ -1,20 +1,26 @@
...
+) {
   if (!gClient) {
     let { resolve } = gClientReady;
     gClient = new ProtocolClient(
       server,
       {
         async onOpen() {
           try {
-            await gClient.setAccessToken(accessToken);
+            await gClient?.setAccessToken(accessToken);
```
markerikson, line 23: :hand: We should be pretty sure it exists on this line, right? Looks like `!` would be more appropriate here

markerikson (approved):
Tiny nitpicks, but LGTM!

packages/replay/metadata/index.ts
ryanjduffy: Agreed. I spent a little time trying to figure out the right type but couldn't and I bailed on it.

ryanjduffy:
Fixed up a few things based on @markerikson 's comments in c1b2f60 and then realized there were some ts issues when you bootstrapped lerna so I also added 3cee941 to get everything working so you can:

* `npm run bootstrap` to install all deps and cross-link packages
* `npm run typecheck` to run `tsc` for each bundle`
* and run `npm link` in the `dist/` folder of each package to use the local copy for testing

😓

PR #31, opened 2022-05-17
Add metadata api method

ryanjduffy:
Refactor the logic to append recording metadata to a first class API so other systems (e.g. GH actions) can augment our metadata.


ryanjduffy:
This unlocks adding `source` metadata from `action-upload` without hacking it in like we've done in the playwright reporter.

jaril (approved):

PR #37, opened 2022-06-03
Fix playwright reporter to set metadata on retry

ryanjduffy:
## Issue

When playwright retries a failed test, it starts a new worker and increments the worker index. As a result, the reporter wasn't writing the metadata so the replay was missing that data

## Resolution

Rather than pre-creating empty metadata files based on the number of workers, just create the metadata files as needed when the test begins. By not checking for an existing file, we always create the correct file regardless of the worker index.


jaril (approved):

ryanjduffy:
@jaril This is published as 0.2.9. It should be automatically picked up by the action but let me know if that doesn't appear to be the case.

PR #40, opened 2022-06-09
Add cypress module

ryanjduffy:
Current Issues:
* Unlike playwright, we can't set environment variables for the browser so I haven't been able to automate setting `RECORD_ALL_CONTENT` or `RECORD_REPLAY_METADATA_FILE`
  * Added docs to the readme to address this for v0.1.0
* cypress (it seems) runs all of the tests in the same browser so we don't get individual recordings. That _might_ be okay for users but it doesn't align well with our test suites dashboard.
* Also because the browser is reused, we lack a way to set the right metadata for the recording since it doesn't map to any one test file or test
  * Treating a "spec" as a single test means at least that we get individual recordings for each of those and then can generate metadata for the spec. It's not 1:1 for playwright tests but I think will still work with our test suites ui.
* In the replay browser, the UI is offset vertically by ~200px so the top of the screen is cut off and there's a white bar at the bottom
  * this went away in my last batch of recordings ... not sure why
* Added docs for CYPRESS_NO_COMMAND_LOG but it's no supported in v10 yet (https://github.com/cypress-io/cypress/issues/21721)

### DIFF
--- a/packages/cypress/src/index.ts
+++ b/packages/cypress/src/index.ts
@@ -3,8 +3,15 @@
 import path from "path";
 import { getPlaywrightBrowserPath } from "@replayio/replay";
 import { getDirectory } from "@replayio/replay/src/utils";
+import ReplayReporter from "./reporter";
 
 const plugin: Cypress.PluginConfig = (on, config) => {
+
+  const reporter = new ReplayReporter(getMetadataFilePath());
+  on("before:browser:launch", (browser) => reporter.onBegin(browser.family));
+  on("before:spec", spec => reporter.onTestBegin(spec));
+  on("after:spec", (spec, result) => reporter.onTestEnd(spec, result));
+
   const chromiumPath = getPlaywrightBrowserPath("chromium");
   if (chromiumPath) {
     Object.assign(config, {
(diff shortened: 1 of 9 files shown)


jasonLaster:
> In the replay browser, the UI is offset vertically by ~200px so the top of the screen is cut off and there's a white bar at the bottom

This feels like the classic follow up item

packages/cypress/src/index.ts
```diff
@@ -0,0 +1,59 @@
+/// <reference types="cypress" />
+
+import path from "path";
+import { getPlaywrightBrowserPath } from "@replayio/replay";
+import { getDirectory } from "@replayio/replay/src/utils";
+import ReplayReporter from "./reporter";
+
+const plugin: Cypress.PluginConfig = (on, config) => {
+  console.log("!!!!!", getMetadataFilePath());
```
jasonLaster, line 9: 🧂

jasonLaster (approved):
Love how clean this is

packages/cypress/src/index.ts
ryanjduffy: IT'S STILL A DRAFT PR, JASON! ;)

ryanjduffy:
Opened <https://github.com/cypress-io/cypress/pull/22250> to see about adding support for browser instance-specific environment variables like we have for playwright. That'll eliminate the requirement for the user to config `RECORD_ALL_CONTENT` and `RECORD_REPLAY_METADATA_FILE` before running their tests.

PR #42, opened 2022-06-16
Allow setting test run id via environment variable

ryanjduffy:


packages/cypress/src/reporter.ts
```diff
@@ -47,19 +46,10 @@ class ReplayReporter {
       process.env.RECORD_REPLAY_METADATA ||
       null;
     if (baseMetadata) {
-      // Since we support either a string in an environment variable or an
```
ryanjduffy, line 50: This block was copied from playwright but isn't relevant because there isn't a config file in which the base metadata can be set.

jaril (approved):

PR #44, opened 2022-06-21
Fix install script for cypress

ryanjduffy:
## Issue

`@replayio/cypress` doesn't install the replay browsers

## Analysis

The code for this was copied from a much earlier version of the plugins which included an install script but it wasn't as robust. Also, the `package.json` didn't include the `install` hook so the existing script wasn't called.

## Resolution

* Update the script to match the newer version
* Add support for `CYPRESS_INSTALL_BINARY` to suppress install (like `PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD`)
* Add `install` script to `package.json`

### DIFF
--- a/packages/cypress/bin/replayio-cypress.ts
+++ b/packages/cypress/bin/replayio-cypress.ts
@@ -1,0 +1,42 @@
+#!/usr/bin/env node
+
+import install from "../src/install";
+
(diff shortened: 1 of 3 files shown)


ceceliacreates (approved):
Looks good, once merged I'll test by deleting my local .replay and confirming it installs from the package.

PR #48, opened 2022-06-29
Log the metadata we are setting

jcmorrow:
These logs can be helpful to when viewing task output directly, and they
would be even *more* helpful if they would tell you which metadata is
associated with what recording upload.

### DIFF
--- a/packages/replay/src/main.ts
+++ b/packages/replay/src/main.ts
@@ -348,7 +348,7 @@ async function doUploadRecording(
   );
   maybeLog(verbose, `Created remote recording ${recordingId}, uploading...`);
   if (metadata) {
-    maybeLog(verbose, `Setting recording metadata for ${recordingId}`);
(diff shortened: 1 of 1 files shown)


ryanjduffy (approved):

packages/replay/src/main.ts
```diff
@@ -348,7 +348,7 @@ async function doUploadRecording(
   );
   maybeLog(verbose, `Created remote recording ${recordingId}, uploading...`);
   if (metadata) {
-    maybeLog(verbose, `Setting recording metadata for ${recordingId}`);
+    maybeLog(verbose, `Setting recording metadata for ${recordingId}: ${metadata}`);
```
jasonLaster, line 351: just want to clarify that metadata is a string

jcmorrow: Ya know what, this should probably have a JSON stringify in there. I'll submit a patch on top

PR #53, opened 2022-07-12
Only upload finished statuses (onDisk, crashed)

jaril:
Fixes FE-410.

### DIFF
--- a/packages/replay/src/bin.ts
+++ b/packages/replay/src/bin.ts
@@ -20,6 +20,7 @@ program
     "--directory <dir>",
     "Alternate recording directory."
   )
+  .option('-a, --all', 'Include all recordings')
   .action(commandListAllRecordings);
 
 program
@@ -37,7 +38,6 @@ program
     "--api-key <key>",
     "Authentication API Key"
   )
-  .option('--include-in-progress', 'Upload all recordings, including ones with an in progress status')
   .action(commandUploadRecording);
 
 program
--- a/packages/replay/src/bin.ts
+++ b/packages/replay/src/bin.ts
@@ -20,7 +20,6 @@ program
     "--directory <dir>",
     "Alternate recording directory."
   )
-  .option('-a, --all', 'Include all recordings')
   .action(commandListAllRecordings);
 
 program
@@ -38,6 +37,7 @@ program
     "--api-key <key>",
     "Authentication API Key"
   )
+  .option('--include-in-progress', 'Upload all recordings, including ones with an in progress status')
   .action(commandUploadRecording);
 
 program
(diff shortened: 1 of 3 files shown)


packages/replay/src/main.ts
```diff
@@ -243,14 +243,13 @@ function listAllRecordings(opts = {}) {
   return recordings.map(listRecording);
 }
 
-function uploadSkipReason(recording: RecordingEntry) {
+function uploadSkipReason(recording: RecordingEntry, includeInProgress: boolean = false) {
```
ryanjduffy, line 246: How are we exposing this to the CLI?

jaril: Added!

packages/replay/src/bin.ts
```diff
@@ -72,6 +73,7 @@ program
     "--api-key <key>",
     "Authentication API Key"
   )
+  .option('-a, --all', 'Upload all recordings')
```
ryanjduffy, line 76: This is a bit nitty but I think `npx replay upload-all --all` is awkward. Maybe `--include-in-progress` or `--include startedWrite,startedUpload` or something else?

ryanjduffy: or maybe `--status all|onDisk|crashed|startedWrite|startedUpload`?

jaril: I wasn't sure what the use case is for individually enumerating the status(es) the user might be filtering for, so kept it simple with  `--include-in-progress` for now

ryanjduffy: I'm not sure either. I don't want to change APIs so I was offered it as a flexible option. I'm okay with your choice for now and we can drop it in v2 if we decide to go a different route.

packages/replay/src/bin.ts
```diff
@@ -20,7 +20,6 @@ program
     "--directory <dir>",
     "Alternate recording directory."
   )
-  .option('-a, --all', 'Include all recordings')
```
ryanjduffy, line 23: I don't think you meant to remove this

packages/replay/src/bin.ts
```diff
@@ -38,6 +37,7 @@ program
     "--api-key <key>",
     "Authentication API Key"
   )
+  .option('--include-in-progress', 'Upload all recordings, including ones with an in progress status')
```
ryanjduffy, line 40: I don't think we need this here because the user has already chosen the recording to upload so the status isn't relevant.

ryanjduffy (changes requested):
Couple notes before we merge but we're close. Thanks for tackling this, @jaril !

jaril:
> Couple notes before we merge but we're close. Thanks for tackling this, @jaril !

Tried to do make the change in the simplest way possible but it was a bit tricky. This [implementation](https://gist.github.com/jaril/1fa50f4d3b2ea0198aed2ceb21e52d80) got a bit messier so settled with the simpler version here.

Behavior should be that:
- `upload` always includes all recordings
- `upload-all` takes the `--include-in-progress` option, and defaults to false

ryanjduffy (approved):

PR #61, opened 2022-07-15
replay: format ls in a human readable format

jazzdan:
The old `ls` dumped a lot of JSON to stdout, so much so that node could
actually fail to write it at all. Rather than try to chase down why that
was happening I decided to implement a human readable version of `ls`.

Now it looks like this:

```
ID                                    Status  Title                      Created At
d5f140d7-a6f4-4eca-9f36-77c39d411f03  onDisk  Replay of jest.js          2022-07-14T18:18:52.775Z
21060f63-a2c1-4dc5-a492-811508554867  onDisk  Replay of yarn-3.2.1.cjs   2022-07-14T18:18:51.131Z
cd443424-6a35-4d1e-a863-8aed960a2722  onDisk  Replay of yarn.js          2022-07-14T18:18:50.631Z
```

I also added some snapshot tests for this command, which required some
refactoring. In my experience snapshot tests are very useful for CLIs
that might be consumed by other programs so we can be aware of when we
might be breaking backwards compatibility.

In a future PR I will add a global `--json` flag to restore the old behavior.

Fixes BAC-1953.

### DIFF
--- a/packages/replay/src/main.ts
+++ b/packages/replay/src/main.ts
@@ -22,8 +22,15 @@ import {
 } from "./install";
 import { getDirectory, maybeLog } from "./utils";
 import { spawn } from "child_process";
-import { ListOptions, Options, RecordingEntry, UploadOptions } from "./types";
+import {
+  ListOptions,
+  Options,
+  ExternalRecordingEntry,
+  RecordingEntry,
+  UploadOptions,
+} from "./types";
 import { add } from "../metadata";
+import { generateDefaultTitle } from "./generateDefaultTitle";
 export type { BrowserName } from "./types";
 
 function getRecordingsFile(dir: string) {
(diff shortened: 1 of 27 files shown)


packages/replay/src/main.ts
```diff
@@ -245,9 +236,10 @@ function updateStatus(recording: RecordingEntry, status: RecordingEntry["status"
 }
 
 // Convert a recording into a format for listing.
-function listRecording(recording: RecordingEntry) {
+function listRecording(recording: RecordingEntry): ExternalRecordingEntry {
   // Remove properties we only use internally.
-  return { ...recording, buildId: undefined, crashData: undefined };
```
jazzdan, line 250: Heads up @jaril I think before the code wasn't actually removing `buildId` and `crashData`, just setting their values to undefined. I'm not enforcing with a type (`Omit<RecordingEntry, "buildId" | "crashData">`) that those keys are removed from the object altogether. Not sure if that's what you originally intended though.

ryanjduffy:
The metadata is also useful particularly for our test suites work so I'd like to find a way to retain that.

jazzdan:
@ryanjduffy ah so sounds like I shouldn't merge this before adding a `--json` flag to get all of the data?

packages/replay/src/main.ts
jaril: did you mean @jasonLaster ?

jazzdan: Whoops, I misread `git blame`. Thanks @jaril :)

ryanjduffy (approved):

PR #63, opened 2022-07-15
cypress: error if no replay browser could be found

jazzdan:
Not sure if this is a bug or not, I figured I would just put it up for discussion via a PR.

### DIFF
--- a/packages/cypress/src/index.ts
+++ b/packages/cypress/src/index.ts
@@ -26,6 +26,8 @@ const plugin: Cypress.PluginConfig = (on, config) => {
         isHeadless: false,
       }),
     });
+
+    return config;
   }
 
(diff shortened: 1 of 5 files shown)


packages/cypress/src/index.ts
```diff
@@ -26,6 +26,8 @@ const plugin: Cypress.PluginConfig = (on, config) => {
         isHeadless: false,
       }),
     });
+
+    return config;
```
ryanjduffy, line 30: We could find both on linux so we shouldn't return early.

ryanjduffy (approved):
I think this makes sense though tbh I'm not sure what is idiomatic for cypress plugins.

cc: @ceceliacreates

PR #66, opened 2022-07-19
Attempt to populate source metadata from know environment variables

ryanjduffy:
## Issue

Much of the potential source control metadata is readily available in CI environments from documented environment variables but the user has to wire them up to our structure. This is mostly handled by the GH actions but aren't for custom CI integrations.

## Resolution

* Attempt to populate values as much as possible from the environment
* Provide Replay-scoped overrides for all values to ease customization without stomping CI-provided variables


packages/replay/metadata/env.ts
```diff
@@ -0,0 +1,15 @@
+import { defaulted, string } from "superstruct";
+
+type Resolver = string | ((env: NodeJS.ProcessEnv) => string | undefined);
+
+const firstEnvValueOf =
```
jazzdan, line 5: This looks very unit testable 🤓

packages/replay/metadata/source.ts
```diff
@@ -1,35 +1,94 @@
...
 
+import { envString } from "./env";
 import { UnstructuredMetadata } from "./types";
 
+const defaultObject = (objStruct: any) => optional(defaulted(object(objStruct), {}));
+
 const VERSION = 1;
 
 const versions: Record<number, Struct> = {
   1: object({
```
jazzdan, line 12: Woah this is cool! I'm just kinda taking it at face value that this does what I expect, but looks good to me.

jazzdan (approved):
LGTM!

packages/replay/metadata/env.ts
ryanjduffy: good call. Added!

packages/replay/metadata/env.test.ts
```diff
@@ -0,0 +1,132 @@
...
+        KEY_2: "key 2",
+      };
+
+      const resolver = firstEnvValueOf(
+        env => env.KEY_1 && env.KEY_1_VALUE,
+        env => env.KEY_2 && env.KEY_2_VALUE
+      );
+      const result = resolver();
+
+      expect(result).toMatchSnapshot();
```
jazzdan, line 129: Should these be snapshot tests? In my mind a snapshot test is most useful when comparing lots and lots of output, but less useful when what is being compared is very small. I think we could rewrite these to be:

```typescript
expect(result).toEqual("value 2"");
```

ryanjduffy: Sure, we could do that. I don't have strong opinions here.

jazzdan: Sorry, I know I used snapshot tests for some of my stuff but my rationale was that the CLI output could get large and might change frequently. But yeah LGTM overall!

ryanjduffy: No worries

PR #68, opened 2022-07-21
Fix replay ls title generation

ryanjduffy:
## Issue

A default title is always generated even if there is one either from gecko or test metadata

## Resolution

Use the title if it's there


jazzdan (approved):
LGTM!

PR #69, opened 2022-07-23
Add recording routines package

bhackett1024:
https://linear.app/replay/issue/FE-440/add-recording-routines-package

This depends on https://github.com/replayio/backend/pull/6097, once that merges and the protocol gets updated then the ts-ignore bits in this PR can be removed.


ryanjduffy:
Noting that I saw this and will review tomorrow!

PR #71, opened 2022-07-25
Add filter option to ls and upload-all

ryanjduffy:
Adds `--filter` option to `ls` and `upload-all` to support a declarative way to filter the recordings

### DIFF
--- a/packages/replay/src/main.ts
+++ b/packages/replay/src/main.ts
@@ -22,9 +22,16 @@ import {
 } from "./install";
 import { getDirectory, maybeLog } from "./utils";
 import { spawn } from "child_process";
-import { ExternalRecordingEntry, ListOptions, Options, RecordingEntry } from "./types";
+import {
+  ExternalRecordingEntry,
+  FilterOptions,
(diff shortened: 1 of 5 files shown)


jaril (approved):

packages/replay/src/main.ts
```diff
@@ -229,6 +236,15 @@ function updateStatus(recording: RecordingEntry, status: RecordingEntry["status"
   recording.status = status;
 }
 
+function filterRecordings(recordings: RecordingEntry[], filter?: string) {
+  if (filter) {
+    const exp = jsonata(`$filter($, ${filter})[]`);
```
jaril, line 241: What was the reason for using jsonata here (instead of the regular array filter)??

jaril (approved):

packages/replay/src/main.ts
ryanjduffy: We need something that can be fed in via the CLI and jsonata enables that more easily than writing our own that translates to an array filter.

PR #72, opened 2022-07-25
Add cli option to generate metadata

ryanjduffy:
* Improves initialization for `test` metadata to pull from env variables like `source` does
* Adds `metadata` command to cli to generate and set metadata for recordings. Can be `--filter`-ed to limit which are updated. (so blocked by #71)

### DIFF
--- a/packages/replay/metadata/test.ts
+++ b/packages/replay/metadata/test.ts
@@ -1,4 +1,5 @@
 import type { Struct } from "superstruct";
+import { envString, firstEnvValueOf } from "./env";
 const {
   array,
   create,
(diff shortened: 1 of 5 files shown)


packages/replay/metadata/test.ts
```diff
@@ -18,16 +19,25 @@ const VERSION = 1;
...
-      })
+      defaulted(
+        object({
+          id: defaulted(
+            define("uuid", (v: any) => isUuid.v4(v)),
+            firstEnvValueOf("RECORD_REPLAY_METADATA_TEST_RUN_ID", "RECORD_REPLAY_TEST_RUN_ID")
+          ),
+          title: optional(envString("RECORD_REPLAY_METADATA_TEST_RUN_TITLE")),
+        }),
+        {}
```
jaril, line 37: What's the purpose of calling defaulted with an empty obj?

jaril (approved):
Should we close out https://github.com/replayio/replay-cli/pull/71 ?

packages/replay/metadata/test.ts
ryanjduffy: I couldn't find a way to prepopulate the values within the object from env variables without a default. So, without those defaults, if you called `test.init()` (with no value) you wouldn't get the commit id even if you had the env variable set because it wouldn't try to evaluate the rules within the obj.

ryanjduffy:
> Should we close out #71 ?

I'm going to rebase this when that merges so they can be self-contained PRs.

PR #80, opened 2022-08-24
Add test-utils package to consolidate shared test reporter logic

ryanjduffy:
* Adds `@replayio/test-utils` package to consolidate reporter logic. This'll make it easier for second/third parties to integrate other test runners and reduce duplication for us.
* Adds `pingTestMetrics` to test utils to allow us to track runtime perf.


ryanjduffy:
I've checked this out against jest, playwright, and cypress tests. Running with `RECORD_REPLAY_TEST_METRICS=1` and `RECORD_REPLAY_WEBHOOK_URL` set will opt into capturing test duration metrics.

jaril (approved):

PR #86, opened 2022-09-08
Add environment variables to cypress browser launch

ryanjduffy:
This is currently a no-op because cypress doesn't yet support `env` on `launchOptions` but will once https://github.com/cypress-io/cypress/pull/23624 lands.

I need to test more that it doesn't upset earlier versions of cypress before we land this though.

### DIFF
--- a/packages/cypress/package-lock.json
+++ b/packages/cypress/package-lock.json
@@ -1,6 +1,6 @@
 {
 	"name": "@replayio/cypress",
-	"version": "0.2.18",
+	"version": "0.0.0-experimental-c674b41b",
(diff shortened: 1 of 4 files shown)


ryanjduffy:
Looks like this does cause errors in earlier versions when you return unsupported values from the plugin callback. Will have to gate it on a cypress version check so we'll be blocked until this feature is released. :(

ryanjduffy:
Gated this on `v10.9` and verified the environment variables are not set in 10.8 and are set in 10.9

jaril (approved):

PR #89, opened 2022-09-28
Add debug log support to cli

ryanjduffy:
Adds `debug` to minimize normal logging with optional verbose debug logging via `DEBUG` environment variable


jaril (approved):

jaril:
Mind adding it to the docs? It'll take a while for it to stick so will probably have to reference it from somewhere 😅

PR #90, opened 2022-10-11
Add errors to test metadata

ryanjduffy:
* Adds `tests` to `test` metadata
* Surfaces errors from jest, playwright, and cypress in `tests`

### DIFF
--- a/packages/cypress/src/index.ts
+++ b/packages/cypress/src/index.ts
@@ -29,22 +29,23 @@ const plugin: Cypress.PluginConfig = (on, config) => {
   });
   on("before:spec", () => reporter.onTestBegin(undefined, getMetadataFilePath()));
   on("after:spec", (spec, result) => {
-    const status = result.tests.reduce<Test["result"]>((acc, t) => {
-      if (acc === "failed" || t.state === "failed") {
-        return "failed";
-      }
-
-      return "passed";
-    }, "passed");
-
-    if (!["passed", "failed"].includes(status)) return;
-
-    reporter.onTestEnd({
-      title: spec.relative,
-      path: ["", selectedBrowser || "", spec.relative, spec.specType || ""],
-      result: status,
-      relativePath: spec.relative,
+    const tests = result.tests.map<Test>(t => {
+      return {
+        title: t.title.pop() || spec.relative,
+        path: ["", selectedBrowser || "", spec.relative, spec.specType || ""],
+        result: t.state == "failed" ? "failed" : "passed",
+        relativePath: spec.relative,
+        error: t.displayError
+          ? {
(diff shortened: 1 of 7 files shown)


packages/replay/metadata/test.ts
```diff
@@ -10,27 +9,49 @@ const {
...
-const versions: Record<number, Struct> = {
+const versions: Record<number, Struct<any, any>> = {
   1: object({
     file: optional(envString("RECORD_REPLAY_METADATA_TEST_FILE")),
     path: optional(array(string())),
     result: defaulted(
       enums(["passed", "failed", "timedOut"]),
       firstEnvValueOf("RECORD_REPLAY_METADATA_TEST_RESULT")
     ),
+    steps: optional(
```
ryanjduffy, line 28: Hesitant to prescribe the format of this yet because we don't really know what we need. I think this is pretty safe minimum set though and maps cleanly from playwright

ryanjduffy:
Need to resolve the difference between multiple tests per recording from cypress and one test per recording in playwright. The current metadata structure doesn't handle that well.

packages/replay/metadata/test.ts
ryanjduffy: Skipping "steps" for now but we can bring it back later when we know how to populate it from cypress

packages/cypress/src/index.ts
```diff
@@ -29,22 +29,23 @@ const plugin: Cypress.PluginConfig = (on, config) => {
...
+      return {
+        title: t.title.pop() || spec.relative,
+        path: ["", selectedBrowser || "", spec.relative, spec.specType || ""],
+        result: t.state == "failed" ? "failed" : "passed",
+        relativePath: spec.relative,
+        error: t.displayError
+          ? {
+              // we don't get line/column from cypress yet but we may be able to
+              // derive it later once we're tracking the steps
+              message: t.displayError.substring(0, t.displayError.indexOf("\n")),
```
jaril, line 42: What else were you thinking that we might add to the error object?

packages/playwright/src/reporter.ts
```diff
@@ -48,13 +60,28 @@ class ReplayPlaywrightReporter implements Reporter {
...
+          id: this.getTestId(test),
+          title: test.title,
+          path: test.titlePath(),
+          result: status,
+          relativePath: test.titlePath()[2] || test.location.file,
+          error: errorMessage
+            ? {
+                message: errorMessage,
+                line: errorStep?.location?.line,
+                column: errorStep?.location?.column,
```
jaril, line 78: Ah there it is. Nevermind!

jaril (approved):

PR #96, opened 2022-11-04
Add asserts to steps

jaril:


packages/cypress/src/support.ts
```diff
@@ -3,7 +3,7 @@ import type { TestError } from "@replayio/test-utils";
 import { TASK_NAME } from "./constants";
 
 export interface StepEvent {
-  event: "step:enqueue" | "step:start" | "step:end" | "test:start" | "test:end";
+  event: "step:enqueue" | "step:start" | "step:end" | "test:start" | "test:end" | "log:added" | "log:changed";
```
ryanjduffy, line 6: I don't think these new keys are used. Right?

packages/cypress/src/support.ts
```diff
@@ -31,10 +31,8 @@ const makeEvent = (event: StepEvent["event"], cmd?: CommandLike, error?: TestErr
 });
 
 const handleCypressEvent = (event: StepEvent["event"], cmd?: CommandLike, error?: TestError) => {
-  if (cmd?.args[0] === TASK_NAME) return;
-
-  const arg = makeEvent(event, cmd, error);
-
+  if (cmd?.args?.[0] === TASK_NAME) return;
+  const arg = makeEvent(event, cmd, error)
```
ryanjduffy, line 35: This could be reverted as well since `args` is still required in `CommandLink`

ryanjduffy (approved):

packages/cypress/src/support.ts
jaril: NOPE!

PR #97, opened 2022-11-04
Fix build scripts to avoid error-warnings on node 14

ryanjduffy:
Switch the `if:exists` script to spawn a process rather than exit to avoid the case where node 14 logs an error (but then continues) if an intermediate step returns a non-zero exit code.


jaril (commented):
YES

jaril (approved):

PR #98, opened 2022-11-07
Skip checking that we're still in the right test for assert/log steps

jaril:
Logs operate on their own timeline, so it's possible for a log to come in after Cypress considers the test to have finished up.

The particular case we kept seeing is that Cypress treats the after each hook as its own test, separate from the actual running test. In such cases, it's possible for the assert log to register during that after each hook, in which case the (previous) `assertCurrentTest` assertion would fail.


ryanjduffy (approved):

ryanjduffy (changes requested):
Blocking to fix the lint/build error

PR #106, opened 2022-11-16
Add test harness for cypress events

ryanjduffy:
Start at a test harness that can capture the events sent from the cypress runner to the plugin and then re-run those events through the plugin to test it.


ryanjduffy:
This was useful at the time to debug an issue but not sure it's worth preserving so i'm closing

PR #108, opened 2022-11-22
Add category and hook data to test steps

ryanjduffy:


packages/playwright/src/reporter.ts
```diff
@@ -126,6 +142,8 @@ class ReplayPlaywrightReporter implements Reporter {
                 ? Math.max(0, s.startTime.getTime() - this.startTime)
                 : undefined,
               duration: s.duration,
+              hook: s.category === "hook" ? "beforeEach" : undefined,
```
ryanjduffy, line 145: The types for playwright do not make it apparent how to determine which hook we're in for a given step so this is wrong at least some of the time but it's a start

jaril (approved):

PR #126, opened 2022-12-21
Only disable cypress command log for non-interactive runs

ryanjduffy:


ryanjduffy:
This doesn't work because Cypress doesn't report the current value to plugins. See https://github.com/cypress-io/cypress/issues/20790

ryanjduffy:
Using `isTextTerminal` seems to do the trick

jaril (approved):

PR #129, opened 2023-01-04
Add replay metadata to crash reports

ryanjduffy:
Adds `id` and `metadata` to crash reports


ryanjduffy:
@miriambudayr - Figure we'll need something like this to `id` crash reports so I threw this together to get started. I can help align this to whatever you were planning to do for the backend portion.

miriambudayr:
Looks good! I'll keep you apprised.

PR #130, opened 2023-01-07
Upload replays in parallel

ryanjduffy:
Tested locally and in CI via https://github.com/replayio/flake/pull/15 and everything seems to work and does reduce the upload time to be the longest single replay (vs the cumulative upload time).


packages/replay/src/upload.ts
```diff
@@ -8,278 +8,259 @@ import { Options, OriginalSourceEntry, RecordingMetadata, SourceMapEntry } from
...
+
+function sha256(text: string) {
+  return crypto.createHash("sha256").update(text).digest("hex");
+}
+
+class ReplayClient {
+  client: ProtocolClient | undefined;
+  clientReady = defer<boolean>();
+
+  async initConnection(server: string, accessToken?: string, verbose?: boolean, agent?: any) {
```
jaril, line 22: small nit, but does this need to be async?

jaril (approved):
LGTM!

ryanjduffy:
@loganfsmyth, @jcmorrow, @jazzdan - This PR is changing the CLI to upload replays in parallel when using `upload-all` and the question came up this morning if we need to add some constraints to this to prevent spikey loads. I'm guessing that most cases are 10 or so replays from CI runs but I'm sure someone will come along with 100 small replays at some point. What do you all think? We could batch in groups of 10 or 20 if that will help mitigate.

packages/replay/src/main.ts
```diff
@@ -466,18 +462,14 @@ async function uploadAllRecordings(opts: Options & FilterOptions = {}) {
...
-        uploadedAll = false;
-      }
-    }
-  }
-  return uploadedAll;
+
+  const recordingIds = await Promise.allSettled(
+    recordings
+      .filter(r => !uploadSkipReason(r))
+      .map(r => doUploadRecording(dir, server, r, opts.verbose, opts.apiKey, opts.agent))
```
loganfsmyth, line 469: We should probably have some maximum on the amount of parallelism here

loganfsmyth: Grouping into 10-20 sounds great. 

The other concern here that I see is that `connectionUploadRecording` is currently written in such a way that the entire recording file could potentially be loaded into memory nearly all at once, so running this in parallel with many large recordings could put this at risk of using up all of Node's memory. We may want to consider updating `connectionUploadRecording` to read data from disk at the rate that it is able to flush the commands from the network buffers

loganfsmyth: It looks like your change went from having a single Buffer, to having many smaller buffers, but there's nothing that controls the _rate_ of the data being read from disk, so the most recording will likely be read into memory, since the machine will be able to read from disk a lot faster than it can send the data over the network.

What I'm saying is that ideally `gClient?.sendCommand("Internal.addRecordingData", { recordingId, offset, length }, b)` would ideally also take some extra callback parameter that would be called when the data has been flushed from the network buffers, so you'd for instance load 1MB from disk, send that, wait for it to flush from the network buffer, then read another 1MB, that way no matter how large the recording is, you only have 1MB of a given recording in memory at once.

loganfsmyth: In the case of this codebase, https://github.com/replayio/replay-cli/blob/391a37c41488d5f5231799152f7bdd187ea10e0d/packages/replay/src/client.ts#L62 the `send` function takes a callback as its second argument https://github.com/websockets/ws/blob/HEAD/doc/ws.md#websocketsenddata-options-callback

ryanjduffy:
@loganfsmyth tagging you for a second review. I've added a `--batch-size` parameter (defaulting to 20, capped at 25) and reworked the read + upload logic to ensure that we pause until the upload buffer has been flushed.

packages/replay/src/upload.ts
ryanjduffy: No. It's returning a promise either way but the `async` isn't required since we're not `await`-ing anything.

packages/replay/src/main.ts
```diff
@@ -458,16 +460,28 @@ async function processRecording(id: string, opts: Options = {}) {
...
+  }
+
+  const batchSize = Math.min(opts.batchSize || 20, 25);
+
+  let recordingsToUpload = recordings.filter(r => !uploadSkipReason(r));
+
+  let recordingIds: (string | null)[] = [];
+  while (recordingsToUpload.length > 0) {
+    const batch = recordingsToUpload.splice(0, batchSize);
+    const batchIds = await Promise.allSettled(
```
loganfsmyth, line 479: The main downside of this implementation is that _all_ handlers in a batch need to finish before the next batch will upload. Might be worth using something like https://github.com/sindresorhus/p-map since it has a `concurrency` option.

packages/replay/src/client.ts
```diff
@@ -66,11 +67,15 @@ class ProtocolClient {
...
+        if (!err && data) {
+          this.socket.send(data, callback);
+        } else {
+          callback?.(err);
+        }
+      }
     );
-    if (data) {
-      this.socket.send(data);
-    }
```
loganfsmyth, line 73: Might be ever so slightly better to do
```
this.socket.send(
      JSON.stringify({
        id,
        method,
        params,
        binary: data ? true : undefined,
        sessionId,
      }),
      data ? undefined : callback
);
if (data) {
  this.socket.send(data, callback);
}
```
but that's very much a nit

packages/replay/src/upload.ts
```diff
@@ -8,278 +8,259 @@ import { Options, OriginalSourceEntry, RecordingMetadata, SourceMapEntry } from
...
-  file.on("data", chunk => {
-    const cb = chunk instanceof Buffer ? chunk : Buffer.from(chunk);
-    buffer = buffer ? Buffer.concat([buffer, cb]) : cb;
+      buffer = buffer ? Buffer.concat([buffer, cb]) : cb;
 
-    if (buffer.length >= ChunkGranularity) {
-      const data = buffer.subarray(0, ChunkGranularity);
-      buffer = buffer.subarray(ChunkGranularity);
-      send(data);
+      if (buffer.length >= ChunkGranularity) {
```
loganfsmyth, line 181: I think this should maybe be a `while` instead of `if` since `buffer` could have enough data to be multiple chunks

packages/replay/src/upload.ts
```diff
@@ -8,278 +8,259 @@ import { Options, OriginalSourceEntry, RecordingMetadata, SourceMapEntry } from
...
+  async connectionUploadRecording(recordingId: string, path: string) {
+    if (!this.client) throw new Error("Protocol client is not initialized");
 
-  await gClient.sendCommand("Internal.reportCrash", { data });
-}
+    const file = fs.createReadStream(path);
 
-// Granularity for splitting up a recording into chunks for uploading.
-const ChunkGranularity = 1024 * 1024;
+    file.on("error", e => {
```
loganfsmyth, line 146: I think now that you're iterating with `for..await`, that will automatically catch and throw errors, so you shouldn't need this

ryanjduffy: I wasn't sure if it would catch all instances but that makes sense.

packages/replay/src/upload.ts
ryanjduffy: I don't think that'll be the case. The default `highwaterMark` is 16384 and `ChunkGranularity` is `1048576` so we should be always doing multiple reads for a chunk (unless there's somehow < 1 chunk which will be caught by the subsequent `send()`).

packages/replay/src/main.ts
ryanjduffy: Works for me!

packages/replay/src/upload.ts
loganfsmyth: ahh yeah fair enough!

packages/replay/src/main.ts
ryanjduffy: That was a nice addition. Thanks, @loganfsmyth ! Added in 06f3ded

packages/replay/src/main.ts
```diff
@@ -1,19 +1,12 @@
...
-  connectionUploadSourcemap,
-  connectionReportCrash,
-  closeConnection,
-  setRecordingMetadata,
-  buildRecordingMetadata,
-  connectionUploadOriginalSource,
-} from "./upload";
+
+// requiring v4 explicitly because it's the last version with commonjs support.
+// Should be upgraded to lhe latest when converting this code to es modules.
```
loganfsmyth, line 6: `lhe` => `the`

loganfsmyth (approved):

PR #131, opened 2023-01-09
Bail in after:spec if there are no tests in the spec

jaril:
Fix SCS-591


packages/cypress/src/reporter.ts
```diff
@@ -39,6 +39,11 @@ class CypressReporter {
       return [];
     }
 
+    // If the spec doesn't have any tests, we should bail here
```
ryanjduffy, line 42: Let's move this up to the top and group it up with the prior bail out. This'll fix SCS-593 too.

ryanjduffy (approved):
Moved the condition up and patched up a couple other things I found when testing it.

jaril:
Got it — thanks!

PR #138, opened 2023-01-19
Fix handling multiple end events for cypress asserts

ryanjduffy:
* Get the correct ref to `failedCommandLog` since it may not be the first log
* Always set the error on the step so if a retry succeeds, the `error` member on the step will be cleared


ryanjduffy:
I think we need to capture all of the errors because there doesn't appear to be any way to know if there's going to be a retry. Instead, when we get all the annotations in the UI, we'll have to find the last one for a given command and use that as the answer. This'll also address cases in which an assertion fails and then succeeds on retry.

jaril (approved):

PR #151, opened 2023-02-09
Add runner with support for diagnostic modes

ryanjduffy:
* Adds support for running cypress + replay in one of 4 modes:
  * `record` - Record everything (the default)
  * `record-on-retry` - Only record failures
  * `diagnostics` - Runs cypress a bunch of times (no record, record, record without asserts, and record with each feature disabled). `--level` option allows selection between "basic" (no record, record, record with asserts disabled) and "full" (basic + running with each feature disabled).
  * `stress` - Is really just `cypress-repeat` with a preset `-n` of 10
* Includes diagnostics config in both replay metadata (under the new custom key `x-replay-diagnostics`) and test metrics sent to webhooks.replay.io

```
npx @replayio/cypress run --mode diagnostics --browser "Replay Firefox"
```

### DIFF
--- a/packages/cypress/bin/replayio-cypress.ts
+++ b/packages/cypress/bin/replayio-cypress.ts
@@ -1,7 +1,8 @@
 #!/usr/bin/env node
 
+import { spawnSync } from "child_process";
 import install from "../src/install";
-import { getDiagnosticRetryCount } from "../src/mode";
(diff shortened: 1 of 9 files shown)


jasonLaster (commented):
Exciting

ryanjduffy:
We should treat this as "extra" beta for now and exercise it with benchmark repos for a while before promoting its use.

jaril (approved):

PR #153, opened 2023-02-13
sourcemap-upload: parameterize API server URL

jazzdan:
Fixes BAC-2792

### DIFF
--- a/packages/sourcemap-upload/src/index.ts
+++ b/packages/sourcemap-upload/src/index.ts
@@ -214,7 +214,7 @@ type PutOptions = {
 };
 
 async function sendUploadPUT(opts: PutOptions): Promise<Response> {
-  return fetch(`${getApiServer()}/sourcemap-upload`, {
+  return fetch(`${getApiServer()}/v1/sourcemap-upload`, {
     method: "PUT",
     headers: {
       "Content-Type": "application/json",
--- a/packages/sourcemap-upload/src/index.ts
+++ b/packages/sourcemap-upload/src/index.ts
@@ -29,6 +29,7 @@ export interface UploadOptions {
   ignore?: Array<string>;
   root?: string;
   log?: LogCallback;
+  apiServer?: string;
 }
 
 export async function uploadSourceMaps(opts: UploadOptions): Promise<void> {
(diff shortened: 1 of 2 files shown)


packages/sourcemap-upload/src/index.ts
```diff
@@ -203,14 +203,18 @@ async function processSourceMaps(opts: NormalizedOptions) {
   );
 }
 
+function getApiServer() {
+  return process.env.RECORD_REPLAY_API_SERVER || "https://api.replay.io/";
```
jazzdan, line 207: We should probably standardize this environment variables somewhere.

packages/sourcemap-upload/src/index.ts
```diff
@@ -203,14 +203,18 @@ async function processSourceMaps(opts: NormalizedOptions) {
...
+}
+
 type PutOptions = {
   groupName: string;
   apiKey: string;
   map: SourceMapToUpload;
 };
 
 async function sendUploadPUT(opts: PutOptions): Promise<Response> {
-  return fetch("https://api.replay.io/v1/sourcemap-upload", {
```
ryanjduffy, line 213: I think we still need the `v1/`, right?

jazzdan: Yup my bad

ryanjduffy (approved):

ryanjduffy:
@jazzdan - we should add this parameter to the CLI too, I think

ryanjduffy:
I suppose that could be a follow up but it'd be nice to not have to use a environment variable for it.

jazzdan:
I agree with that @ryanjduffy, what do you think of this?

ryanjduffy:
Let's also add the option to the CLI too:

https://github.com/replayio/replay-cli/blob/59dfcaf950662c7195cac94df0c29c66cd71f5f3/packages/replay/src/bin.ts#L98-L116

which will let us use `npx @replayio/replay upload-sourcemaps --group my-group --server <api-server-url>`. Leaning towards using `server` for the flag to match other commands that have that option.

wdyt?

jazzdan:
@ryanjduffy I'm fine with that. I don't think normal users of this package will ever need to change this so I don't think it needs to be advertised per se, but for the sake of standardization I will add it there.

ryanjduffy:
I expect you're right but it's nice for us when these configs are predictable too :)

PR #161, opened 2023-02-21
Sort steps with the same timestamp by event order

ryanjduffy:
## Issue

We were failing to build step metadata for replays in which step events occurred at the same timestamp which caused our sorting logic to order them incorrectly (or at least inconsistently).

## Resolution

Sort by the event type (to ensure starts are sorted before ends) if the timestamp is the same

## Additional Considerations

I think this occurs for commands or assertions that run synchronously like wrapping a jquery result and then `expect`-ing on it or some custom commands that for some reason aren't wrapped in a promise.

### DIFF
--- a/packages/cypress/src/steps.ts
+++ b/packages/cypress/src/steps.ts
@@ -30,6 +30,10 @@ function toRelativeTime(timestamp: string, startTime: number) {
   return toTime(timestamp) - startTime;
 }
 
+function toEventOrder(event: StepEvent) {
+  return ["test:start", "step:enqueue", "step:start", "step:end", "test:end"].indexOf(event.event);
+}
+
 function assertCurrentTest(
   currentTest: Test | undefined,
   step: StepEvent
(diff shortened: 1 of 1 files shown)


packages/cypress/src/steps.ts
```diff
@@ -30,6 +30,10 @@ function toRelativeTime(timestamp: string, startTime: number) {
   return toTime(timestamp) - startTime;
 }
 
+function toEventOrder(event: StepEvent) {
+  return ["test:start", "step:enqueue", "step:start", "step:end", "test:end"].indexOf(event.event);
```
jcmorrow, line 34: Tiny perf nit: an object lookup would be a lot faster here.

https://stackoverflow.com/questions/32234733/javascript-what-lookup-is-faster-array-indexof-vs-object-hash

But I know it does not matter, so feel free to ignore 🙃

jaril (approved):

packages/cypress/src/steps.ts
ryanjduffy: That tiny perf nit would lead to a tiny perf improvement ;)

PR #163, opened 2023-02-25
Pass through `RECORD_REPLAY_TEST_ENVIRONMENT` env var

Domiii:
* Pass through `RECORD_REPLAY_TEST_ENVIRONMENT` env var.
* https://linear.app/replay/issue/RUN-1433/enable-record-replay-test-environment-in-replayiocypress


ryanjduffy:
We shouldn't have to plumb through arbitrary environment variables. If we aren't seeing this one pass through, it might be a bug in either our plug-in or cypress.

PR #170, opened 2023-03-31
Add support for before and after hooks

ryanjduffy:
* Adds a `hooks` key to the test metadata to track `beforeAll` and `afterAll` hooks
* Updates the cypress step grouper to put any steps for these hooks into a separate array
* Updates the cypress and test-utils reporters to optionally accept `hooks` and feed them to metadata

Example: https://app.replay.io/recording/cypresse2eadding-spects--363dba41-5803-489b-9178-7c34e002b999


ryanjduffy:
I think there might be gaps with `before()` calls from within a `describe()` so i want to check that before landing this.

ryanjduffy:
Now we have `path` fields for each hook so we can group `before()` calls for multiple levels in the same spec file:

image

jaril (approved):

PR #174, opened 2023-04-21
Add reporter errors to test metadata

ryanjduffy:
Adds a new `reporterErrors` key to `test` metadata so we can better distinguish in clients when steps are missing and when the reporter failed in some unexpected way.

For replays made with this version or later of the plugin,
* `reporterErrors` will always be valued and will be an empty array when no errors occurred
* `reporterErrors` should only contain objects with the name `ReporterError` and include a numeric code and string message. May optionally include a `detail` object.

image

Current Error Codes
| Number | Description |
| -- | -- |
| -1 | Unknown error |
| 101 | No tests ran* |
| 201 | Mismatched steps (resulting in no steps for tests) |
| 202 | Missing test for a step (resulting in no steps for tests) |

* For cypress, can occur when the browser crashes or for a spec file that doesn't contain any tests (e.g. a util file that is matches the file pattern but isn't a test file)

### DIFF
--- a/packages/cypress/src/reporter.ts
+++ b/packages/cypress/src/reporter.ts
@@ -8,6 +8,7 @@ import {
 } from "@replayio/test-utils";
 import debug from "debug";
 
+import { Errors } from "./error";
 import { appendToFixtureFile, initFixtureFile } from "./fixture";
 import { getDiagnosticConfig } from "./mode";
 import { groupStepsByTest, mapStateToResult } from "./steps";
@@ -105,7 +106,7 @@ class CypressReporter {
     ) {
       const msg = "No test results found for spec " + spec.relative;
       this.debug(msg);
-      this.reporter.addError(new ReporterError(spec.relative, msg));
+      this.reporter.addError(new ReporterError(Errors.NoTestResults, msg, spec.relative));
 
       return {
         hooks: [],
@@ -118,6 +119,7 @@ class CypressReporter {
             path: [spec.relative],
             result: "unknown",
             relativePath: spec.relative,
+            steps: [],
           },
         ],
       };
@@ -125,12 +127,13 @@ class CypressReporter {
 
     let testsWithSteps: Test[] = [];
     let hooksWithSteps: Hook[] = [];
+
     try {
       const grouped = groupStepsByTest(spec, result.tests, this.steps, this.startTime!, this.debug);
       testsWithSteps = grouped.tests;
       hooksWithSteps = grouped.hooks;
     } catch (e: any) {
-      console.warn("Failed to build test step metadata for this replay.");
+      console.warn("Failed to build test step metadata for this replay");
       console.warn(e);
 
       this.reporter.addError(e);
@@ -160,6 +163,7 @@ class CypressReporter {
 
       return {
         title: t.title[t.title.length - 1] || spec.relative,
+        steps: [],
         // If we found the test from the steps array (we should), merge it in
         // and overwrite the default title and relativePath values. It won't
         // have the correct path or result so those are added and we bubble up
--- a/packages/replay/metadata/test.ts
+++ b/packages/replay/metadata/test.ts
@@ -83,6 +83,7 @@ const versions: Record<number, Struct<any, any>> = {
         {}
       )
     ),
+    reporterErrors: defaulted(array(any()), []),
     version: defaulted(number(), () => 1),
   }),
 };
--- a/packages/replay/package.json
+++ b/packages/replay/package.json
@@ -43,6 +43,7 @@
   "devDependencies": {
     "@types/debug": "^4.1.7",
     "@types/jest": "^28.1.5",
+    "@types/node-fetch": "^2.6.3",
     "@types/text-table": "^0.2.2",
     "@types/ws": "^8.5.3",
     "jest": "^28.1.3",
--- a/packages/test-utils/package.json
+++ b/packages/test-utils/package.json
@@ -32,5 +32,8 @@
     "node-fetch": "^2.6.7",
     "uuid": "^8.3.2"
   },
-  "gitHead": "e5d1208c1691f442ef4aeba6a238ec0b76a03a9b"
+  "gitHead": "e5d1208c1691f442ef4aeba6a238ec0b76a03a9b",
+  "devDependencies": {
+    "@types/debug": "^4.1.7"
+  }
 }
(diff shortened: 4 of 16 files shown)


ryanjduffy:
@bvaughn - tagged you review primarily for awareness but let me know if this approach doesn't align with your expectations!

packages/replay/metadata/test.ts
```diff
@@ -57,6 +57,7 @@ const versions: Record<number, Struct<any, any>> = {
           relativeStartTime: optional(number()),
           duration: optional(number()),
           steps: optional(array(any())),
```
ryanjduffy, line 59: Keeping this as `optional()` for now because I don't want to make any metadata validation more restrictive without bumping the version. We have plans to rework the test metadata structure soon so we'll address this change then

packages/replay/package.json
```diff
@@ -43,6 +43,7 @@
   "devDependencies": {
     "@types/debug": "^4.1.7",
     "@types/jest": "^28.1.5",
+    "@types/node-fetch": "^2.6.3",
```
ryanjduffy, line 46: Unrelated to the fix but this was missing and I didn't notice until I had to rebuild everything.

packages/test-utils/package.json
```diff
@@ -32,5 +32,8 @@
     "node-fetch": "^2.6.7",
     "uuid": "^8.3.2"
   },
-  "gitHead": "e5d1208c1691f442ef4aeba6a238ec0b76a03a9b"
+  "gitHead": "e5d1208c1691f442ef4aeba6a238ec0b76a03a9b",
+  "devDependencies": {
+    "@types/debug": "^4.1.7"
```
ryanjduffy, line 37: Unrelated to the fix but this was missing and I didn't notice until I had to rebuild everything.

packages/cypress/src/reporter.ts
```diff
@@ -118,19 +118,23 @@ class CypressReporter {
...
             relativePath: spec.relative,
+            steps: [],
           },
         ],
       };
     }
 
     let testsWithSteps: Test[] = [];
     let hooksWithSteps: Hook[] = [];
+    let stepsError: string | undefined;
```
bvaughn, line 129: So there would now be both an optional `error` attribute _and_ an optional `stepsError` attribute – both describing why a given test failed? That seems kind of confusing to me. What if both are present?

I prefer to avoid ambiguous data situations if possible. I guess how "okay" I am with this depends on how temporary it is. If you think we'll be able to get to SCS-1023 soon then it doesn't matter too much. Else I think I might want to re-think this design to incorporate a new `result` enum and then use a single` error` attribute.

(For that matter, I'd like the `error` attribute to _always_ be the thing we use when showing a top-level failure. Right now we have to scan steps for an error first, because it has more details– but I guess that's getting side tracked.)

ryanjduffy: It's not ideal but those error fields are referring to different things. `error` means the test failed and `stepsError` means the plugin had a problem. Either could be true without the other. IOW, a test can pass but we fail to build the step metadata so `stepsError` would be valued.

bvaughn: I realize that you're proposing they refer to different things, but I think that's an awkward data shape to work with.

It's particularly awkward naming, IMO, since test have "steps" which might have "errors" and "error" is more closely related to a `steps.error` than `stepsError` is.

ryanjduffy: I think this is the best we can do without changing the overall structure. If you'd rather wait for that, that's fine with me

bvaughn: I'm okay with us taking an incremental step, so long as we _do_ address the overall structure at some point (preferably not super far out) – but I don't think `stepsError` is the best name, even for an incremental step.

Maybe something like– `pluginError` or `integrationError` ? So it's clearly _separate_ from any errors that may have happened in one or more of the test steps.

ryanjduffy:
@bvaughn - i've reworked this (and pruned out the unrelated build-related changes) to add a `reporterErrors` array within the `test` key to house any errors. I also added numeric codes to the known errors (and a fallback code for unexpected errors) so we can match up to those on the FE.

```
// reporterErrors will be there for any new recording but still has to 
// be optional for now because earlier recordings will not have it.
// we can tighten this up in the v2 metadata release
if (replay.metadata.test.reporterErrors?.some(e => e.code === 201)) {
  console.log("mismatch step error ... no steps[]");
}
```

bvaughn:
Thanks for the update and for the PR description!

Error code 101 makes sense, I think.

I'm a little unclear on what the difference is between an error 201 and 202 though.

What does "mismatched steps" mean?
What does "missing test for a step" mean?

ryanjduffy:
> Thanks for the update and for the PR description!
> 
> Error code 101 makes sense, I think.
> 
> I'm a little unclear on what the difference is between an error 201 and 202 though.
> 
> What does "mismatched steps" mean? What does "missing test for a step" mean?

Mismatched steps is the most common error state right now. In short, it means that we received a step event (usually a step end event) and we couldn't match it up to the start event.

The "Missing test" error is rare; i can't think of a case I've seen it. It'll throw when we receive a step event but we can't find the test to which it belongs.

ryanjduffy:
@bvaughn - Can you confirm this API will work for the FE work?

@jaril - nudge for code review

bvaughn:
Hm, I see. I guess that makes sense, although my concern would be that I'm not sure what error messaging we would show for those errors. Any wording suggestions?

Otherwise I think the structure sounds okay.

> Adds a new reporterErrors key to test metadata

Can you clarify what it meant by "test metadata".

Does this new attribute get added to `TestItem` (the thing with the "steps") or to `TestMetadata`?

ryanjduffy:
> Hm, I see. I guess that makes sense, although my concern would be that I'm not sure what error messaging we would show for those errors. Any wording suggestions?

For the 2xx errors, something in the spirit of "We were unable to capture steps for this test due to a problem with the `@replayio/cypress` plugin when the replay was created."

> > Adds a new reporterErrors key to test metadata
> 
> Can you clarify what it meant by "test metadata".
> 
> Does this new attribute get added to `TestItem` (the thing with the "steps") or to `TestMetadata`?

Not sure off hand how it maps to the types in devtools but it's a top-level key under `test`: `recording.metadata.test.reporterErrors`

PR #177, opened 2023-04-28
Allow uploading directly to S3

jcmorrow:
It's pretty easy to overwhelm our servers by sending them a ton of CI recordings at once. It's been happening multiple times a day sometimes for weeks. In the case of dozens or hundreds of recordings coming from CI, there is not really any advantage to using a websocket over HTTP, and HTTP is a *lot* simpler (and we don't have to worry about scale ever, it just goes straight to AWS).

### DIFF
--- a/packages/replay/src/main.ts
+++ b/packages/replay/src/main.ts
@@ -360,6 +360,8 @@ async function doUploadRecording(
     return recording.id;
   }
 
+  const { size } = await fs.promises.stat(recording.path!);
+
   debug("Uploading recording %o", recording);
   const client = new ReplayClient();
   if (!(await client.initConnection(server, apiKey, verbose, agent))) {
@@ -372,7 +374,11 @@ async function doUploadRecording(
   const metadata = recording.metadata
     ? await client.buildRecordingMetadata(recording.metadata, { verbose })
     : null;
-  const recordingId = await client.connectionCreateRecording(recording.id, recording.buildId!);
+  const { recordingId, uploadLink } = await client.connectionCreateRecording(
+    recording.id,
+    recording.buildId!,
+    size
+  );
   debug(`Created remote recording ${recordingId}`);
   if (metadata) {
     try {
@@ -388,7 +394,14 @@ async function doUploadRecording(
     recordingId,
   });
 
-  await client.connectionUploadRecording(recordingId, recording.path!);
+  await client.uploadRecording(recording.path!, uploadLink, size);
+
+  debug("%s: Uploaded %d bytes", recordingId, size);
+
+  // Explicitly mark the recording complete so the server knows that all
+  // recording data has been uploaded. This means if someone presses Ctrl+C,
+  // the server doesn't save a partial recording.
+  await client.finishRecording(recordingId);
 
   for (const sourcemap of recording.sourcemaps) {
     try {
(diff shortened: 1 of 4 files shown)


packages/replay/src/upload.ts
```diff
@@ -141,57 +140,14 @@ class ReplayClient {
...
-    }
-
-    debug("%s: Uploaded %d bytes", recordingId, offset);
-
-    // Explicitly mark the recording complete so the server knows that it has
-    // been sent all of the recording data, and can save the recording.
-    // This means if someone presses Ctrl+C, the server doesn't save a
-    // partial recording.
-    await this.client.sendCommand("Internal.finishRecording", { recordingId });
+    const size = fs.statSync(path).size;
```
loganfsmyth, line 145: We should avoid the Sync function, especially since we can `await` here with no problem

packages/replay/src/main.ts
```diff
@@ -388,7 +394,7 @@ async function doUploadRecording(
     recordingId,
   });
 
-  await client.connectionUploadRecording(recordingId, recording.path!);
+  await client.connectionUploadRecording(recording.path!, uploadLink);
```
loganfsmyth, line 397: Looks like we can pass the size in here too?

packages/replay/src/main.ts
```diff
@@ -360,6 +360,8 @@ async function doUploadRecording(
     return recording.id;
   }
 
+  const size = fs.statSync(recording.path!).size;
```
loganfsmyth, line 363: Let's avoid the `Sync`-suffix functions

packages/sourcemap-upload/package-lock.json
```diff
@@ -6,7 +6,7 @@
 	"packages": {
 		"": {
 			"name": "@replayio/sourcemap-upload",
-			"version": "1.0.6",
+			"version": "1.0.7",
 			"license": "BSD-3-Clause",
```
loganfsmyth, line 10: Is all this from an unrelated PR? Can we land that first or in a separate commit so it's easier to see what's changing I this PR?

packages/replay/src/upload.ts
```diff
@@ -141,57 +140,14 @@ class ReplayClient {
...
-      await send(buffer);
-    }
-
-    debug("%s: Uploaded %d bytes", recordingId, offset);
-
-    // Explicitly mark the recording complete so the server knows that it has
-    // been sent all of the recording data, and can save the recording.
-    // This means if someone presses Ctrl+C, the server doesn't save a
-    // partial recording.
-    await this.client.sendCommand("Internal.finishRecording", { recordingId });
```
loganfsmyth, line 194: We maaaye still want to call this afterward so the API knows when the CLI thinks it is done? At least then if the user disconnects without finishing uploading to S3, we'll be able to see the disconnect-without-finish and clear the DB the DB entry?

jcmorrow: Yeah I think I'm good with that.

packages/sourcemap-upload/package-lock.json
jcmorrow: I think this was all from me installing `node-fetch`, but maybe I ran some unnecessary commands alongside that. I can try and reduce the noise here.

packages/replay/src/upload.ts
jcmorrow: Done!

packages/sourcemap-upload/package-lock.json
jcmorrow: Should be significantly reduced now!

loganfsmyth (approved):

PR #178, opened 2023-05-02
Change protocol methods used to upload CLI recording

jcmorrow:
We will now use `Internal.beginRecordingUpload` and `Internal.endRecordingUpload` to mark these events. We won't actually be doing integrity checking to see if the whole recording got properly uploaded right now, but we could add in that (or any other integrity checks) later if we wanted to.

Pairs with https://github.com/replayio/backend/pull/7623

### DIFF
--- a/packages/replay/src/upload.ts
+++ b/packages/replay/src/upload.ts
@@ -52,21 +52,20 @@ class ReplayClient {
     return this.clientReady.promise;
   }
 
-  async connectionCreateRecording(id: string, buildId: string, size: number) {
+  async connectionBeginRecordingUpload(id: string, buildId: string, size: number) {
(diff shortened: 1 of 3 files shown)


loganfsmyth:
> and having it be part of the API but not having it behave in a robust way seems worse than just not having it at all.

Let's chat, I don't agree.

packages/replay/src/upload.ts
```diff
@@ -55,18 +55,17 @@ class ReplayClient {
   async connectionCreateRecording(id: string, buildId: string, size: number) {
```
loganfsmyth, line 55: Maybe worth renaming this now too to `connectionBeginRecordingUpload`?

loganfsmyth (approved):

PR #181, opened 2023-05-23
Use retry with exponential backoff

jcmorrow:
To try and fix S3 ECONNRESET errors

Fixes https://linear.app/replay/issue/BAC-3304/http-uploads-for-recordings-failing

### DIFF
--- a/packages/replay/src/main.ts
+++ b/packages/replay/src/main.ts
@@ -14,7 +14,7 @@ import {
   getPuppeteerBrowserPath,
   updateBrowsers,
 } from "./install";
-import { getDirectory, maybeLog } from "./utils";
+import { exponentialBackoffRetry, getDirectory, maybeLog } from "./utils";
 import { spawn } from "child_process";
 import {
   ExternalRecordingEntry,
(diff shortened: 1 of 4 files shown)


packages/replay/src/main.ts
```diff
@@ -394,7 +394,7 @@ async function doUploadRecording(
     recordingId,
   });
 
-  await client.uploadRecording(recording.path!, uploadLink, size);
+  await exponentialBackoffRetry(() => client.uploadRecording(recording.path!, uploadLink, size));
```
ryanjduffy, line 397: Can you add a `debug()` entry within `client.uploadRecording()` so we have visibility into retries?

ryanjduffy (changes requested):
Good plan! One note on some observability but otherwise let's give this a go!

packages/replay/src/main.ts
jcmorrow: yeah good call, I'll also record how long it took to fail, because I feel like that might tell us something.

jcmorrow: Looks like `debug` does the time offset automatically! ❤️

PR #185, opened 2023-06-02
Install Chrome for mac

jasonLaster:
This is enough to run `cypress run --browser "Replay Chromium"` in our Admin app and get some replays!

### DIFF
--- a/packages/replay/src/install.ts
+++ b/packages/replay/src/install.ts
@@ -12,6 +12,7 @@ const debug = dbg("replay:cli:install");
 
 const EXECUTABLE_PATHS = {
   "darwin:firefox": ["firefox", "Nightly.app", "Contents", "MacOS", "firefox"],
+  "darwin:chromium": ["chromium", "Contents", "MacOS", "Chromium"],
   "linux:chromium": ["chrome-linux", "chrome"],
   "linux:firefox": ["firefox", "firefox"],
 } as const;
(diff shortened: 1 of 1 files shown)


packages/replay/src/install.ts
```diff
@@ -20,6 +21,9 @@ function getBrowserDownloadFileName<K extends keyof typeof EXECUTABLE_PATHS>(key
   switch (key) {
     case "darwin:firefox":
       return process.env.RECORD_REPLAY_FIREFOX_DOWNLOAD_FILE || "macOS-replay-playwright.tar.xz";
+    case "darwin:chromium":
+      return process.env.RECORD_REPLAY_CHROMIUM_DOWNLOAD_FILE || "macOS-replay-chromium.tar.xz";
```
jasonLaster, line 25: we might want to extend this to also support Chrome for ARM which can be downloaded as well and would be faster

ryanjduffy (approved):

ryanjduffy:
I tacked on a few other changes to this I want to note:

* Browsers are now installed into `$HOME/.replay/runtimes` rather than `$HOME/.replay/playwright` and `$HOME/.replay/puppeteer`. The separation is only duplicating the chrome install and not adding any value. Both FF for playwright and chromium will be installed in the same spot. This doesn't effect manual installs of either browser for interactive use.
* `replay update-browsers` will now install or update browsers. Previously, it would log if an potential browser couldn't be updated because it wasn't installed
* Chromium for mac will install *either* the ARM or the x86 version based on `process.arch`

I tested the installer locally and ran cypress with both ff and chromium locally and was able to record and upload successfully.

cc: @bhackett1024 + @jaril for feedback on any of the above changes before I land this.

PR #191, opened 2023-06-16
Add early bailout when annotation hook is not found

jasonLaster:


packages/cypress/src/support.ts
```diff
@@ -252,9 +255,12 @@ function addAnnotation(path: string[], event: string, data?: Record<string, any>
     titlePath: path,
   });
 
-  window.top &&
-    window.top.__RECORD_REPLAY_ANNOTATION_HOOK__ &&
-    window.top.__RECORD_REPLAY_ANNOTATION_HOOK__("replay-cypress", JSON.stringify(payload));
+  if (!window.top || !window.top.__RECORD_REPLAY_ANNOTATION_HOOK__) {
+    debug("No annotation hook found");
+    return;
+  }
```
ryanjduffy, line 261: This file runs in the browser process so debug logging doesn't really help us (and we wouldn't want to log when not running in the replay browser) but this early return will help us see when we're not hitting L263 more easily

ryanjduffy (changes requested):

ryanjduffy (approved):

PR #195, opened 2023-06-26
Buffer cypress plugin tasks to dispatch after test

ryanjduffy:
## Issue

Cypress 12.14 changed the behavior of `Cypress.backend` to no longer support sending tasks that didn't participate in the command chain. It now throws for tasks which we were discarding because the prior implementation threw spurious errors.

## Analysis

I tried to find ways to use the same architecture using either `cy.task` or `Cypress.backend("run:privileged")` but neither worked as I hoped:

* `cy.task` would inject itself into the command chain and swallow the prior command's response so that commands like `cy.get("div").eq(0).click()` would fail because a new `cy.task()` would be inserted after the `get` causing its returned subject to be lost. 
* `Cypress.backend("run:privileged")` introduces additional "verification" constraints to try to enforce the privileged state. I suspect this could be worked around but would expose us to future regressions because it isn't a public API

## Resolution

Now that we no longer use the metadata for timings, it doesn't matter when they arrive at the plugin. So instead of dispatching them as they arrive, we can enqueue them and send them all at the end of the test. This allows us to use `cy.task()` (public API) and not interrupt any commands because they are not invoked in response to any specific commands.


ryanjduffy:
I left a couple WIP commits in the branch in case they're relevant in the future. The last commit has the final fix.

jasonLaster (approved):

PR #200, opened 2023-06-29
Delete recordings by default after an upload

jaril:


ryanjduffy (changes requested):
This will work but will also have the side effect of deleting the entries from the recording log which will mean you can't see them with `replay ls --all`. Instead, I think we should just remove the recording file (and probably the sourcemaps) by calling `maybeRemoveRecordingFile(recording)` from `doUploadRecording()`. WDYT?

ryanjduffy:
> This will work but will also have the side effect of deleting the entries from the recording log which will mean you can't see them with `replay ls --all`. Instead, I think we should just remove the recording file (and probably the sourcemaps) by calling `maybeRemoveRecordingFile(recording)` from `doUploadRecording()`. WDYT?

noting also that `maybeRemoveRecordingFile` only removes the .dat file right now but should probably also remove the associated sourcemaps and such.

ryanjduffy (approved):

PR #206, opened 2023-07-09
Add launch command

jasonLaster:
Adds `npx replay launch` command for launching chromium

### DIFF
--- a/packages/replay/src/bin.ts
+++ b/packages/replay/src/bin.ts
@@ -14,12 +14,15 @@ import {
   updateMetadata,
 } from "./main";
 import {
+  BrowserName,
   CommandLineOptions,
   FilterOptions,
   MetadataOptions,
   SourcemapUploadOptions,
   UploadOptions,
 } from "./types";
+import { spawn } from "child_process";
+import { ensurePlaywrightBrowsersInstalled, getPlaywrightBrowserPath } from "./install";
 
 // TODO(dmiller): `--json` should probably be a global option that applies to all commands.
 program
@@ -38,6 +41,11 @@ program
   .option("--server <address>", "Alternate server to upload recordings to.")
   .option("--api-key <key>", "Authentication API Key")
   .action(commandUploadRecording);
+
+program
+  .command("launch <browser> [url]")
+  .description("Launch chromium or gecko")
+  .action(commandLaunchBrowser);
 
 program
   .command("process <id>")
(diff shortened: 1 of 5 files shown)


packages/replay/src/bin.ts
```diff
@@ -153,6 +161,25 @@ async function commandUploadRecording(id: string, opts: CommandLineOptions) {
...
+  const browserName = browser as BrowserName;
+  await ensurePlaywrightBrowsersInstalled(browserName);
+
+  const execPath = getPlaywrightBrowserPath(browserName);
+  if (!execPath) {
+    console.log("Path could not be found");
+    process.exit(1);
+  }
+
+  spawn(execPath);
```
JoelEinbinder, line 179: You should use some launch args here to prevent the welcome screen and maybe some other config. https://peter.sh/experiments/chromium-command-line-switches/#no-first-run

packages/replay/src/bin.ts
```diff
@@ -153,6 +161,25 @@ async function commandUploadRecording(id: string, opts: CommandLineOptions) {
...
+  process.exit(1);
```
JoelEinbinder, line 180: Why exit code 1 here?

packages/replay/src/bin.ts
```diff
@@ -39,6 +42,11 @@ program
   .option("--api-key <key>", "Authentication API Key")
   .action(commandUploadRecording);
 
+program
+  .command("launch <browser>")
```
JoelEinbinder, line 46: I'd like `launch <browser> <url>` but that's just me.

jasonLaster: we can make that happen

packages/replay/src/bin.ts
jasonLaster: good catch

packages/replay/src/bin.ts
```diff
@@ -153,6 +161,29 @@ async function commandUploadRecording(id: string, opts: CommandLineOptions) {
...
+  const browserName = browser as BrowserName;
+  await ensurePlaywrightBrowsersInstalled(browserName);
+
+  const execPath = getPlaywrightBrowserPath(browserName);
+  if (!execPath) {
+    console.log("Path could not be found");
+    process.exit(1);
+  }
+
+  spawn(execPath, ["--no-first-run", url]);
```
JoelEinbinder, line 183: TIL --no-first-run works in firefox

JoelEinbinder (approved):

packages/replay/src/bin.ts
jasonLaster: good catch - these should be browser specific

ryanjduffy:
@jasonLaster - I made a few changes. They're mostly refactoring but wanted to call out that moved the browser to an option defaulting to chromium since that's the direction we want folks to take.

The usage now is:
```
# launch chromium with optional url
npx @replayio/replay launch [url]

# launch firefox with optional url
npx @replayio/replay launch -b firefox [url]
```

ryanjduffy (approved):

PR #207, opened 2023-07-10
Add e2e test action for cypress plugin

ryanjduffy:
Extends the earthly script to run the [flake](https://github.com/replayio-public/flake) benchmark repo with the current (unreleased) cypress plugin code to identify regressions.

Next will be to extend this with metadata test cases from the produced replays but for now we can verify with the replay links.


.github/workflows/ci.yml
```diff
@@ -7,7 +7,11 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+      - uses: actions/checkout@v2
```
ryanjduffy, line 10: This should probably be in the earthfile via `GIT CLONE` but I started with this so it's here for now.

ryanjduffy: Removing this means we can remove the ignore file changes too

Earthfile
```diff
@@ -15,6 +16,38 @@ test:
   FROM +build
   RUN npm test
 
+setup:
+  FROM +build
+  RUN apt update && apt install xz-utils
+  RUN npx @replayio/playwright install
+  # download binary openssl packages from Impish builds
+  RUN wget http://security.ubuntu.com/ubuntu/pool/main/o/openssl/openssl_1.1.1f-1ubuntu2.19_amd64.deb
```
ryanjduffy, line 24: Adds support for openssl 1.1.1

package.json
```diff
@@ -17,7 +17,8 @@
   "license": "ISC",
   "devDependencies": {
     "lerna": "^4.0.0",
-    "prettier": "^2.7.1"
+    "prettier": "^2.7.1",
+    "typescript": "^5.1.6"
```
ryanjduffy, line 21: The build steps for the packages were failing because `tsc` wasn't installed.

Earthfile
```diff
@@ -1,11 +1,12 @@
-VERSION 0.6
-FROM node:lts-alpine
+VERSION --wait-block 0.6
```
jazzdan, line 1: Do we need `--wait-block`?

jazzdan (approved):

Earthfile
ryanjduffy: Nope. I tried to use `WAIT` at one point but we can drop this now.

PR #208, opened 2023-07-11
Wrap annotation creation with transpile-safe fn

ryanjduffy:


ryanjduffy:
The build failure was due to a replay upload issue and can be ignored. i just landed that testing code so it needs some hardening.

jasonLaster (approved):

PR #210, opened 2023-07-13
Add install option for windows chromium

ryanjduffy:
We don't build gecko+playwright for windows so this only adds support for installing the beta chromium build on windows.

### DIFF
--- a/packages/replay/src/install.ts
+++ b/packages/replay/src/install.ts
@@ -129,7 +129,7 @@ function updateBrowsers(opts: Options) {
 }
 
 function getPlatformKey(browserName: BrowserName) {
-  const key = `${"windows"}:${browserName}`;
+  const key = `${process.platform}:${browserName}`;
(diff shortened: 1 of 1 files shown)


packages/replay/src/install.ts
```diff
@@ -114,13 +129,9 @@ function updateBrowsers(opts: Options) {
...
 
 function getPlatformKey(browserName: BrowserName) {
-  const key = `${process.platform}:${browserName}`;
-  switch (key) {
-    case "darwin:firefox":
-    case "linux:firefox":
-    case "darwin:chromium":
-    case "linux:chromium":
-      return key;
+  const key = `${"windows"}:${browserName}`;
```
hbenl, line 132: I guess `"windows"` should be changed back to `process.platform`(?)

ryanjduffy: thanks!

hbenl (approved):

PR #214, opened 2023-07-17
Drop command options from cypress step args

ryanjduffy:


ryanjduffy:
Also noting we can eventually show live references to the actual args since we have the `enqueue` event and can get at the `args` array from the annotation point.
image

hbenl (approved):

PR #219, opened 2023-07-20
Fix dropping args from cy.intercept()

ryanjduffy:
`cy.intercept()` does not accept cypress options as a trailing arg so we shouldn't drop the last object arg which could be a valid arg in this case.


ryanjduffy:
This is a follow up to #218 which fixes the runtime failure. This fixes the reporting bug in which we won't show this arg in the UI (although we'll have simplified it to `[object Object]` right now so the UI isn't terribly useful).

jaril (approved):

PR #229, opened 2023-08-09
install: arm releases no longer have a suffix

jazzdan:
Fixes SCS-1342


ryanjduffy:
Is this only a problem for us if we have an old build and installs from now on will follow the expected path? If so, I don't think we need to fix this and have extra compatibility code hanging around.

jazzdan:
@ryanjduffy that's right. Sounds good: I'll change this so that it isn't backwards compatible.

ryanjduffy (approved):

PR #237, opened 2023-08-17
Add login method to cli

ryanjduffy:
Ported the "remote" auth logic from gecko to the CLI so users can `replay login` to interact with the cli instead of requiring creating an API key first.

### DIFF
--- a/packages/replay/src/auth.ts
+++ b/packages/replay/src/auth.ts
@@ -5,7 +5,7 @@ import { readFile, writeFile } from "fs/promises";
 import path from "path";
 
 import { query } from "./graphql";
-import { getDirectory, openExecutable } from "./utils";
+import { getDirectory, maybeLog, openExecutable } from "./utils";
 import { Options } from "./types";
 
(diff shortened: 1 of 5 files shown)


packages/replay/src/auth.ts
```diff
@@ -0,0 +1,265 @@
...
+    if (timedOut) {
+      debug("Timed out waiting for auth request");
+      throw { id: "timeout" };
+    }
+
+    try {
+      const refreshToken = await fetchToken(key);
+      maybeLog(options.verbose, "🔑 Fetching token");
+
+      return await refresh(refreshToken);
```
hbenl, line 185: Why do we use a refresh token here? Why doesn't `CloseAuthRequest` simply return the access token?

ryanjduffy: We only store the refresh token in the db right now and exchange it for a token right away so the refresh token is invalidated and we're not holding into anything that could be used to impersonate the user.

PR #238, opened 2023-08-18
Add fixture for playwright annotations

ryanjduffy:
Beta-rewrite of playwright metadata using an injected fixture instead of the reporter.

Adds the following features:

* Detailed command arguments
* Annotations in the browser replay for step start/end events
* Call stack and source file details for each step

Controversial Implementation details:

* We are hot wiring the playwright fixture internals to inject our fixture. This is the lowest friction version but the PR also exports our fixture so users could manually apply it.
* We are patching `TestInfoImpl` to give us a way to capture non-api events (currently just `expect()`s). Those are available via the reporter and it's possible we could merge from there but the reporter data is missing the `params` which include the expected value which would is available via the patch.

Known issues:

- [X] Retried commands are emitted multiple times by the fixture for a single reporter entry
  - This isn't actually what was happening. devtools has a `waitFor` util that retries a callback which was producing different numbers of calls between my old/new tests 
- [x] expect's the are api calls (e.g. `expect.toBeVisible`) are emitted from both the step handler and the api call handler
- [x] We're using a single WebSocket port to communicate between the fixture and the reporter which I expect will break when running tests in parallel


ryanjduffy:
@hbenl + @jasonLaster - this is still pretty rough but it's getting closer. This version will auto-add itself to the playwright/test default fixtures. We get the frames here and a step id but the step id isn't passed through to the reporter at all. More to do there.

ryanjduffy:
Pushed a commit that adds a websocket to communicate between the fixture and the reporter. it includes the internal playwright step id and (i think) the same call stack. I didn't get to matching them up with the reporter steps but I don't expect that to be too bad. Need to try it out with a more complex test case than my two step playwright playground.

ryanjduffy:
@hbenl - This still has a ways to go before it's ready for widespread use but wanted to get another set of eyes on it with the hope we can do an alpha release of it and start to try it out in devtools.

ryanjduffy:
Fixed the known issues. I expect the whole PR needs a refactoring pass to clean things up and make the APIs sane but I think  it's pretty close functionally.

I will note that I haven't tested the annotations nor adapted the devtools code to find the stacks in the new metadata home so that should probably at least be drafted before we land this so we know it all works e2e.

packages/playwright/src/fixture.ts
```diff
@@ -0,0 +1,296 @@
...
+      title: testInfo.title,
+      scope: testInfo.titlePath.slice(3, -1),
+    },
+  };
+
+  function addAnnotation(event: string, id?: string, data?: Record<string, any>) {
+    if (id) {
+      page
+        .evaluate(ReplayAddAnnotation, [event, id, JSON.stringify({ ...data, test: testIdData })])
+        .catch(e => console.error);
```
hbenl, line 161: Shouldn't this be `catch(e => console.error(e))` or `catch(console.error)`?

packages/playwright/src/fixture.ts
```diff
@@ -0,0 +1,296 @@
...
+
+export async function replayFixture(
+  { playwright, page }: { playwright: any; page: Page },
+  use: () => Promise<void>,
+  testInfo: TestInfo
+) {
+  debug("Setting up replay fixture");
+
+  const port = getServerPort();
+  const ws = new WebSocket(`ws://localhost:${port}`);
```
hbenl, line 139: Are we sure that the websocket server has already been started or do we need to eventually add some retry logic here?

hbenl (approved):

packages/playwright/src/fixture.ts
ryanjduffy: Yup!

packages/playwright/src/fixture.ts
ryanjduffy: As far as I can tell, the reporter (where the server lives) always starts first.

ryanjduffy:
I've retargeted this against a new branch, `playwright-next`, so we can iterate on this under beta npm releases without `main` stomping on it. Once it all seems stable, i'll open a PR to merge the feature branch back to main and push a full release of the playwright plugin.

PR #240, opened 2023-08-21
Add upload option to cypress plugin

ryanjduffy:
Add upload support after cypress run completes

Supporting changes:
* Supporting a function for `filter` in the node API in addition to the jsonata string
* Exporting `log` from test-utils logging
* Adding `directory` to `updateMetadata`

Fixes SCS-1407

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,6 +11,7 @@ jobs:
         with:
           repository: replayio-public/flake
           path: e2e-repos/flake
+          ref: ryan/upload-test
       - name: Download latest earthly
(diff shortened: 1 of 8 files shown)


jasonLaster:
What's the best way to review / test this work?

For example I'm wondering if we want some examples projects in this repo that exercise this functionality.

.github/workflows/ci.yml
```diff
@@ -11,6 +11,7 @@ jobs:
         with:
           repository: replayio-public/flake
           path: e2e-repos/flake
+          ref: ryan/upload-test
```
ryanjduffy, line 14: Remove before merging

hbenl (approved):

PR #241, opened 2023-08-24
Add --warn to CLI commands

ryanjduffy:
When running the CLI in CI environments, users may want to suppress non-zero exit codes to prevent failing builds when steps like uploading replays fails. The --warn option was already supported by the metadata command so I have extended it to all commands and add try/catch blocks to ensure we are handling throw errors correctly.

### DIFF
--- a/packages/replay/src/bin.ts
+++ b/packages/replay/src/bin.ts
@@ -1,5 +1,6 @@
 import { LogCallback, uploadSourceMaps } from "@replayio/sourcemap-upload";
 import { program } from "commander";
+import dbg from "debug";
 import { formatAllRecordingsHumanReadable, formatAllRecordingsJson } from "./cli/formatRecordings";
 import {
   listAllRecordings,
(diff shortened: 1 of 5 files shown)


packages/replay/src/bin.ts
```diff
@@ -71,42 +82,27 @@ program
...
   .description("Remove a specific recording.")
-  .option("--directory <dir>", "Alternate recording directory.")
   .action(commandRemoveRecording);
 
-program
-  .command("rm-all")
-  .description("Remove all recordings.")
-  .option("--directory <dir>", "Alternate recording directory.")
-  .action(commandRemoveAllRecordings);
+program.command("rm-all").description("Remove all recordings.").action(commandRemoveAllRecordings);
```
hbenl, line 99: Shouldn't this also use `commandWithGlobalOptions()`?

ryanjduffy: Yep! Missed it

ryanjduffy: Fixed!

hbenl (approved):

PR #243, opened 2023-09-04
Wrap cypress on with multi-handler safe code

ryanjduffy:
## Issue

The `on` fn passed to cypress plugins doesn't universally support multiple handlers and will sometimes just overwrite the previously applied handler (e.g. for `after:run`).

## Resolution

Add our own wrapper which users can call at the start of their `setupNodeEvents` to wrap `on` with a multi-plugin safe version.

```
import replayCypress from '@replayio/cypress'
// ...
setupNodeEvents(on, config) {
  on = replayCypress.wrapOn(on);
  replayCypress.plugin(on, config);
  // other plugins ...
}
```

Fixes SCS-1394

### DIFF
--- a/packages/cypress/src/index.ts
+++ b/packages/cypress/src/index.ts
@@ -132,62 +132,60 @@ function onReplayTask(value: any) {
   return true;
 }
 
-const plugin: Cypress.PluginConfig = (on, config) => {
-  cypressReporter = new CypressReporter(config, debug);
+const cypressOnWrapper = (base: Cypress.PluginEvents): Cypress.PluginEvents => {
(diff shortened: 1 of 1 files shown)


jasonLaster:
Lol. At some point we should just gift cypress a safer event listener

ryanjduffy:
I'm treating this a beta feature that we can hand out to customers to try when necessary. I've done some testing with it and it appears to fix integration issues where Cypress is overwriting event handlers from subsequent plugins but I'm ready to document it as a supported feature just yet.

hbenl (approved):

PR #244, opened 2023-09-07
Fix handling of playwright errors without a message

ryanjduffy:
## Issue

The error message parsing wasn't handling cases in which non-Error objects were thrown causing a run-time error accessing  `.split()` on an undefined value.

## Resolution

Handle when either `value` or `message` is set with a fallback to "Unknown Error"

Fixes SCS-1413

### DIFF
--- a/packages/playwright/src/reporter.ts
+++ b/packages/playwright/src/reporter.ts
@@ -27,10 +27,18 @@ export function getMetadataFilePath(workerIndex = 0) {
 }
 
 function extractErrorMessage(error: TestError) {
-  const errorMessageLines = removeAnsiCodes(error.message)!.split("\n");
-  let stackStart = errorMessageLines.findIndex(l => l.startsWith("Call log:"));
-  stackStart = stackStart == null || stackStart === -1 ? 10 : Math.min(stackStart, 10);
-  return errorMessageLines.slice(0, stackStart).join("\n");
+  if (error.message) {
+    // Error message. Set when [Error] (or its subclass) has been thrown.
+    const errorMessageLines = removeAnsiCodes(error.message)!.split("\n");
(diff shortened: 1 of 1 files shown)


packages/playwright/src/reporter.ts
```diff
@@ -27,10 +27,18 @@ export function getMetadataFilePath(workerIndex = 0) {
 }
 
 function extractErrorMessage(error: TestError) {
-  const errorMessageLines = removeAnsiCodes(error.message)!.split("\n");
-  let stackStart = errorMessageLines.findIndex(l => l.startsWith("Call log:"));
-  stackStart = stackStart == null || stackStart === -1 ? 10 : Math.min(stackStart, 10);
-  return errorMessageLines.slice(0, stackStart).join("\n");
+  if (error.message) {
+    // Error message. Set when [Error] (or its subclass) has been thrown.
+    const errorMessageLines = removeAnsiCodes(error.message)!.split("\n");
```
bvaughn, line 32: Is `removeAnsiCodes(...)!` safe? If that method says it returns null/undefined, seems like we should check for that case?

ryanjduffy: It is. it only returns undefined if the input parameter is undefined which we guarded for above.

ryanjduffy: I reworked this to remove the assert to be safer though. Thanks for the nudge!

bvaughn: > It is. it only returns undefined if the input parameter is undefined which we guarded for above.

Oooh! Makes sense. Seems like the ideal type for that would be `(value: string) => string)`

> I reworked this to remove the assert to be safer though

Thanks!

hbenl (approved):

PR #246, opened 2023-09-13
replay: add attach flag to launch command

jazzdan:
Fixes BAC-3976

### DIFF
--- a/packages/replay/src/bin.ts
+++ b/packages/replay/src/bin.ts
@@ -64,6 +64,11 @@ commandWithGlobalOptions("upload <id>")
 commandWithGlobalOptions("launch [url]")
   .description("Launch the replay browser")
   .option("-b, --browser <browser>", "Browser to launch", "chromium")
+  .option(
+    "--attach <true|false>",
+    "Whether to attach to the browser process after launching",
+    false
+  )
   .allowUnknownOption()
   .action(commandLaunchBrowser);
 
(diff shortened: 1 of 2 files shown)


packages/replay/src/main.ts
```diff
@@ -711,7 +711,7 @@ async function updateMetadata({
   }
 }
 
-async function launchBrowser(browserName: BrowserName, args: string[] = []) {
+async function launchBrowser(browserName: BrowserName, detached: boolean, args: string[] = []) {
```
ryanjduffy, line 714: It's probably not used by anyone but let's add detached to the end with a default value so we don't break the API.

packages/replay/src/bin.ts
```diff
@@ -64,6 +64,11 @@ commandWithGlobalOptions("upload <id>")
 commandWithGlobalOptions("launch [url]")
   .description("Launch the replay browser")
   .option("-b, --browser <browser>", "Browser to launch", "chromium")
+  .option(
+    "--attach <true|false>",
+    "Whether to attach to the browser process after launching",
+    false
```
ryanjduffy, line 70: I think a default-false option is right but you're using `detached` in the handler. Let's use `attach` everywhere.

ryanjduffy (changes requested):

packages/replay/src/main.ts
ryanjduffy: The `attach` parameter still needs to move after `args` so the API doesn't have a breaking change.

ryanjduffy (approved):

PR #249, opened 2023-09-15
Adjust event buffering logic

ryanjduffy:
## Issue

We added an `eventBuffer` to reduce the number of tasks sent because it was causing performance issues for larger test suites. Recently, a customer's very large suite seems to be dropping events which I suspect is either caused by them coming in after the last flush or flushes that are too large and being truncated.

## Resolution

* Start and stop buffering in `beforeEach` and `afterEach`. Anything that occurs when not buffering is sent directly through the task. This should address missing events.
* Slice the buffer into 500 element arrays which are dispatched separately. This would address any truncation that could be happening with very large payloads

### DIFF
--- a/packages/cypress/src/support.ts
+++ b/packages/cypress/src/support.ts
@@ -43,6 +43,9 @@ let gLastTest: MochaTest | undefined;
 // it for retries
 let gLastOrder: number | undefined;
 
+let gBuffering = false;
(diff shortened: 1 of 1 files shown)


ryanjduffy:
The flake benchmark repo checks out with this so it seems to be safe. We'll see if it fixes the reported problem!

hbenl (approved):

PR #259, opened 2023-10-04
Add auto record

jasonLaster:
Proof of concept for what it would look like to launch chrome in an auto-record mode. The biggest missing piece is that we'd need a workflow for which recordings users would want to view.

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -1,0 +1,1 @@
+HASURA_API_KEY=<>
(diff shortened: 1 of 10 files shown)


jasonLaster:
Just pushed a commit with a new `replay play` command that launches the browser with playwright and listens for new and closed tabs. Sadly there's something wrong because new recordings are not created...

PR #264, opened 2023-10-30
Add runnerGroupId for cypress tests

ryanjduffy:
Adds `runnerGroupId` to the graphql requests to add tests to a run


packages/test-utils/src/reporter.ts
```diff
@@ -16,6 +16,7 @@ const debug = dbg("replay:test-utils:reporter");
 
 interface TestRunTestInputModel {
   testId: string;
+  runnerGroupId: string | null;
```
ryanjduffy, line 19: Get rid of this

callingmedic911 (approved):

PR #269, opened 2023-11-07
Change upload-all to upload crash reports by default

ryanjduffy:
* Changes `upload-all` to include crash reports by default
* Adds `--exclude-crashes` option

### DIFF
--- a/packages/replay/src/bin.ts
+++ b/packages/replay/src/bin.ts
@@ -20,7 +20,7 @@ import {
   MetadataOptions,
   Options,
   SourcemapUploadOptions,
-  UploadOptions,
+  UploadAllOptions,
 } from "./types";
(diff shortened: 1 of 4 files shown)


ryanjduffy:
I don't like this API because the user is explicitly choosing the uploads with `filter` and then required to explicitly exclude crashes. We could switch it to an opt-in flag (`--include-crashes`) but that defeats the purpose of uploading crashes by default.

ryanjduffy:
I've changed this to switch the flag to be opt in (`--include-crashes`) to avoid the conflict of specifying a filter to select a subset and the tool adding to that. I also added a warning so that if there are crashes but the user specifies a filter, they'll be notified that the crashes won't be uploaded and to use the new flag to include them.

hbenl (approved):

PR #273, opened 2023-11-12
Inherit `stdio` in `launch` + make `browsers` optional for `update-browsers`

Domiii:
1. Inherit `stdio` in `launch`, s.t. those who want to debug it can (else `RECORD_REPLAY_VERBOSE` is sent to Nirvana).
   * Also, on my Linux machine, not inheriting `stdio` seems to deadlock the browsers. I tried to debug by enabling it, but that just fixes it...
   * Either way, it would be good to have an option for the user to have any chance to get debugging data out and forward it to us, if they are getting stuck :woman_shrugging:
2. Don't `unref` from the child process when `attach` is enabled, and instead wait for it to finish. This way, the `replay launch` process stays in sync with the lifecycle of the runtime. (I have a reason for this change that I will demo hopefully after the weekend.)
3. Make `browsers` optional for `update-browsers`, so its easier to use for the average lazy person (like myself 👀)

### DIFF
--- a/packages/replay/src/bin.ts
+++ b/packages/replay/src/bin.ts
@@ -109,7 +109,7 @@ commandWithGlobalOptions("rm-all")
 
 commandWithGlobalOptions("update-browsers")
   .description("Update browsers used in automation.")
-  .arguments("<browsers...>")
+  .arguments("[<browsers...>]")
   .action(commandUpdateBrowsers);
 
 commandWithGlobalOptions("upload-sourcemaps")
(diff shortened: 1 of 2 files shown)


jazzdan (approved):

packages/replay/src/bin.ts
```diff
@@ -109,7 +109,7 @@ commandWithGlobalOptions("rm-all")
 
 commandWithGlobalOptions("update-browsers")
   .description("Update browsers used in automation.")
-  .arguments("<browsers...>")
+  .arguments("[<browsers...>]")
```
ryanjduffy, line 112: This is a valid change but please don't add put unrelated changes in the same PRs

ryanjduffy (approved):

PR #283, opened 2023-12-01
Encode failure message when appending replay URL

ryanjduffy:
Cypress > 10 is adding more content to the `<failure>` node including characters like `<` that need to be escaped. Our XML parser, `txml`, doesn't have a facility to re-encode content but the requirements for encoding are very basic so I've added a replacer to address the three offending characters.

### DIFF
--- a/packages/cypress/src/junit.ts
+++ b/packages/cypress/src/junit.ts
@@ -110,6 +110,10 @@ function addReplayLinkProperty(node: INode, replayUrls: string[]) {
   }
 }
 
+function escapeForXml(content: string) {
+  return content.replace(/&/g, "&amp;").replace(/>/g, "&gt;").replace(/</g, "&lt;");
+}
+
 function appendReplayUrlsToFailureNodes(node: INode, replayUrls: string[]) {
   try {
     const failures = findDescendentsByTagName(node, "failure");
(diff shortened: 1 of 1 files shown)


packages/cypress/src/junit.ts
```diff
@@ -120,8 +124,11 @@ function appendReplayUrlsToFailureNodes(node: INode, replayUrls: string[]) {
         return;
       }
 
-      failure.children[0] +=
-        "\n\n View in Replay\n" + replayUrls.map(url => ` * ${url}`).join("\n");
+      const output = `${failure.children[0]}\n\nView in Replay\n${replayUrls
+        .map(url => ` * ${url}`)
+        .join("\n")}`;
+
+      failure.children[0] = escapeForXml(output);
```
callingmedic911, line 131: If we do string replacement on whole output, then won't it also replace ampersand in URL's queryparams? Like if url is app.replay.io/recording/some-id?test=1&somethingelse=1, it'll be converted to app.replay.io/recording/some-id?test=1&amp;somethingelse=1. 🤔

callingmedic911 (changes requested):

packages/cypress/src/junit.ts
ryanjduffy: It has to be encoded for XML output. When decoded, it'll revert to a valid URL.

callingmedic911 (approved):

packages/cypress/src/junit.ts
ryanjduffy: Also, sort of funny, your comment showed exactly what happens because it converted the entity sequence back to &

PR #284, opened 2023-12-02
Pull commit message on buildkite

ryanjduffy:
Use `BUILDKITE_MESSAGE` for the commit message when triggered by a webhook (which we think is the trigger for a git push).

### DIFF
--- a/packages/replay/metadata/source.ts
+++ b/packages/replay/metadata/source.ts
@@ -71,6 +71,12 @@ function getCircleCIMergeId(env: NodeJS.ProcessEnv) {
   }
 }
 
+function getBuildkiteMessage(env: NodeJS.ProcessEnv) {
+  if (env.BUILDKITE_SOURCE === "webhook") {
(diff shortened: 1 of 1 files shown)


packages/replay/metadata/source.ts
```diff
@@ -71,6 +71,12 @@ function getCircleCIMergeId(env: NodeJS.ProcessEnv) {
   }
 }
 
+function getBuildkiteMessage(env: NodeJS.ProcessEnv) {
+  if (env.BUILDKITE_SOURCE === "webhook") {
```
callingmedic911, line 75: Just double checking, do we want to support any other sources? https://buildkite.com/docs/pipelines/environment-variables#BUILDKITE_SOURCE

callingmedic911 (approved):

packages/replay/metadata/source.ts
ryanjduffy: Starting with just `webhook` based on the conversation on Linear

PR #286, opened 2023-12-02
Add replay record command to launch and record

ryanjduffy:
Adds `replay record` command to launch and record and update `replay launch` to only launch the browser. Both now can be manually overridden by setting `RECORD_ALL_CONTENT` in the env.

### DIFF
--- a/packages/replay/src/bin.ts
+++ b/packages/replay/src/bin.ts
@@ -18,6 +18,7 @@ import {
 import {
   BrowserName,
   FilterOptions,
+  LaunchOptions,
   MetadataOptions,
   Options,
   SourcemapUploadOptions,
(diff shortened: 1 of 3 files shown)


packages/replay/src/bin.ts
```diff
@@ -197,20 +209,35 @@ async function commandUploadRecording(id: string, opts: CommandLineOptions) {
...
+  opts: Pick<CommandLineOptions, "warn" | "directory"> & LaunchOptions
 ) {
   try {
     debug("Options", opts);
 
     const browser = fuzzyBrowserName(opts.browser) || "chromium";
     assertValidBrowserName(browser);
 
-    const attach = opts.attach || false;
+    await launchBrowser(browser, [url || "about:blank"], false, { ...opts, verbose: true });
```
callingmedic911, line 220: Correct me if I'm wrong but this is a breaking change. Do you know if users reply on this command and  if we need to have any migrate path?

jasonLaster (approved):

PR #287, opened 2023-12-02
Use GH pull request SHA for commit when possible

ryanjduffy:
On GitHub actions, the `COMMIT_SHA` often points to a merge commit which isn't terribly useful. Using the PR head commit should provide a more meaningful message.


ryanjduffy:
Yup, that's better
image

PR #289, opened 2023-12-05
Sort recording log entries to ensure that createRecording is first

ryanjduffy:
* Sorts events in the recording log by type so we ensure we handle the `createRecording` first
* Refactors the recording log functions into a separate file and adds a test for this bug

### DIFF
--- a/packages/replay/src/recordingLog.ts
+++ b/packages/replay/src/recordingLog.ts
@@ -1,0 +1,253 @@
+import dbg from "debug";
+import fs from "fs";
+import path from "path";
+import { RecordingEntry } from "./types";
+import { generateDefaultTitle } from "./generateDefaultTitle";
+import { updateStatus } from "./main";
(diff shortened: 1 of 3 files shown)


packages/replay/src/recordingLog.ts
```diff
@@ -0,0 +1,253 @@
...
+  const lines = readRecordingFile(dir)
+    .map(line => {
+      try {
+        return JSON.parse(line) as RecordingLogEntry;
+      } catch {
+        return null;
+      }
+    })
+    .filter((o): o is RecordingLogEntry => o != null)
+    .sort((a, b) => RECORDING_LOG_KIND.indexOf(a.kind) - RECORDING_LOG_KIND.indexOf(b.kind));
```
ryanjduffy, line 60: Pointing out the new code since the refactoring makes it harder to see.

hbenl (approved):

PR #290, opened 2023-12-07
Bump chromium version string

ryanjduffy:
Update our hard-coded version string to 108 for chromium


callingmedic911 (approved):

callingmedic911:
I don't have a lot context around the change but if we are also using newer firefox (probably not) do we also want to bump version for firefox?

PR #297, opened 2023-12-26
Ignore replay-connect callback event

ryanjduffy:
## Issue

The plugin is adding `then` commands from our `replay-connect` callback to the `beforeAll` steps for each test. This wasn't currently visible to users because we aren't showing `beforeAll` and `afterAll` yet but it was introducing some confusion in troubleshooting SCS-1642 because there were events in places we didn't expect them to be.

## Resolution

Suppress our `handleReplayConnectResponse` handler called by `then` from being added to steps and from creating annotations.

### DIFF
--- a/packages/cypress/src/support.ts
+++ b/packages/cypress/src/support.ts
@@ -62,16 +62,19 @@ function handleReplayConnectResponse(v: unknown) {
   }
 }
 
+function isReplayConnectCallbackCommand(cmd: Cypress.EnqueuedCommand) {
+  return (
+    cmd.name === "then" && Array.isArray(cmd.args) && cmd.args[0] === handleReplayConnectResponse
+  );
+}
+
 function shouldIgnoreCommand(cmd: Cypress.EnqueuedCommand | Cypress.CommandQueue) {
   if (isCommandQueue(cmd)) {
     cmd = cmd.toJSON() as any as Cypress.EnqueuedCommand;
   }
 
-  if (
-    cmd.name === "then" &&
-    Array.isArray(cmd.args) &&
-    cmd.args[0] === handleReplayConnectResponse
-  ) {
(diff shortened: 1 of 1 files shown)


packages/cypress/src/support.ts
```diff
@@ -54,11 +54,27 @@ let gPluginServer: WebSocket | undefined;
 // user-facing
 const COMMAND_IGNORE_LIST = ["log-restore", "within-restore", "end-logGroup"];
 
+function handleReplayConnectResponse(v: unknown) {
+  if (v && typeof v === "object" && "port" in v && typeof v.port === "number") {
```
jasonLaster, line 58: purely a nit, but something like this is a bit cleaner for me

```js
if (typeof v?.port !== "number") {
  cy.log("[replay.io] Received unexpected response when connecting to plugin");
  return;
}
gServerPort = v.port;
```

packages/cypress/src/support.ts
```diff
@@ -54,11 +54,27 @@ let gPluginServer: WebSocket | undefined;
...
+    cy.log("[replay.io] Received unexpected response when connecting to plugin");
+  }
+}
+
 function shouldIgnoreCommand(cmd: Cypress.EnqueuedCommand | Cypress.CommandQueue) {
   if (isCommandQueue(cmd)) {
     cmd = cmd.toJSON() as any as Cypress.EnqueuedCommand;
   }
 
+  if (
```
jasonLaster, line 70: it'd help to have a comment here

jasonLaster (approved):
left some nits, but otherwise good

packages/cypress/src/support.ts
ryanjduffy: It is cleaner but it doesn't pass typechecking since `v` is unknown so more narrowing is required.

PR #302, opened 2024-01-18
Add a 'strict' flag to have a recording upload fail if any part of the upload fails.

klochek:


packages/replay/src/types.ts
```diff
@@ -22,6 +22,11 @@ export interface Options {
   apiKey?: string;
   verbose?: boolean;
   agent?: any;
+
+  /**
+   * Fail the recording upload if any part of the upload fails.
+   */
+  strict?: boolean;
```
jazzdan, line 29: Are there situations where we wouldn't want this behavior? I guess maybe people wouldn't appreciate their CI failing if the recording upload fails?

jazzdan (approved):
LGTM! Curious what @ryanjduffy thinks

packages/replay/src/types.ts
klochek: I have no idea why you wouldn't want this behaviour, but this is public/in the wild, and so making this anything other than opt-in could break users downstream, thus I'm adding the flag.

PR #303, opened 2024-01-19
Log when handling upload error

callingmedic911:
Closes SCS-1847.

Uses `maybeLog` instead of `debug` in `handleUploadingError` added in https://github.com/replayio/replay-cli/pull/302

### DIFF
--- a/packages/replay/src/main.ts
+++ b/packages/replay/src/main.ts
@@ -174,8 +174,13 @@ class RecordingUploadError extends Error {
   }
 }
 
-function handleUploadingError(err: string, strict: boolean, interiorError?: any) {
-  debug(err);
+function handleUploadingError(
+  err: string,
+  strict: boolean,
+  verbose?: boolean,
+  interiorError?: any
+) {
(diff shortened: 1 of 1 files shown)


ryanjduffy (approved):

klochek:
@callingmedic911 Did you test this change?  Did you see a log statement when called from devtools?  I ran into issues with that (I think dbg is swallowing console output), and so used dbg to ensure I could see anything.

callingmedic911:
@klochek No, I have marked this as draft for now but I created it to bring up to discussion with Ryan before I forget it.

If console.log are getting swallowed, that _may_ explain larger issue where a user won't see some failed logs. I'll look in it.

klochek:
> @klochek No, I have marked this as draft for now but I created it to bring up to discussion with Ryan before I forget it.
> 
> If console.log are getting swallowed, that _may_ explain larger issue where a user won't see some failed logs. I'll look in it.

Note that to see the logs from dbg, you have to have the DEBUG env var set, like so: `DEBUG=replay:cli`

PR #309, opened 2024-01-26
Fix socket host to use same logic on client and server

ryanjduffy:
Fixes the logic for selecting the websocket host to use the same default (`0.0.0.0`) and the same env variable (`CYPRESS_REPLAY_SOCKET_HOST`) to override to default.

Noting that we switch from `REPLAY_CYPRESS_SOCKET_HOST` to `CYPRESS_REPLAY_SOCKET_HOST` because Cypress requires that environment variables exposed to the spec file (and the support files) be prefixed with `CYPRESS_`. The variable is then referred to by the unprefixed name (`REPLAY_SOCKET_HOST`).

### DIFF
--- a/packages/cypress/src/server.ts
+++ b/packages/cypress/src/server.ts
@@ -20,20 +20,21 @@ export async function createServer() {
   });
 
   return new Promise<{ server: WebSocketServer; port: number }>(resolve => {
-    server.listen(
-      {
-        // Pick any available port unless set by user
-        port: process.env.REPLAY_CYPRESS_SOCKET_PORT
-          ? Number.parseInt(process.env.REPLAY_CYPRESS_SOCKET_PORT)
-          : 0,
-        // Explicitly use ipv4 unless set by user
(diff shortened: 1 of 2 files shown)


callingmedic911 (approved):
> Noting that we switch from REPLAY_CYPRESS_SOCKET_HOST to CYPRESS_REPLAY_SOCKET_HOST because Cypress requires that environment variables exposed to the spec file (and the support files) be prefixed with CYPRESS_. The variable is then referred to by the unprefixed name (REPLAY_SOCKET_HOST).

This is potentially as breaking change, right? I'm not sure if we should right away, but as we work towards quality we should look into:
- we follow semver for breaking changes
- have a changelog (do we have it? sorry if I missed it)
- really long term, not anytime soon: codemods to do breaking changes automatically

ryanjduffy:
You're right but since this is an undocumented environment variable, I'm willing to treat it as a patch fix rather than a major release item.

PR #310, opened 2024-01-29
put our package versions in the User-Agent header

toshok:
we can gain a lot of intel in server o11y by including version numbers in requests made by the cli/plugins.  

basically everywhere except auth we call `fetch()`, give it a package-specific user-agent.  `sourcemap-upload/` also takes a `userAgentAddition` argument and prepends it to its own user-agent.  `sourcemap-upload-webpack-plugin/` uses that to make sure its upload contain both package versions.

This does mean that our User-Agent header will change from the current `node-fetch/1.0 (+https://github.com/bitinn/node-fetch)` to list the cli/plugin package versions, but I think that is outweighed by the gains.

### DIFF
--- a/packages/replay/src/graphql.ts
+++ b/packages/replay/src/graphql.ts
@@ -1,5 +1,6 @@
 import dbg from "./debug";
 import fetch from "node-fetch";
+import { getUserAgent } from "./utils";
 
 const debug = dbg("replay:cli:graphql");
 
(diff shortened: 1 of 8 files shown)


jasonLaster (commented):
Smart!

toshok:
(going to add some tests to make sure things operate as expected)

toshok:
@ryanjduffy okay this is (finally) ready for review

ryanjduffy (approved):

PR #311, opened 2024-02-01
fix: remove npm from install script

styfle:
The [`install` script](https://docs.npmjs.com/cli/v10/using-npm/scripts#examples) is executed when running `pnpm add @replayio/playwright` and it can cause a failure if `npm` is missing from the PATH, or if the `packageManager` field is not `npm` (see [corepack](https://nodejs.org/api/corepack.html)).


bvaughn:
Hey @styfle thanks for posting this PR. Sorry that it seems to have fallen through the cracks.

The packages mentioned in this PR have already been updated to use `"$npm_node_execpath"` rather than assuming `npm` which should address the issue you raised (#391)

PR #316, opened 2024-02-23
Report recorder crashes correctly

ryanjduffy:
Report uploaded crashes correctly instead of showing a app.replay.io URL

### DIFF
--- a/packages/test-utils/src/reporter.ts
+++ b/packages/test-utils/src/reporter.ts
@@ -110,6 +110,10 @@ function getTestResult(recording: RecordingEntry): TestRun["result"] {
 }
 
 function getTestResultEmoji(recording: RecordingEntry) {
+  if (recording.status === "crashUploaded") {
+    return "☠️";
+  }
+
   const result = getTestResult(recording);
   switch (result) {
     case "unknown":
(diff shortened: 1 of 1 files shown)


packages/test-utils/src/reporter.ts
```diff
@@ -900,12 +904,18 @@ class ReplayReporter {
       output.push(`\n🚀 Successfully uploaded ${uploads.length} recordings:\n`);
```
callingmedic911, line 904: It'll first log "successfully uploaded..." and then later in output we'll say the recording failed to upload. That feels a bit odd if my understanding is right.

packages/test-utils/src/reporter.ts
```diff
@@ -900,12 +904,18 @@ class ReplayReporter {
...
+        if (r.status === "uploaded" || r.status === "crashUploaded") {
+          output.push(
+            `   ${getTestResultEmoji(r)} ${(r.metadata.title as string | undefined) || "Unknown"}`
+          );
+          if (r.status === "uploaded") {
+            output.push(
+              `      ${process.env.REPLAY_VIEW_HOST || "https://app.replay.io"}/recording/${r.id}\n`
+            );
+          } else {
+            output.push(`      Browser crashed while recording. Report uploaded to replay.\n`);
```
callingmedic911, line 916: "Report uploaded to replay." sounds a little ambiguous, like the user can access the report. If we can something like "Replay team have been notified", or something along the line because we are not really alerted for each crash.

callingmedic911 (changes requested):

ryanjduffy:
Good points, @callingmedic911 ! That's what I get for not checking my work from this morning before pushing. Pushed a fix that I think will be better

callingmedic911 (approved):

PR #320, opened 2024-03-12
Log a protocol error when upload fails

Andarist:
SCS-1564

I have studied code here and the code in the backend to see how things are roughly handled. I think it should be quite a strong guarantee that `msg.error` can exist and that it's of a certain shape when it's there. I left the `else` branch as-is though - just in case.

I've copy-pasted `ProtocolError` from the backend as I find it helpful to have such a custom error class in situations like this. The CLI doesn't define the codes and messages though so I decided to slim down the original and just passthrough what is received from the backend.

I looked through the error handling in the repo here and I think it could be improved. I think it would be worth cleaning this up as it would make it easier to deal with this stuff in a consistent manner. Right now a boolean result sometimes means an error, sometimes means a success, sometimes a string-as-error gets *returned*, sometimes an error is thrown. Similarly, promises are sometimes rejected with strings and sometimes with errors

### DIFF
--- a/packages/replay/src/client.ts
+++ b/packages/replay/src/client.ts
@@ -13,6 +13,31 @@ interface Callbacks {
   onError: (socket: WebSocket) => void;
 }
 
+type ErrorDataValue = string | number | boolean | null;
+type ErrorData = Record<string, ErrorDataValue>;
+type ProtocolErrorBase = {
+  code: number;
+  message: string;
+  data: ErrorData;
+};
+
+export class ProtocolError extends Error {
+  readonly protocolCode: number;
+  readonly protocolMessage: string;
+  readonly protocolData: unknown;
+
+  constructor(err: ProtocolErrorBase) {
+    super(`protocol error ${err.code}: ${err.message}`);
+    this.protocolCode = err.code;
+    this.protocolMessage = err.message;
+    this.protocolData = err.data ?? {};
+  }
+
+  toString() {
+    return `Protocol error ${this.protocolCode}: ${this.protocolMessage}`;
+  }
+}
+
 class ProtocolClient {
   socket: WebSocket;
   callbacks: Callbacks;
@@ -96,6 +121,8 @@ class ProtocolClient {
       this.pendingMessages.delete(msg.id);
       if (msg.result) {
         resolve(msg.result);
+      } else if (msg.error) {
+        reject(new ProtocolError(msg.error));
       } else {
         reject(`Channel error: ${JSON.stringify(msg)}`);
       }
--- a/packages/replay/src/main.ts
+++ b/packages/replay/src/main.ts
@@ -44,6 +44,7 @@ import {
   removeRecordingFromLog,
   addRecordingEvent,
 } from "./recordingLog";
+import { ProtocolError } from "./client";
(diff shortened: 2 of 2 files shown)


packages/replay/src/client.ts
```diff
@@ -96,6 +121,8 @@ class ProtocolClient {
       this.pendingMessages.delete(msg.id);
       if (msg.result) {
         resolve(msg.result);
+      } else if (msg.error) {
+        reject(new ProtocolError(msg.error));
```
Andarist, line 125: After a second though, it might be best to **temporarily** change this to:
```suggestion
        reject(msg.error);
```

Even though I feel like the current change is better, I can't quickly analyze how it might affect all the other callers. There is a really high chance that it wouldn't affect them negatively (or at least not in a disastrous way :P) but we never know.

WDYT?

callingmedic911: I don't think it would impact in any negative way.

packages/replay/src/main.ts
```diff
@@ -334,28 +335,38 @@ async function doUploadRecording(
...
+    }
+  } catch (err) {
+    handleUploadingError(
+      err instanceof ProtocolError ? err.protocolMessage : String(err),
       strict,
-      verbose
+      verbose,
+      err
     );
+    return null;
```
callingmedic911, line 369: The name `handleUploadingError` is misleading. Check out more context here: https://github.com/replayio/replay-cli/pull/302. In essence, it introduced a strict mode causing the process to fail on any error, even if we proceeded with some errors (like failing to upload a sourcemap). Irrespective of the strict flag, there's no point continuing if we fail to upload the recording binary.

callingmedic911 (changes requested):
Protocol changes looks great! 🚀 But don't think we should ship changes to the error handling in the upload process as is. We definitely could improve error handling but I feel that would be more broader change.

packages/replay/src/main.ts
Andarist: Thanks for the review and the pointers! I see what you are saying about `handleUploadingError` and the `strict`'s flag purpose etc. This one is still treated as a fatal error though since we immediately `return null`. It's just like you are saying - there is no point continuing if this fails (or well, we could retry or something on some specific error codes - but other than the one that the ticker is about because when the limit of per API key recordings has been reached then there is truly no point in continuing here).

I based this logic on the code that handles `initConnection` failure:
https://github.com/replayio/replay-cli/blob/1e6b7ec1a9629764b2be24760ab4e3e7ef6abb87/packages/replay/src/main.ts#L328-L331

I think they are mostly the same in terms of their severity etc. If this is already handled like this here - is there a point in hoisting this added error handler to one of the callers? One advantage of reusing `handleUploadingError` is that we automatically get a nice prefix string (`Upload failed: `) for the logged error (it's logged based on the `verbose` flag but from what I can see it's basically - almost - always set to `true` anyway).

Let me know what do you think, I don't mind hoisting it if you think it would be better.

callingmedic911: Ah, I missed the return null part.

callingmedic911 (approved):

PR #322, opened 2024-03-13
Add factories for reporter configs to improve contextual types

Andarist:
Fixes SCS-1898

The goal of the task is to improve types for `filter` so the user can use the **known** metadata more easily. 

To do that I decided to make this per-recording metadata generic. It's the `@replayio/test-utils` that is actually responsible for:
- setting this metadata on the recording~
- calling the user-provided `filter`

However, it's not the one that defines what the metadata is. So we can either thread some extra generics between those layers or... cast stuff 😉 My preference is to use the former. End users shouldn't really interact with any generic code here so it's mostly used for internal purposes (with the caveat that **technically** `@replayio/test-utils` is a public package).

In addition to that, I have added `createReplayReporterConfig` util that should help people to get access to those improved types. I've seen @ryanjduffy using `as` to get contextual typing in some configs and I think this would be a safer and easier way to do that (a safe alternative to `as` would be to use `satisfies`). 

TODO:
- [x] apply the same changes in `@replayio/cypress`

### DIFF
--- a/packages/playwright/src/index.ts
+++ b/packages/playwright/src/index.ts
@@ -1,7 +1,7 @@
 import { getPlaywrightBrowserPath, BrowserName } from "@replayio/replay";
 import { initMetadataFile } from "@replayio/test-utils";
 
-import { getMetadataFilePath } from "./reporter";
+import { ReplayPlaywrightConfig, getMetadataFilePath } from "./reporter";
 
 function getDeviceConfig(browserName: BrowserName) {
   const executablePath = getExecutablePath(browserName);
@@ -57,5 +57,9 @@ export const devices = {
     return getDeviceConfig("chromium");
   },
 };
+
+export function createReplayReporterConfig(config: ReplayPlaywrightConfig) {
+  return ["@replayio/playwright/reporter", config];
+}
 
 export { getMetadataFilePath };
--- a/packages/playwright/src/reporter.ts
+++ b/packages/playwright/src/reporter.ts
@@ -66,12 +66,18 @@ function mapTestStepHook(step: TestStep): "beforeEach" | "afterEach" | undefined
   }
 }
 
-interface ReplayPlaywrightConfig extends ReplayReporterConfig {
+type ReplayPlaywrightRecordingMetadata = {
+  title: string;
+  test: TestMetadataV2.TestRun;
+};
+
+export interface ReplayPlaywrightConfig
+  extends ReplayReporterConfig<ReplayPlaywrightRecordingMetadata> {
   captureTestFile?: boolean;
 }
 
 class ReplayPlaywrightReporter implements Reporter {
-  reporter?: ReplayReporter;
+  reporter?: ReplayReporter<ReplayPlaywrightRecordingMetadata>;
   captureTestFile = ["1", "true"].includes(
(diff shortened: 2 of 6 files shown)


packages/replay/src/types.ts
```diff
@@ -108,11 +108,11 @@ export interface SourceMapEntry {
   originalSources: OriginalSourceEntry[];
 }
 
-export interface RecordingEntry {
+export interface RecordingEntry<TMetadata extends UnstructuredMetadata = UnstructuredMetadata> {
```
Andarist, line 111: I added a default to all of the touched types to keep this change backwards compatible

packages/test-utils/src/reporter.ts
```diff
@@ -698,7 +700,7 @@ class ReplayReporter {
     recordings.forEach(rec => add(rec.id, mergedMetadata));
 
     // Re-fetch recordings so we have the most recent metadata
-    const allRecordings = listAllRecordings({ all: true });
+    const allRecordings = listAllRecordings({ all: true }) as RecordingEntry<TRecordingMetadata>[];
```
Andarist, line 703: This part is still something that I don't quite understand but I take it for granted that it's kinda how it is supposed to be here. Even though the metadata is not owned by `test-utils` - it's currently still initiated here (just a couple of lines above this, using `testMetadata.init(testRun)`, I believe). 

It's written back to disk using `add` and "prefetched" using `listAllRecordings` - so there is no actual full type-safety here hence the usage of the cast.

packages/playwright/src/reporter.ts
```diff
@@ -66,12 +66,18 @@ function mapTestStepHook(step: TestStep): "beforeEach" | "afterEach" | undefined
   }
 }
 
-interface ReplayPlaywrightConfig extends ReplayReporterConfig {
+type ReplayPlaywrightRecordingMetadata = {
```
Andarist, line 69: I'm open to suggestions for a naming scheme of those things, this one is already a mouthful 😅

packages/test-utils/src/reporter.ts
ryanjduffy: The CLI uses the `recordings.log` as the source of truth for recording status. It doesn't maintain an in-memory view of this data so when we update it on L700 above, we have to re-read it here to get the current data.

packages/playwright/src/index.ts
```diff
@@ -58,4 +58,8 @@ export const devices = {
   },
 };
 
+export function createReplayReporterConfig(config: ReplayPlaywrightConfig) {
+  return ["@replayio/playwright/reporter", config];
+}
```
ryanjduffy, line 63: This is a good pattern to use here. 🚀

packages/playwright/src/reporter.ts
ryanjduffy: I think this name is okay. We could shorten it to `PlaywrightRecordingMetadata` since it's coming from a replay package or maybe `PlaywrightReplayMetadata` if we wanted to keep `Replay` in the name.

ryanjduffy (commented):
This looks like it's on the right track. Let's adopt the same patterns for Cypress.

packages/test-utils/src/reporter.ts
Andarist: That I understand - the ownership of this type is confusing to me.

ryanjduffy (approved):

PR #323, opened 2024-03-13
Retry when `setRecordingMetadata` fails

Andarist:
SCS-1678

### DIFF
--- a/packages/replay/src/main.ts
+++ b/packages/replay/src/main.ts
@@ -212,7 +212,12 @@ async function setMetadata(
 ) {
   if (metadata) {
     try {
-      await client.setRecordingMetadata(recordingId, metadata);
+      await exponentialBackoffRetry(
+        () => client.setRecordingMetadata(recordingId, metadata),
+        e => {
+          debug("Failed to set recording metadata. Will be retried:  %j", e);
+        }
+      );
     } catch (e) {
       handleUploadingError(`Failed to set recording metadata ${e}`, strict, verbose, e);
     }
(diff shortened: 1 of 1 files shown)


packages/replay/src/main.ts
```diff
@@ -212,7 +212,12 @@ async function setMetadata(
 ) {
   if (metadata) {
     try {
-      await client.setRecordingMetadata(recordingId, metadata);
+      await exponentialBackoffRetry(
```
Andarist, line 215: I just stick to the default 5 retries here

packages/replay/src/main.ts
```diff
@@ -212,7 +212,12 @@ async function setMetadata(
...
+        () => client.setRecordingMetadata(recordingId, metadata),
```
Andarist, line 216: Note that I didn't quite test through this code as it's not easy to mock this or trigger a backend failure on purpose (I think? :P). I'll manually "mock" this with a throwaway code tomorrow morning - just to be extra sure. I don't see anything going wrong here though.

callingmedic911 (approved):

packages/replay/src/main.ts
callingmedic911: Just for thought: Do we want to retry on all errors?

Andarist: Good point. With the recent change related to introducing `ProtocolError`, I could try to be more granular here (that's the only kind of error that we should be getting here in practice). I've seen a file in the backend repo with all of the errors but it's not synced here anyhow. I think it's kinda fine for now to just retry at all times but in the future it would be good to have more control over this and assign some kind of severities or smth to particular error codes on our side.

PR #324, opened 2024-03-14
Upgrade GitHub Actions and set node version to the current LTS

Andarist:
part of SCS-2048

I plan to stack changes on top of this so each separate chunk can be reviewed/tweaked/dropped easily


.github/actions/playwright/action.yml
```diff
@@ -14,13 +14,13 @@ inputs:
 runs:
   using: composite
   steps:
-    - uses: actions/checkout@v2
-    - uses: actions/setup-node@v2
+    - uses: actions/checkout@v4
+    - uses: actions/setup-node@v4
       with:
-        node-version: "17"
-    - uses: bahmutov/npm-install@v1
```
Andarist, line 21: from what I can tell the main purpose of this action was caching - this is supposed to be builtin into `actions/setup-node` nowadays so hopefully we can just use it to simplify things

.node-version
```diff
@@ -0,0 +1 @@
+v18.19.1
```
Andarist, line 1: I don't mind what we put here, I just have picked the current LTS

.github/workflows/create-react-app-typescript-ci.yml
```diff
@@ -9,7 +9,7 @@ jobs:
     name: Playwright tests
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v4
```
Andarist, line 12: This one is a little bit redundant because `./.github/actions/playwright` already does it. That playwright action is separated for the reusability but does it even have to be? Is there any specific reason why it's not inlined here?

ryanjduffy (approved):

PR #325, opened 2024-03-14
Drop Semgrep action

Andarist:
I think this might not be used since this action reports an error on `main`:
> Error: ing configuration from Semgrep Cloud Platform  [ERROR] No config given. Run with `--config auto` or see https://semgrep.dev/docs/running-rules/ for instructions on running with a specific config

if this is meant to be kept then it should get fixed (and migrated to the official semgrep action)


ryanjduffy (approved):
cc: @jazzdan - I don't think anyone else is using semgrep any longer so this is probably fine. It hasn't worked for a while and no one noticed so it seems fine.

jazzdan:
SGTM

PR #326, opened 2024-03-14
Use Yarn to manage workspaces

Andarist:
⚠️ this might not be perfect everywhere yet - the change turned out to be little bit convoluted. I had to patch 2 packages:

- `sha-1` because the repo relies on a little bit funky publishing setup where `src` is emitted into `dist`, `package.json` is copied over there and the publish happens there. It creates subtle weirdness between how things are resolved locally and how things are resolved after the packages are published. In a way, this is quite clever - but it's also surprising and creates issues into which I have bumped here. I'm not like 100% sure why I have even bumped into this problem because I didn't change anything substantially to create this problem (just yet?). I treat it as a testament to the slight weirdness of the existing setup ;p
- `@jest/environment` - the repo relies on multiple versions of Jest and their types are conflicting

It's worth rechecking if existing scripts still work. I've tried not to break anything but who knows. I haven't used Earthly yet and I've made some changes there too so I'd like to ask for some careful eyes there.

The current version of Lerna isn't particularly happy about the usage of `patch:` protocol so this PR can't be merged before the publish story gets revamped too. It shouldn't be hard to adjust that, Lerna doesn't do anything that special when publishing. I'm not sure if yarn will handle `publishConfig.directory` though. I'd like to propose removing this particular bit of setup **anyway** - so maybe it's would just be best to do that here too. With this change I wouldn't expect yarn to hae any problems with the publishing here.

**UPDATE 03/20** Lerna got successfully ditched. This still requires changes around build and TS experience but I think they can be done in a followup PR.


packages/playwright/tsconfig.json
```diff
@@ -67,7 +67,8 @@
     "forceConsistentCasingInFileNames": true /* Disallow inconsistently-cased references to the same file. */,
 
     // Use ECMA-standard behavior for class properties.
-    "useDefineForClassFields": true
+    "useDefineForClassFields": true,
+    "skipLibCheck": true
```
Andarist, line 71: it's somewhat bizarre but the existing playwright types here (external package) have type issues in them so I had to disable `node_modules` checking here. I might investigate later if we can upgrade some versions to fix the issue but it isn't particularly important rn

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+  "typescript.tsdk": "node_modules/typescript/lib"
```
Andarist, line 2: this wont have any effect on contributors who don't manually (and explicitly) allow this to be used. I recommend everybody to select "Select TypeScript's Version..." from the command palette and choose the workspace version. It can save you from seeing a class of mismatched results between the VS Code and `tsc`.

If you open multiple projects in a single workspace... I'm not sure how that works though. I don't think VS Code is able to use different TS versions between such projects (like when u open both backend and reaply-cli in a single VS Code instance)

Earthfile
```diff
@@ -5,16 +5,16 @@ WORKDIR /usr/build
 
 build:
   COPY . .
-  RUN npm install && npm --unsafe-perm run bootstrap
-  RUN npm link ./packages/cypress/dist
+  RUN yarn && yarn run bootstrap && yarn
```
Andarist, line 8: second `yarn` run is needed here because the first one can't put symlinks into `node_modules/.bin` as those files don't exist (yet). They are created within `yarn run boostrap` so this second `yarn` acts as a fixup here. I plan to fix this later

packages/cypress/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "@replayio/cypress",
   "version": "1.7.20",
   "description": "Plugin to record your Cypress tests with Replay",
-  "main": "src/index.js",
+  "main": "dist/src/index.js",
```
Andarist, line 5: This is a pretty big change that might affect local development right now. I plan to follow up this PR with some rework in this area to address any pain points that might arise from this.

The main challenge is that we want the IDE to reflect cross-package changes immediately for the fast feedback loop. SImilarly, it would be great if tests could always use the most recent version of all packages. And for various tools to generally understand how those packages are structured to avoid running into corner cases.

The last one is addressed by using dist paths here. The setup becomes clear - we have some set of input files and we also have a set of output files. The published package looks roughly the same as the package looks locally. This also means that I actually still need to take a moment to tweak this setup and publishing step (no more publishing from an alternative directory).

The first problem is fixable using TS project references. They are inconvenient to use since one needs to declare dependencies in them manually (or generate them somehow - which I've done before). It works though.

The second thing (tests) isn't addressed by any of the above. There are 3 popular choices here:
- run a watcher in the background. The user makes a change in an input file, this gets picked up by the watcher and the output gets regenerated. It's fairly straightforward but requires an extra manual step - the user has to actually run this watcher. I think this is fine and it also lends itself to caching etc (it's not a bottleneck in this repo though)
- alias packages - we can force a tool like Jest to resolve `pkg` to just any location. So we can force `@replayio/cypress` to resolve to its `src` directory (over its `dist` that gets picked up by the default). Aliasing like this is tool-specific, it's usually a fairly easy addition to their configs though. 
- spit out some "redirecting" files in dev. I like this strategy as it's quite easy to use - but it's quirky looking and I've only used it through a tool. The tool ([Preconstruct](https://github.com/preconstruct/preconstruct)) is pretty cool and makes package authoring a breeze for me. It's Babel-based though... in a sense that it's using Babel to transpile inputs. So it's flexible but slow (it has never been an issue for me though in the context of monorepos like this one here). This repo already has certain structures and conventions and it might be best - for the time being - to avoid incorporating yet-another-new-tool into this toolchain here.

Andarist: So overall, I think that running an extra watcher would be just good enough here

package.json
```diff
@@ -1,26 +1,36 @@
...
+    "typecheck": "yarn workspaces foreach --all run typecheck",
+    "lint": "prettier --check .",
+    "release": "yarn run bootstrap && yarn workspaces foreach --all --exclude=. --topological npm publish --tolerate-republish"
   },
   "author": "",
   "license": "ISC",
   "devDependencies": {
-    "lerna": "^4.0.0",
+    "@replayio/playwright": "^1.1.16",
+    "@replayio/replay": "^0.21.4",
```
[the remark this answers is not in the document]
Andarist: Yes-ish. That's what makes `npx` calls in [Earthfile](https://github.com/replayio/replay-cli/blob/ca72e926c901f6b3f3199bdac5670f260bcc5b3f/Earthfile) to work. Alternatively, we could call those bin scripts using relative (in-monorepo) paths instead of `npx`. I assume the goal is to call the local versions of those and not the ones that are published

packages/replay/package.json
```diff
@@ -2,15 +2,11 @@
...
-  "main": "src/main.js",
+  "bin": "./dist/bin/replay.js",
+  "main": "./dist/src/main.js",
   "scripts": {
-    "if:exists": "node -e \"require('fs').existsSync(process.argv[1]) ? require('child_process').spawnSync(process.argv[2], process.argv.slice(3), {shell: true, stdio: 'inherit'}) : 0\"",
-    "if:source": "npm run if:exists -- src/main.ts",
-    "prepare": "npm run if:source -- npm run build",
-    "build": "rm -rf dist/ && tsc && chmod 755 dist/bin/* && cp package.json README.md dist/",
+    "prepare": "yarn run build",
+    "build": "rm -rf dist/ && tsc && chmod 755 dist/bin/*",
```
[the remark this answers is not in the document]
Andarist: > The build directory changes break the version command which is trying to resolve the package.json in packages/replay/src/utils.ts

Great catch and thanks for pointing this out. The easiest solution would be to temporarily bring back `package.json` copies in `dist`. I think I should get away here with package self-referencing though and I'm gonna try to use it in a moment. This relies on using `package.json#exports` but I think it should be fine here. As long as we stick to `require`-ing those `package.json`s then it won't even matter for TS. Alternatively, `find-up` could be used here but I think that the `package.json#exports` solution is gonna be the best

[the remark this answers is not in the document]
Andarist: Another solution would be to emit `lib` as a sibling to the root's `src` - instead of emitting `dist/src`. It has some implications on how `bin` is currently not in `src` and it's part of the same TS compilation. So I prefer not to do it right now but it's certainly something I'd like to simplify later too.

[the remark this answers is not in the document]
Andarist: self-referencing implemented in: https://github.com/replayio/replay-cli/pull/326/commits/614be04db0a5ee2a0141cc5aa1fdb0ad0a0e8153

Andarist:
> When I loaded this branch in VSCode, I was still encountering the same TS loading issues in which I had to reload the TS server to pick up changes made in workspace.

this gets addressed by https://github.com/replayio/replay-cli/pull/329

ryanjduffy (approved):
Let's try it! Thanks for digging into this, @Andarist

PR #327, opened 2024-03-20
Setup LaunchDarkly

callingmedic911:
Closes SCS-2035.

- Updates `commander` to the latest to have hooks support. Optional but we can initialize LD cleanly with it.
- Caches authInfo call to the backend to `${replayDir}/profile/authInfo.json`, so subsequent call with same API key doesn't need a round trip to the backend.


packages/replay/src/main.ts
```diff
@@ -683,52 +683,45 @@ async function updateMetadata({
...
-    console.error(e);
+  const recordings = listAllRecordings({ directory, filter, includeCrashes });
 
-    process.exit(1);
-  }
+  recordings.forEach(r => {
+    maybeLog(verbose, `Setting metadata for ${r.id}`);
+    add(r.id, sanitized);
+  });
 }
```
callingmedic911, line 725: I have just removed try/catch because it was using `process.exit` directly and the caller (in bin.ts) already has the same log and error handling.

Earthfile
```diff
@@ -19,7 +19,7 @@ test:
 setup:
   FROM +build
   RUN apt update && apt install xz-utils
-  RUN npx @replayio/playwright install
+  RUN npx --yes @replayio/playwright@latest install
```
Andarist, line 22: Why the change here? It should already use the version contained in the monorepo - which is like latest-latest 😅

callingmedic911: I was just trying out something. This command never exits: https://github.com/replayio/replay-cli/actions/runs/8395294816/job/22994255887?pr=327 I don't know if npx is being weird or there's something odd about the new version of node and npx.

callingmedic911: Or the version in monorepo is somehow mangled?

packages/replay/src/auth.ts
```diff
@@ -263,3 +264,128 @@ export async function maybeAuthenticateUser(options: Options = {}) {
...
+          node: {
+            id: string;
+          };
+        }[];
+      };
+    };
+  };
+
+  const encodedUserId = response?.viewer?.user?.id;
+  const encodedWorkspaceId = response?.auth?.workspaces?.edges?.[0]?.node?.id;
```
ryanjduffy, line 325: This is probably too defensive and we should check more strictly earlier so we can report on unexpected conditions.

packages/replay/src/auth.ts
```diff
@@ -263,3 +264,128 @@ export async function maybeAuthenticateUser(options: Options = {}) {
...
+  if (!authInfo) {
+    return;
+  }
+
+  if (authInfo.userId) {
+    await launchDarkly.identify({ type: "user", id: authInfo.userId });
+  } else if (authInfo.workspaceId) {
+    await launchDarkly.identify({ type: "workspace", id: authInfo.workspaceId });
+  }
+}
```
ryanjduffy, line 391: This might be better placed in the `./launchdarkly` module and export a fn from here that is responsible for reading (and possibly updating) the cache.

packages/replay/src/auth.ts
```diff
@@ -263,3 +264,128 @@ export async function maybeAuthenticateUser(options: Options = {}) {
...
+    const cache = JSON.parse(cacheJson);
+    return cache[authInfoCacheKey(key)];
+  } catch (e) {
+    debug("Failed to read auth info cache: %o", e);
+    return;
+  }
+}
+
+export async function initLDContextFromKey(options: Options = {}) {
+  const apiKey = options.apiKey ?? (await readToken(options));
```
ryanjduffy, line 368: It's also possible that the identity is in an environment variable so we should probably standardize on how we read this value to ensure we always end up with the same value.

ryanjduffy (commented):
Few things to consider but I think this is in pretty good shape. I'll leave it to you and @Andarist to get this across the line next week.

packages/replay/package.json
```diff
@@ -36,10 +36,11 @@
   "dependencies": {
     "@replayio/sourcemap-upload": "workspace:^",
     "@types/semver": "^7.5.6",
-    "commander": "^7.2.0",
+    "commander": "^12.0.0",
```
Andarist, line 39: just an observation: this changes the minimal required node version quite a bit (it's now node 18 whereas previously it was probably smth like node 10). Other dependencies might already require newer versions than node 10 though ;p and in general, I think it's probably just fine for this CLI here

packages/replay/src/launchdarkly.ts
```diff
@@ -0,0 +1,118 @@
...
+  error() {}
+  warn() {}
+  info() {}
+  debug() {}
+}
+
+class LaunchDarkly {
+  private client: LDClient;
+  private initialized = true;
+  private lastCall = Date.now();
```
Andarist, line 35: If I read correctly your comments then this should be updated in `variant`

packages/replay/src/launchdarkly.ts
```diff
@@ -0,0 +1,118 @@
...
+
+export const AnonymousFeatureProfile = {
+  type: "anonymous",
+  id: "anonymous",
+};
+
+export type FeatureProfile =
+  | typeof AnonymousFeatureProfile
+  | UserFeatureProfile
+  | WorkspaceFeatureProfile;
```
Andarist, line 23: I think it would be good to have all of those as objects or all of those as types. It's a mixed bag now and it might lead to confusion at consuming sites.

packages/replay/src/launchdarkly.ts
Andarist: if u'd be doing that then I think it would be best to use an actual debounce in `reClose` instead of checking the time manually etc.

The current logic is also not completely clear to me and I'm unsure if it's by design or not. For example, when you initialize the client you don't actually call `reClose` (so its timer is not set up) - only `identify` and `variant` lead to `reinitialize` and thus to `reClose`.

It feels to me that it would be more predictable if this timer would be tightly coupled with `this.initialized`.

packages/replay/src/launchdarkly.ts
```diff
@@ -0,0 +1,118 @@
...
+      return defaultValue;
+    }
+
+    const val = await this.client.variation(name, defaultValue);
+    return val;
+  }
+
+  public async close() {
+    await this.client.close();
+    this.initialized = false;
```
Andarist, line 114: There is a small race condition here. This `.close()` call takes some time and it doesn't resolve immediately. So while it's processing we might call `.variant()`, then I'd expect LaunchDarkly's client to maybe even drop that event and resolve `close`. That in turn, will switch our wrapper to the uninitialized state here.

I think this should be easily addressable here - we can just switch those two lines. By changing what we control *first* we maintain the internal consistency of our wrapper. From its PoV, we already disposed an actual LaunchDarkly's client as soon as our `.close()` gets called. 

In addition to that, I'd probably add `.catch(noop)` when calling their `.close()` method. We likely don't want to crash the process if something goes wrong with that. 

I don't think there is an actual risk of their client to throw anything there. That `close` method is fairly straightforward and only contains sync operations (although spread through at least 2 microtasks): https://github.com/launchdarkly/js-sdk-common/blob/0c4a8c0f10cfd22c089b875f62fc0a2e0ba6c504/src/index.js#L746

packages/replay/src/launchdarkly.ts
```diff
@@ -0,0 +1,118 @@
...
+
+    try {
+      await this.client.waitForInitialization();
+    } catch (e) {
+      debug("Failed to wait for LaunchDarkly initialization %j", e);
+      return defaultValue;
+    }
+
+    const val = await this.client.variation(name, defaultValue);
+    return val;
```
Andarist, line 109: I think it would be good to move this into the existing `try` block so we can return `defaultValue` when this one fails.

packages/replay/src/auth.ts
```diff
@@ -263,3 +264,128 @@ export async function maybeAuthenticateUser(options: Options = {}) {
...
+  };
+  await writeFile(cachePath, JSON.stringify(cache, undefined, 2), { encoding: "utf-8" });
+}
+
+async function readAuthInfoCache(key: string, options: Options = {}) {
+  try {
+    const cachePath = getAuthInfoCachePath(options);
+    const cacheJson = await readFile(cachePath, { encoding: "utf-8" });
+    const cache = JSON.parse(cacheJson);
+    return cache[authInfoCacheKey(key)];
```
Andarist, line 360: could we validate this somehow in case it got malformed on disk?

packages/replay/src/auth.ts
```diff
@@ -263,3 +264,128 @@ export async function maybeAuthenticateUser(options: Options = {}) {
...
+  } catch (e) {
+    debug("Failed to read auth info cache: %o", e);
+    return;
+  }
+}
+
+export async function initLDContextFromKey(options: Options = {}) {
```
Andarist, line 367:
```suggestion
export async function initLDContextFromApiKey(options: Options = {}) {
```

packages/replay/src/auth.ts
```diff
@@ -263,3 +264,128 @@ export async function maybeAuthenticateUser(options: Options = {}) {
...
+    };
+  };
+
+  const encodedUserId = response?.viewer?.user?.id;
+  const encodedWorkspaceId = response?.auth?.workspaces?.edges?.[0]?.node?.id;
+
+  return {
+    userId: encodedUserId ? parseId(encodedUserId) : null,
+    workspaceId: encodedWorkspaceId ? parseId(encodedWorkspaceId) : null,
+  };
```
Andarist, line 330: nit: a discriminated union would work nicely here

packages/replay/src/auth.ts
```diff
@@ -263,3 +264,128 @@ export async function maybeAuthenticateUser(options: Options = {}) {
...
+  );
+
+  if (resp.errors) {
+    throw {
+      id: "graphql-error",
+      message: resp.errors
+        .map((e: any) => e.message)
+        .filter(Boolean)
+        .join(", "),
+    };
```
Andarist, line 304: I'd prefer to always throw actual errors - `.stack` is super useful when diagnosing problems. You could introduce a `class GraphQL extends Error {}` here if it doesn't exist in the codebase yet.

packages/replay/src/auth.ts
Andarist: unless those two are not mutually exclusive? 🤔

Earthfile
Andarist: I think it should be reverted as it most likely caught a legitimate issue here. I'm still digging into this but it's some kind of a leak~ with open handles. The node process doesn't ever exit when executing `npx @replayio/playwright install` and I can repro this locally. So far I managed to narrow it down to the fact that some write streams are still open:
process._getActiveHandles() returning 2 WriteStreems

Andarist: I pushed out the fix ( https://github.com/replayio/replay-cli/pull/327/commits/a4454e3d26e9b91124ed8d4ac7fba79101525d76 ) for this here. The code itself might be freely tweaked - I just pushed out a quick fix of sorts to validate that this actually fixes the problem.

The problem was that `@replayio/playwright` was importing a util from `@replayio/replay`. That led to its whole module being initialized and that included initializing the LaunchDarkly wrapper class (and the underlying LaunchDarkly client). Even though this wasn't *actually* used by anything in this scenario it was trying to periodically flush things here using a timer loop:
https://github.com/launchdarkly/js-sdk-common/blob/0c4a8c0f10cfd22c089b875f62fc0a2e0ba6c504/src/EventProcessor.js#L163-L169

That prevented the process from exiting. The `@replayio/playwright` package didn't have any logic to close that client since it's not concerned with it at all - it just wanted to import a util and that's it ;p

packages/replay/src/auth.ts
callingmedic911: This function depends both on auth graphql calls and then calls launchdarkly. I discussed it with @Andarist felt it's more dependent on auth and decided to keep it here.

packages/replay/src/auth.ts
callingmedic911: Created getApiKey function that reads --apiKey, env vars, saved token in this order.

packages/replay/src/auth.ts
```diff
@@ -7,9 +7,20 @@ import path from "path";
...
 
+class GraphQLError extends Error {
+  constructor(message: string, public errors: any[]) {
+    const errorsMessage = errors
+      .map((e: any) => e.message)
+      .filter(Boolean)
+      .join(", ");
+    super(`${message}: ${errorsMessage}`);
+  }
+}
```
callingmedic911, line 22: Ideally, I want to throw directly in the `query` function but it's being used at a lot of places with conditional logic based on error. So, it should be a separate PR.

packages/replay/src/auth.ts
```diff
@@ -263,3 +282,115 @@ export async function maybeAuthenticateUser(options: Options = {}) {
...
+          };
+        }[];
+      };
+    };
+  };
+
+  const { viewer, auth } = response;
+
+  if (viewer?.user?.id) {
+    viewer.user.id;
```
Andarist, line 335:
```suggestion
    return viewer.user.id;
```

packages/replay/src/auth.ts
```diff
@@ -263,3 +282,115 @@ export async function maybeAuthenticateUser(options: Options = {}) {
     return false;
   }
 }
+
+async function getAuthInfo(key: string): Promise<string> {
```
Andarist, line 286: It could be nice to add somewhere a definition of the `authInfo` since it's just a string rn. This function wouldn't be the right place to do that since when we are reading it that information is quite close - I was wondering about it when reading some other parts of the code. But then... where would a good place for this be? I don't know tbh ;p 

Maybe consolidating all of this authInfo-related stuff into a module/class/object would be a good thing? Then all of the operations related to it would be close to each other and a piece of comment explaining what an authinfo is could live above it.

A piece of context why I was looking for it in the first place: I was wondering what `auth.json` and `authInfo.json` are about and how they relate to each other.

I know that everything here is already in `auth.json` file so maybe I'm just splitting hairs here as I'm just learning how things work. Feel free to treat this comment as a nit and ignore it.

packages/replay/src/auth.ts
```diff
@@ -263,3 +282,115 @@ export async function maybeAuthenticateUser(options: Options = {}) {
...
+  };
+
+  const { viewer, auth } = response;
+
+  if (viewer?.user?.id) {
+    viewer.user.id;
+  }
+
+  if (auth?.workspaces?.edges?.[0]?.node?.id) {
+    return auth.workspaces.edges[0].node.id;
```
Andarist, line 339: We lost here the information about the type of this id. I assume it's not needed right now but I wonder if it won't become important for some other things in the future. Since we are already fetching it - would it make sense to return the type alongside the id?

packages/replay/src/auth.ts
```diff
@@ -263,3 +282,115 @@ export async function maybeAuthenticateUser(options: Options = {}) {
...
+}
+
+function getAuthInfoCachePath(options: Options = {}) {
+  const directory = getDirectory(options);
+  return path.resolve(path.join(directory, "profile", "authInfo.json"));
+}
+
+function authInfoCacheKey(key: string) {
+  return createHash("sha256").update(key).digest("hex");
+}
```
Andarist, line 352: A code comment explaining this could be helpful. I **assume** that we want to hide the actual API key in the file on disk **if** the user didn't use `login` (in such a case the API key is written on disk anyway) but a different way of supplying the API key (like an env variable)

packages/replay/src/launchdarkly.ts
```diff
@@ -0,0 +1,123 @@
...
+  debug() {}
+}
+
+class LaunchDarkly {
+  private client: LDClient;
+  private initialized = true;
+  private lastCall = Date.now();
+
+  constructor() {
+    this.client = LaunchDarkly.initializeClient();
```
Andarist, line 31: One slight thing that still bothers me here is that this call want lead to setting up the reClose timer. It's only after the first call to `variant`/`identify` when we set it up. The CLI itself is shortlived and we are always meant to call either of those so it's not a big concern.

packages/replay/src/auth.ts
callingmedic911: Encoded ID has the prefix that defines the type. If we ever need it we can parse that.

packages/replay/src/auth.ts
callingmedic911: Good catch, thanks!

packages/replay/src/auth.ts
callingmedic911: For an explanation about authInfo, I see your point but right now I think it's good enough if I explain what it is right and why we need it right above `initLDContextFromApiKey`. We can look into splitting out in separate class but I don't want prematurely do it.


>  I was wondering what auth.json and authInfo.json are about and how they relate to each other

`auth.json` has a user token but not the user's ID. We could potentially save everything in a single file but I felt authInfo is more like a cache that we should be able to remove if we need it without any change in functionality and that's why kept it separate. We can revisit this if we want to consolidate these two.

packages/replay/src/launchdarkly.ts
callingmedic911: Reintialziation/reclose logic is more relevant in the backend (where I copied this code from). Removing this whole thing.

packages/replay/src/auth.ts
```diff
@@ -263,3 +282,119 @@ export async function maybeAuthenticateUser(options: Options = {}) {
...
+
+export async function initLDContextFromApiKey(options: Options = {}) {
+  const apiKey = await getApiKey(options);
+  if (!apiKey) {
+    return;
+  }
+
+  let targetId: string | undefined = await readAuthInfoCache(apiKey, options);
+  console.log("targetId: ", targetId);
```
Andarist, line 391: looks like a leftover
```suggestion
```

Andarist (approved):

packages/replay/src/auth.ts
```diff
@@ -263,3 +282,117 @@ export async function maybeAuthenticateUser(options: Options = {}) {
...
+  throw new Error("Unrecognized type of an API key: Missing both user ID and workspace ID.");
+}
+
+function getAuthInfoCachePath(options: Options = {}) {
+  const directory = getDirectory(options);
+  return path.resolve(path.join(directory, "profile", "authInfo.json"));
+}
+
+// We don't want to store the API key in plain text, especially when provided
+// via env or CLI arg. Hashing it would prevent leaking the key
```
bvaughn, line 351: We store the auth token in plain text in the `auth.json` file that's right beside the newly-aded `authInfo.json` file FWIW.

bvaughn: Still seems kind of arbitrary.

If someone has access to your disk and home directory, they could read your API key directly.

PR #330, opened 2024-03-21
Configure a `tsc -w` task

Andarist:


.vscode/tasks.json
```diff
@@ -0,0 +1,21 @@
...
+        "echo": true,
+        "reveal": "silent",
+        "focus": false,
+        "panel": "dedicated",
+        "showReuseMessage": true,
+        "clear": false
+      },
+      "runOptions": {
+        "runOn": "folderOpen"
+      }
```
Andarist, line 18: This is just a proposal. It works nicely when you open this project alone. In such a context you almost always want to have this task running. OTOH, if you have a multi-project workspace like this one:
```ts
{
  "folders": [
    {
      "path": "backend"
    },
    {
      "path": "devtools"
    },
    {
      "path": "replay-cli"
    }
  ],
}
```

Then it might not always be desirable to spin up this task. So I think I'm opting for removing this option. You will always be able to run this manually~ or through "Task: Run Task" command in VS Code. It might be less ergonomic for devs working primarily in this repo but it wont be disruptive for others.

ryanjduffy: I have several replay folders in my vscode workspace so I'd lean towards opt-in

Andarist: 👍 pushed out a change that removes this automatic setting so it will be an opt-in now

ryanjduffy (approved):

PR #331, opened 2024-03-21
Use workspace references for monorepo deps

ryanjduffy:


callingmedic911 (approved):

packages/replay/package.json
```diff
@@ -25,7 +25,7 @@
   },
   "homepage": "https://github.com/replayio/replay-cli/blob/main/packages/replay/README.md",
   "dependencies": {
-    "@replayio/sourcemap-upload": "^1.1.2",
+    "@replayio/sourcemap-upload": "workspace:^",
```
bvaughn, line 28: I added this same dependency to the `replayio` package:

https://github.com/replayio/replay-cli/blob/bae05d161b5854af33d270f9ca726be5f5e83cb7/packages/replayio/package.json#L22-L24

And it seems to be causing NPM install failures when I test a pre-release of the package via `npm pack`

When I try to install the packed package, it fails with:
```
 p install ../replayio-0.0.2.tgz       
 ERR_PNPM_WORKSPACE_PKG_NOT_FOUND  In : "@replayio/sourcemap-upload@workspace:^" is in the dependencies but no package named "@replayio/sourcemap-upload" is present in the workspace

This error happened while installing the dependencies of replayio@0.0.2

Packages found in the workspace: 
Progress: resolved 1, reused 0, downloaded 0, added 0
```

I was wondering why the CLI doesn't _also_ fail but looking at [the most recent release](https://unpkg.com/@replayio/replay@0.22.0/package.json), it seems the dependency is pinned to:
```
"@replayio/sourcemap-upload": "^2.0.0",
```

How is that being done? I must be missing something.

bvaughn: I see.

`yarn release` fails with:
```
[replayio]: Process exited (exit code 1), completed in 0s 322ms
The command failed for workspaces that are depended upon by other workspaces; can't satisfy the dependency graph
Failed with errors in 3s 556ms
```

I might _remove_ the `@replayio/sourcemap-upload` (since it's not being used yet) to unblock me to publish a 0.0.2 beta for people to use today, and we can sort this out tomorrow.

Andarist: Good idea 👍 I'll tackle this one tomorrow

Andarist: I can't repro this. The package has been packed just ok with `yarn pack` and `yarn release` in the root succeeds too. The latter is sort of a dry-run at this moment though - so it's not exactly the same situation that you were going through. I can go through the next real release to make sure that it works OK and solve any issues that might arise during the process.

bvaughn: The importance of testing via packing is much less now that the first release has gone out. I was hoping to uncover things like the "files" entry for the initial one.

PR #335, opened 2024-03-22
Add a Cypress example

Andarist:


ryanjduffy:
@callingmedic911 - hoping you can review this next week since i'll be out Mon + Tuesday

callingmedic911:
I have a limited context of internal cypress plugins but I don't think that's needed here? I'll take a stab at it on Monday.

packages/cypress/example/README.md
callingmedic911, line 1: nit: Do we want to keep this file as is, with CRA instructions?

packages/cypress/example/cypress/support/e2e.js
callingmedic911, line 1: Do we need this empty file?

callingmedic911 (approved):
Changes looks good. I have left some minor comments.

I was trying to understand motivation and scope for this change: is it to have a something within this repo to generate recordings? I know we were using this repo to do that: https://github.com/replayio-public/flake

Andarist:
> I was trying to understand motivation and scope for this change: is it to have a something within this repo to generate recordings?

It's not plugged into any particular task right now. To test changes I had to setup this up locally anyway and I decided to set it up as part of this project so others could benefit too. So the main purpose is to aid in developing and local testing.

packages/cypress/example/cypress/support/e2e.js
Andarist: we don't need it but since this example is meant for local experimentation we might just leave it as-is

packages/cypress/example/README.md
Andarist: i'll remove it

PR #338, opened 2024-03-22
Make yarn.lock immutable in CI

callingmedic911:
CI env is getting lost between Github CI -> Earthly / Docker -> yarn install step, resulting yarn continuing even if the lock file update was needed, which is the default behavior in CI envs.

Tested here: https://github.com/replayio/replay-cli/actions/runs/8394619455/job/22992164149?pr=338

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -5,7 +5,7 @@ WORKDIR /usr/build
 
 build:
   COPY . .
-  RUN yarn && yarn run bootstrap
+  RUN yarn --immutable && yarn run bootstrap
   RUN npm link --prefix ./packages/cypress
 
 lint:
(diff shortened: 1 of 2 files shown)


Earthfile
```diff
@@ -5,7 +5,7 @@ WORKDIR /usr/build
 
 build:
   COPY . .
-  RUN yarn && yarn run bootstrap
+  RUN yarn --immutable && yarn run bootstrap
```
Andarist, line 8: We also call it [here](https://github.com/replayio/replay-cli/blob/84e218d0cb05cbdbedda2e0d373ca332fbbbc9c7/.github/actions/playwright/action.yml#L22) in the example's dir. In there it probably doesn't matter but it might be best to pass it there explicitly too for consistency

Andarist (approved):

PR #340, opened 2024-03-27
Add explicit `--access public` to the `release` script

Andarist:


package.json
```diff
@@ -16,7 +16,7 @@
     "test:unit": "yarn workspaces foreach --all --exclude=. --topological run test",
     "typecheck": "yarn workspaces foreach --all run typecheck",
     "lint": "prettier --check .",
-    "release": "yarn run bootstrap && yarn workspaces foreach --no-private --all --topological npm publish --tolerate-republish"
+    "release": "yarn run bootstrap && yarn workspaces foreach --no-private --all --topological npm publish --tolerate-republish --access public"
```
Andarist, line 19: A don't think we are going to ever want to publish a restricted package from this repo and private package are already excluded from this script. Alternatively we can add `publishConfig: { access: 'public' }` to all packages to have a more granular control over this (but I don't think it's needed here)

ryanjduffy (approved):

PR #341, opened 2024-03-27
Support passing along additional browser arguments

jasonLaster:
Adds support for passing along additional browser arguments

```
replay record -a "--load-extension=~/src/_daily/Mar-22/chrome-extensions-samples/functional-samples/tutorial.hello-world/"
```

### DIFF
--- a/packages/replay/src/bin.ts
+++ b/packages/replay/src/bin.ts
@@ -91,7 +91,7 @@ commandWithGlobalOptions("launch [url]")
 commandWithGlobalOptions("record [url]")
   .description("Launch the replay browser and start recording")
   .option("-b, --browser <browser>", "Browser to launch", "chromium")
-  .option("-a --browser-args <args>", "Browser arguments", "")
+  .option("--browser-args <args>", "Browser arguments", "")
   .option(
     "--attach <true|false>",
     "Whether to attach to the browser process after launching",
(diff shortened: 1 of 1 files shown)


packages/replay/src/bin.ts
```diff
@@ -86,6 +91,7 @@ commandWithGlobalOptions("launch [url]")
 commandWithGlobalOptions("record [url]")
   .description("Launch the replay browser and start recording")
   .option("-b, --browser <browser>", "Browser to launch", "chromium")
+  .option("-a --browser-args <args>", "Browser arguments", "")
```
Andarist, line 94: re API choice - next is accepting a similar thing through `NODE_OPTIONS` env variable: https://nextjs.org/docs/pages/api-reference/next-cli . I don't have any particular opinion why it would be better though so I'm OK with the current solution

Andarist: Ok, I educated myself a little bit and it seems that it's just a standard env variable for node and it works in a very similar way as CLI arguments for node (you can pass both using both ways). I can't find any env variable like this for Chrome.

So, I think, the current proposed solution is more than fine and I would just go with it. However, I probably wouldn't allow this to be used with `-a`. This particular command *already* comes with the support for `--attach` and `-a` could be easily mistaken as a shortcut for that option.
```suggestion
  .option("--browser-args <args>", "Browser arguments", "")
```

jasonLaster:
Good call

Andarist (approved):

PR #342, opened 2024-03-28
Add extra entrypoints to allow packages to avoid pulling to many dependencies

Andarist:


packages/cypress/src/support.ts
```diff
@@ -1,5 +1,5 @@
 import type { TestMetadataV2 } from "@replayio/test-utils";
-import { buildTestId } from "@replayio/test-utils";
+import { buildTestId } from "@replayio/test-utils/testId";
```
Andarist, line 2: Allowing this is the core of the fix. 

Currently `@replayio/test-utils` depends on `@replayio/replay` so when we were importing from the root entrypoint we were also indirectly importing the whole `@replayio/replay`. That created problems for Cypress' webpack bundling

packages/replay/src/utils.ts
```diff
@@ -1,3 +1,9 @@
+// This module is meant to be somewhat browser-friendly.
```
Andarist, line 1: I tested the core of this PR with Cypress **without** changes to those utils here and it worked. So it might not be strictly needed but, for the time being, I feel more comfortable with adding this extra entrypoint for utils too.

ryanjduffy (approved):

PR #344, opened 2024-03-30
replayio: new CLI package

bvaughn:
Continued in #346

New Replay CLI (`replayio` package) with the following commands ¹
- [x] `list`: List all local recordings
  * `list` or `list --json`
- [x] `login`: Sign-in via system browser; only necessary if no API key env var is set
  - [x] Store access token _and_ refresh token (so we can regenerate expired access token)
- [x] `logout`: Sign-out; has no effect if API key env var is set
- [x] `record`: Launch Replay browser, record, and upload
  * `record` or `record <url>`
  - [x] Verify authentication (either API key or prompt for sign-in)
  - [x] Prompt user to update NPM package if a newer release is available
  - [x] Prompt user to update runtime if a newer release is available
  - [x] Launch Replay Chrome in record mode
  - [x] Prompt user to select which new recordings to upload ²
- [x] `remove`: Remove one or more recordings
  * `remove <id>` or `remove --all`
- [x] `update`: Update installed Replay runtime; (Chrome only for now)
- [ ] `upload`: Upload one or more recordings ³
  * `upload <id>` or `upload --all`
  - [ ] Process uploaded recordings ³
- [ ] `upload-source-maps`: Upload source-maps for a Workspace ³
- [x] `view`

---

### Example help output

Screenshot 2024-04-03 at 10 01 36 AM

Screenshot 2024-04-03 at 10 02 00 AM

### Example `replay list` output

Screenshot 2024-04-03 at 9 45 57 AM

### Example `replay upload` output

Screenshot 2024-04-03 at 10 03 54 AM

### Example `replay record` flow

Screenshot 2024-04-03 at 10 02 56 AM

---

¹ A complete list of commands can be seen on the Notion project page
² Pending the outcome of TT-211, we may make this step more automated
³ This functionality will be implemented in a separate follow-up PR

### DIFF
--- a/packages/replayio/src/commands/list.ts
+++ b/packages/replayio/src/commands/list.ts
@@ -1,0 +1,23 @@
+import { registerAuthenticatedCommand } from "../utils/commander";
+import { exitProcess } from "../utils/exitProcess";
+import { getRecordings } from "../utils/recordings/getRecordings";
+import { printRecordings } from "../utils/recordings/printRecordings";
+
+registerAuthenticatedCommand("list")
+  .description("List all local recordings")
+  .option("--json", "Format output as JSON")
+  .action(list);
+
+async function list({ json = false }: { json?: boolean }) {
+  const recordings = getRecordings();
+
+  if (json) {
+    console.log(JSON.stringify(recordings, null, 2));
+  } else if (recordings.length === 0) {
+    console.log("No recordings found");
+  } else {
+    console.log(printRecordings(recordings));
+  }
+
+  await exitProcess(0);
+}
--- a/packages/replayio/src/commands/logout.ts
+++ b/packages/replayio/src/commands/logout.ts
@@ -1,0 +1,19 @@
+import { getAccessToken } from "../utils/authentication/getAccessToken";
+import { logoutIfAuthenticated } from "../utils/authentication/logoutIfAuthenticated";
+import { registerCommand } from "../utils/commander";
+import { exitProcess } from "../utils/exitProcess";
+
+registerCommand("logout").description("Sign out of your Replay account").action(logout);
+
+async function logout() {
+  await logoutIfAuthenticated();
+
+  const token = await getAccessToken();
+  if (token) {
+    console.log("Cannot sign out sessions authenticated with an API_KEY");
+  } else {
+    console.log("You are now signed out");
+  }
+
+  await exitProcess(0);
+}
--- a/packages/replayio/src/commands/record.ts
+++ b/packages/replayio/src/commands/record.ts
@@ -1,0 +1,54 @@
+import { launchBrowser } from "../utils/browser/launchBrowser";
+import { registerAuthenticatedCommand } from "../utils/commander";
+import { confirm } from "../utils/confirm";
+import { exitProcess } from "../utils/exitProcess";
+import { promptForUpdate } from "../utils/installation/promptForUpdate";
+import { getRecordings } from "../utils/recordings/getRecordings";
+import { processUploadedRecordings } from "../utils/recordings/processUploadedRecordings";
+import { selectRecordings } from "../utils/recordings/selectRecordings";
+import { uploadRecordings } from "../utils/recordings/uploadRecordings";
+
+registerAuthenticatedCommand("record [url]")
+  .description("Launch the replay browser in recording mode")
+  .action(record);
+
+async function record(url: string = "about:blank") {
+  await promptForUpdate();
+
+  const recordingsBefore = await getRecordings();
+
+  await launchBrowser(url);
+
+  const recordingsAfter = await getRecordings();
+  const recordingsNew = recordingsAfter.filter(
+    recording => !recordingsBefore.some(({ id }) => id === recording.id)
+  );
+
+  console.log(""); // Spacing for readability
+
+  if (recordingsNew.length > 0) {
+    const selectedRecordings = await selectRecordings(recordingsNew, {
+      prompt: "New recording(s) found. Which would you like to upload?",
+      selectionMessage: "The following recording(s) will be uploaded:",
+    });
+
+    if (selectedRecordings.length > 0) {
+      const shouldProcess = await confirm(
+        "Would you like the selected recording(s) to be processed?"
+      );
+      if (shouldProcess) {
+        console.log("After upload, the selected recording(s) will be processed.\n");
+      }
+
+      await uploadRecordings(selectedRecordings);
+
+      if (shouldProcess) {
+        await processUploadedRecordings(selectedRecordings);
+      }
+    }
+  } else {
+    console.log("No new recordings were created");
+  }
+
+  await exitProcess(0);
+}
--- a/packages/replayio/src/commands/upload.ts
+++ b/packages/replayio/src/commands/upload.ts
@@ -1,0 +1,62 @@
+import chalk from "chalk";
+import { requireAuthentication } from "../utils/authentication/requireAuthentication";
+import { registerCommand } from "../utils/commander";
+import { confirm } from "../utils/confirm";
+import { exitProcess } from "../utils/exitProcess";
+import { getRecordings } from "../utils/recordings/getRecordings";
+import { processUploadedRecordings } from "../utils/recordings/processUploadedRecordings";
+import { selectRecordings } from "../utils/recordings/selectRecordings";
+import { LocalRecording } from "../utils/recordings/types";
+import { uploadRecordings } from "../utils/recordings/uploadRecordings";
+
+registerCommand("upload")
+  .argument("[ids]", `Recording ids ${chalk.dim("(comma-separated)")}`)
+  .option("-a, --all", "Upload all recordings")
+  .option("-p, --process", "Process uploaded recording(s)")
+  .description("Upload recording(s)")
+  .action(upload);
+
+async function upload(
+  ids: string[] | undefined,
+  {
+    all = false,
+    process: shouldProcess,
+  }: {
+    all?: boolean;
+    process?: boolean;
+  } = {}
+) {
+  await requireAuthentication(false);
+
+  const recordings = await getRecordings();
+
+  let selectedRecordings: LocalRecording[] = [];
+  if (ids && ids.length > 0) {
+    selectedRecordings = recordings.filter(recording => ids.includes(recording.id));
+  } else if (all) {
+    selectedRecordings = recordings;
+  } else {
+    selectedRecordings = await selectRecordings(recordings, {
+      maxRecordingsToDisplay: 10,
+      prompt: "Which recordings would you like to upload?",
+      selectionMessage: "The following recording(s) will be uploaded:",
+    });
+  }
+
+  if (selectedRecordings.length > 0) {
+    if (shouldProcess == null) {
+      shouldProcess = await confirm("Would you like the selected recording(s) to be processed?");
+      if (shouldProcess) {
+        console.log("After upload, the selected recording(s) will be processed.\n");
+      }
+    }
+
+    await uploadRecordings(selectedRecordings);
+
+    if (shouldProcess) {
+      await processUploadedRecordings(selectedRecordings);
+    }
+  }
+
+  await exitProcess(0);
+}
--- a/packages/replayio/src/commands/view.ts
+++ b/packages/replayio/src/commands/view.ts
@@ -1,0 +1,14 @@
+import { spawn } from "child_process";
+import { registerCommand } from "../utils/commander";
+import { exitProcess } from "../utils/exitProcess";
+import { getSystemOpenCommand } from "../utils/getSystemOpenCommand";
+
+registerCommand("view <id>").description("Upload one or more recordings").action(view);
+
+async function view(id: string) {
+  const url = `https://app.replay.io/recording/${id}`;
+
+  spawn(getSystemOpenCommand(), [url]);
+
+  await exitProcess(0);
+}
--- a/packages/replayio/src/utils/authentication/getAccessToken.ts
+++ b/packages/replayio/src/utils/authentication/getAccessToken.ts
@@ -1,0 +1,72 @@
+import { readFromCache, writeToCache } from "../cache";
+import { updateLaunchDarklyCache } from "../launch-darkly/updateLaunchDarklyCache";
+import { maskString } from "../maskString";
+import { cachedAuthPath } from "./config";
+import { debug } from "./debug";
+import { refreshAccessTokenOrThrow } from "./refreshAccessTokenOrThrow";
+import { CachedAuthDetails } from "./types";
+
+export async function getAccessToken(): Promise<string | undefined> {
+  if (process.env.REPLAY_API_KEY) {
+    debug("Using token from env (REPLAY_API_KEY)");
+    return process.env.REPLAY_API_KEY;
+  } else if (process.env.RECORD_REPLAY_API_KEY) {
+    debug("Using token from env (RECORD_REPLAY_API_KEY)");
+    return process.env.RECORD_REPLAY_API_KEY;
+  }
+
+  let { accessToken, refreshToken } = readFromCache<CachedAuthDetails>(cachedAuthPath) ?? {};
+  if (typeof accessToken !== "string") {
+    debug("Unexpected accessToken value: " + accessToken);
+    return;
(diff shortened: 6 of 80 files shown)


packages/replayio/src/utils/recordings/getRecordings.ts
```diff
@@ -0,0 +1,170 @@
...
+            if (entry.metadata?.uri) {
+              let host = metadata.uri;
+              if (host && typeof host === "string") {
+                try {
+                  const url = new URL(host);
+                  host = url.host;
+                } finally {
+                  recording.metadata.host = host;
+                }
+              }
```
Andarist, line 42: I think I understand how you have used `finally` here but it looked a little bit tricky and took me a moment to recheck this control flow. Perhaps it would be more clear with a more standard try/catch?
```suggestion
            if (metadata?.uri) {
              if (typeof metadata?.uri === "string") {
                try {
                  const url = new URL(metadata?.uri);
                  recording.metadata.host = url.host;
                } catch (e) {
                  recording.metadata.host = metadata.uri;
                }
              }
```

packages/replayio/src/commands/upload.ts
```diff
@@ -0,0 +1,62 @@
...
+      if (shouldProcess) {
+        console.log("After upload, the selected recording(s) will be processed.\n");
+      }
+    }
+
+    await uploadRecordings(selectedRecordings);
+
+    if (shouldProcess) {
+      await processUploadedRecordings(selectedRecordings);
+    }
```
Andarist, line 58: For perf reasons it could be good to merge those 2 so each uploaded recording could be processed immediately - this could help especially when uploading a lot of recordings at once (although perhaps that's just not a super common scenario when working with the CLI)

packages/replayio/src/commands/upload.ts
```diff
@@ -0,0 +1,62 @@
...
+    process?: boolean;
+  } = {}
+) {
+  await requireAuthentication(false);
+
+  const recordings = await getRecordings();
+
+  let selectedRecordings: LocalRecording[] = [];
+  if (ids && ids.length > 0) {
+    selectedRecordings = recordings.filter(recording => ids.includes(recording.id));
```
Andarist, line 35: It would be nice to warn the user that one of the specified IDs couldn't be found

packages/replayio/src/commands/view.ts
```diff
@@ -0,0 +1,14 @@
+import { spawn } from "child_process";
+import { registerCommand } from "../utils/commander";
+import { exitProcess } from "../utils/exitProcess";
+import { getSystemOpenCommand } from "../utils/getSystemOpenCommand";
+
+registerCommand("view <id>").description("Upload one or more recordings").action(view);
+
+async function view(id: string) {
+  const url = `https://app.replay.io/recording/${id}`;
```
Andarist, line 9: We'll have to make this (and similar places) configurable somehow to aid testing against cloud dev.

packages/replayio/src/utils/getSystemOpenCommand.ts
```diff
@@ -0,0 +1,10 @@
+export function getSystemOpenCommand() {
+  switch (process.platform) {
+    case "darwin":
+      return "open";
+    case "linux":
+      return "xdg-open";
+    default:
+      throw new Error("Unsupported platform");
```
Andarist, line 8: brb, I'm calling Mark ☎️ 😉  ChatGPT suggests that `start` could work in case of `win32` but I guess that maybe we don't support windows chromium atm?

packages/replayio/src/commands/view.ts
Andarist: I found that you already have `replayAppHost` here so this could be used here

packages/replayio/src/utils/authentication/getAccessToken.ts
```diff
@@ -0,0 +1,72 @@
...
+  } else if (process.env.RECORD_REPLAY_API_KEY) {
+    debug("Using token from env (RECORD_REPLAY_API_KEY)");
+    return process.env.RECORD_REPLAY_API_KEY;
+  }
+
+  let { accessToken, refreshToken } = readFromCache<CachedAuthDetails>(cachedAuthPath) ?? {};
+  if (typeof accessToken !== "string") {
+    debug("Unexpected accessToken value: " + accessToken);
+    return;
+  } else if (typeof refreshToken !== "string") {
```
Andarist, line 22: We still use `refreshToken` unconditionally later on in this function so we should check both here

```suggestion
  }
  if (typeof refreshToken !== "string") {
```

packages/replayio/src/utils/authentication/refreshAccessTokenOrThrow.ts
```diff
@@ -0,0 +1,40 @@
...
+  const json: any = await resp.json();
+
+  if (json.error) {
+    debug("OAuth token request failed: %O", json.error);
+
+    throw {
+      id: "auth0-error",
+      message: json.error,
+      refreshToken: maskString(refreshToken),
+    };
```
Andarist, line 27: I was afraid to make this change in the old CLI due to lack of tests but it would be really cool to throw actual errors here since we have a fresh start anyway :p

packages/replayio/src/utils/prompt/prompt.ts
```diff
@@ -0,0 +1,38 @@
+import { readFromCache, writeToCache } from "../cache";
+import { promptHistoryPath } from "./config";
+import { PromptHistory } from "./types";
+
+export async function prompt(id?: string): Promise<boolean> {
+  return new Promise(resolve => {
+    const stdin = process.stdin;
+    const wasRaw = stdin.isRaw;
```
Andarist, line 8: nit: `wasRaw` suggests that it we know it has been `true` - maybe `oldRaw`/`prevRaw` would fit this better?
```suggestion
    const oldRaw = stdin.isRaw;
```

packages/replayio/src/utils/prompt/prompt.ts
```diff
@@ -0,0 +1,38 @@
...
+        case "\r":
+          resolve(true);
+          break;
+        default:
+          resolve(false);
+          break;
+      }
+    }
+
+    stdin.on("data", onData);
```
Andarist, line 36: TODO: it would be great to recheck if exit signals etc are still handled correctly with this. I'm not familiar with the in and outs of those node APIs so this very well might be already working alright - just a thought about what might be worth rechecking here.

packages/replayio/src/utils/graphql/queryGraphQL.ts
```diff
@@ -0,0 +1,31 @@
...
+    options.headers.Authorization = `Bearer ${apiKey.trim()}`;
+  }
+
+  debug("Querying %s over %s graphql endpoint", name, graphQLServer);
+  const result = await fetch(`${graphQLServer}/v1/graphql`, options);
+
+  const json: any = await result.json();
+  debug("GraphQL Response: %O", json);
+
+  return json;
```
Andarist, line 30: what do you think about always throwing `GraphQLError` here when `json.errors` are present?

packages/replayio/src/utils/launch-darkly/close.ts
```diff
@@ -0,0 +1,13 @@
+import { debug } from "./debug";
+import { getLaunchDarklyClient } from "./getLaunchDarklyClient";
+
+export async function close() {
+  const client = getLaunchDarklyClient(false);
```
Andarist, line 5: smart

packages/replayio/src/utils/installation/types.ts
```diff
@@ -0,0 +1,24 @@
+export type Executable = "darwin:chromium" | "linux:chromium" | "win32:chromium";
+export type Platform = "macOS" | "linux" | "windows";
+
+// This CLI only supports Chromium for the time being
+export type Runtime = "chromium" | "gecko" | "node";
```
Andarist, line 5: It won't ever support gecko, right? 
```suggestion
export type Runtime = "chromium" | "node";
```

packages/replayio/src/utils/getReplayPath.ts
```diff
@@ -0,0 +1,16 @@
+import assert from "assert";
+import { join, resolve } from "path";
+
+export function getReplayPath(...path: string[]) {
+  let basePath;
+  if (process.env.RECORD_REPLAY_DIRECTORY) {
+    basePath = process.env.RECORD_REPLAY_DIRECTORY;
```
Andarist, line 7: I was told before that everything with `RECORD_REPLAY_*` is considered to be legacy-ish. I guess that this one isn't often used by the user but it would be cool to have a consistent naming scheme for all of those options-like env vars on our side

packages/replayio/src/utils/date.ts
```diff
@@ -0,0 +1,47 @@
+import differenceInCalendarDays from "date-fns/differenceInCalendarDays";
+import differenceInMinutes from "date-fns/differenceInMinutes";
+import differenceInMonths from "date-fns/differenceInMonths";
+import differenceInWeeks from "date-fns/differenceInWeeks";
+import differenceInYears from "date-fns/differenceInYears";
+import padStart from "lodash.padstart";
```
Andarist, line 6: native `''.padStart` is supported since node v8 :)

packages/replayio/src/utils/authentication/requireAuthentication.ts
```diff
@@ -0,0 +1,115 @@
...
+
+  return data.closeAuthRequest.token as string;
+}
+
+async function pollForAuthentication(key: string) {
+  const timeout = setTimeout(() => {
+    debug("Timed out waiting for authentication");
+
+    exitProcess(1);
+  }, 60_000);
```
Andarist, line 93: This exits a little bit "out of band". I'd expect the `exitProcess` to be always be called within the command's implementation. To address this we could create a `Promise.race` and use a regular `throw` when the timeout "wins"

Andarist (approved):
Great stuff. This is much easier to follow than the old CLI and I'm excited about this :)

packages/replayio/src/commands/list.ts
```diff
@@ -0,0 +1,23 @@
+import { registerAuthenticatedCommand } from "../utils/commander";
+import { exitProcess } from "../utils/exitProcess";
+import { getRecordings } from "../utils/recordings/getRecordings";
+import { printRecordings } from "../utils/recordings/printRecordings";
+
+registerAuthenticatedCommand("list")
+  .description("List all local recordings")
```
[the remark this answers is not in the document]
bvaughn: Good question, and I was not 100% on this. The reason I decided to require auth for the `list` command was that...you can't do anything interesting with the data files unless you're authenticated. (You can't upload or view them.) I guess you could _remove_ them so maybe that's a reason to relax the constraint. I don't feel strongly about this.

packages/replayio/src/commands/logout.ts
```diff
@@ -0,0 +1,19 @@
...
+import { exitProcess } from "../utils/exitProcess";
+
+registerCommand("logout").description("Sign out of your Replay account").action(logout);
+
+async function logout() {
+  await logoutIfAuthenticated();
+
+  const token = await getAccessToken();
+  if (token) {
+    console.log("Cannot sign out sessions authenticated with an API_KEY");
```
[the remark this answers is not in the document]
bvaughn: I like this suggestion.

packages/replayio/src/utils/withResolvers.ts
```diff
@@ -0,0 +1,38 @@
...
+  return {
+    promise: new Promise<Type>((...args) => {
+      resolve = args[0];
+      reject = args[1];
+
+      if (rejected) {
+        reject(rejectedReason);
+      } else if (resolved) {
+        resolve(resolvedValue as Type);
+      }
```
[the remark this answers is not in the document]
bvaughn: Yeah that's a good catch. I'm going to just delete this since I ended up using the "deferred" approach instead.

packages/replayio/src/commands/logout.ts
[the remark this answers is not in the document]
bvaughn: Really like this idea.

Screenshot 2024-04-05 at 9 58 33 AM

packages/replayio/src/utils/browser/launchBrowser.ts
```diff
@@ -0,0 +1,46 @@
...
+  };
+
+  debug(
+    `Launching browser: ${runtimeExecutablePath} with args:\n`,
+    args.join("\n"),
+    "\n",
+    processOptions
+  );
+
+  console.log(`Recording ${chalk.dim("(quit browser to continue)")}`);
```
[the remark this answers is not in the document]
bvaughn: Simplifying decision: I'm just going to remove that option for now and we can revisit this when we get to the Test Suites integration.

packages/replayio/src/utils/recordings/getRecordings.ts
bvaughn: Yeah! Haha. I thought it looked weird when I went back and read it as well, so I changed it in the next PR already.

```
let host = metadata.uri;
if (host && typeof host === "string") {
  try {
    recording.metadata.host = new URL(host).host;
  } catch (error) {
    recording.metadata.host = host;
  }
}
```

packages/replayio/src/commands/upload.ts
bvaughn: Yeah, I agree that would be better. Going to drop a "TODO" there for now but I like your thinking.

packages/replayio/src/commands/record.ts
```diff
@@ -0,0 +1,54 @@
...
+  await launchBrowser(url);
+
+  const recordingsAfter = await getRecordings();
+  const recordingsNew = recordingsAfter.filter(
+    recording => !recordingsBefore.some(({ id }) => id === recording.id)
+  );
+
+  console.log(""); // Spacing for readability
+
+  if (recordingsNew.length > 0) {
```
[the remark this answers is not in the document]
Andarist: I think the logs are fine but I was thinking that `selectRecordings` step could just be replaced in this case with "do you want to upload the recording" prompt. I didn't yet go through manual Q/A so maybe this doesn't feel bad in practice - I just imagine that there might be one extra step here than needed in this scenario

packages/replayio/src/commands/view.ts
bvaughn: Yup! This has already been made testable via the `REPLAY_APP_SERVER` var in the next branch.

packages/replayio/src/utils/getSystemOpenCommand.ts
bvaughn: Yeaaah TBH I don't know. The old CLI threw in this case, so I carried the behavior over. I guess I could toss `start` in here though and at least we'd be one step closer.

packages/replayio/src/utils/authentication/refreshAccessTokenOrThrow.ts
bvaughn: Yeah! Agreed.

This was almost the first new part that I wrote, and I wasn't sure if there was a larger reason why we were throwing these weird object values.

packages/replayio/src/utils/graphql/queryGraphQL.ts
bvaughn: I _think_ I lean towards letting the caller handle `response.error` 🤔  because that seems to be more idiomatic with other GraphQL uses I've seen, though admittedly I'm not super experienced with GraphQL so I could be convinced otherwise!

packages/replayio/src/utils/date.ts
bvaughn: Good point!

bvaughn:
PR feedback: https://github.com/replayio/replay-cli/pull/346/commits/e5a061f41fbc8327d7dcd86e2fb0ee2264f7acda

packages/replayio/src/commands/record.ts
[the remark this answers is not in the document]
bvaughn: Ah, I misunderstood what you were suggesting.

Yeah, maybe! I'm going to sit on that and think about it a bit more. Going to be iterating on this particular part now that the new RUN "root" vs other types has shipped too.

bvaughn:
Merged via #346

PR #345, opened 2024-04-03
Add a custom debug wrapper package

Andarist:
This PR creates a package that we could use to address [the referenced ticket](https://linear.app/replay/issue/SCS-1865).

The idea is we can create a single instance of `debug` (the wrapped one) and pass it around as part of "context" so every caller has access to it. This introduces a lot of noise here and manual plumbing but it has some advantages:
- **everything** within the CLI has access to the same `debug` instance and we avoid blocking ourselves synchronously on writes so it should be slightly faster than the current approach
- we can *inject* a different instance of this `debug` instance when calling exported utils externally (like from the cypress/playwright plugins). This ensures that the final log won't refer to confusing namespaces (why the playwright log would refer to `cli` namespace?)
- it becomes easier to test this stuff out
- we can extend `ctx` with extra things in the future

The new `ctx` argument is threaded as the first argument to the internal function. However, to avoid breaking changes, it's added as the last - optional - one in all functions that are actually exported and utilized by the Playwright and Cypress plugins. Since those outermost `ctx` parameters are optional, I use optional chaining to get access to the current `debug` instance in all "affected" inner layers. We could use a default arguments there but in reality, the goal is to always provide something there. The optionality just ensures backward compatibility. All of those optional chains are quite greppable so it should be straightforward to make this required in the future.

With the upcoming CLI rewrite, we should be able to address the above and unify this further. I think there is still some room for improvement around logging, levels and verbosity. This package could absorb those responsibilities in the near future too.

- [ ] QA this against Cypress/Playwright after the initial review


packages/dumpable-debug/package.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "@replayio/dumpable-debug",
```
Andarist, line 2: TODO: this needs a better name 😅

packages/replay/src/launchdarkly.ts
```diff
@@ -1,7 +1,7 @@
-import dbg from "./debug";
+import weakMemoize from "@emotion/weak-memoize";
```
Andarist, line 1: it's a fairly small package. We could inline it. I mostly have resorted to using the `WeakMap`-based memoization for now since this module acts as a singleton. With the proposed change and relying on dependency injection, we can no longer assume that it's just a singleton. 

It could be nice to move LauncDarkly's initialization closer to the `ctx` creation and just pass it around everywhere with it. This would address the issue and remove the need for this package (or its equivalent). Since the PR already touches a lot of files I decided to stay way from changing this as part of this PR

packages/replay/src/install.ts
```diff
@@ -1,15 +1,12 @@
...
 import { spawnSync } from "child_process";
-import dbg from "./debug";
 import fs from "fs";
 import https from "https";
 import path from "path";
-import { BrowserName, Options } from "./types";
+import { BrowserName, Ctx, Options } from "./types";
 import { defer, getDirectory, maybeLog } from "./utils";
 
-const debug = dbg("replay:cli:install");
```
Andarist, line 11: In a couple of files here the extended namespace was lost. Those modules are not objects into which I could shove `this._debug`. They export "unrelated" functions/utils. So to avoid the hassle of addressing this right now I decided to avoid some of the namespaces like this as I don't think they were that important. We could bring them back using weakmap-based caching or by refactoring those modules so they gain "instance scope" in which we could keep the extended debug's instance.

packages/replay/src/debug.ts
```diff
@@ -1,62 +0,0 @@
...
-}
-
-let size = 0;
-export default function debug(namespace: string, pathToLog: string = logPath) {
-  size = Math.max(size, namespace.length);
-  const d = dbg(namespace);
-
-  if (process.env.REPLAY_CLI_DISABLE_LOG) {
-    return d;
-  }
```
Andarist, line 35: I didn't reimplement this one anywhere now. I wonder if there are any big concerns about it. It's not hard to add this but it's very much a "hidden feature" right now. Do we need it?

ryanjduffy (commented):
I left a couple comments but I didn't get through it all to leave an official review. One general thought I wanted to share was if we should consider adding the logger or the context to the default `Options` (the node API options, not the CLI `CommandLineOptions`) so we didn't have to add a new parameter in different argument positions throughout the codebase. This might not make sense and i wasn't able to validate it today but wanted to share the thought.

I do think there are some conflicts with @bvaughn 's branch so you may want to check in.

bvaughn:
> I do think there are some conflicts with @bvaughn 's branch so you may want to check in.

My branch is all new code, so there shouldn't be any conflicts (outside of maybe the `yarn.lock` file)

That being said, I'm not sure it's worth us making major any kind of substantial changes to the "legacy" CLI at this point. What's the motivation for this?

Andarist:
I got tasked with this (and started working on this) before I even knew the new CLI is coming. I continued on this work yesterday as a spike~ and to learn more about different pieces and what kind of logging is even happening here.

I agree this might just not be worth it to pull into the legacy CLI at the moment. Like we talked yesterday - even if the first release of the new CLI will focus on the user experience etc... it shouldn't take a lot of effort to reorganize it slightly later so playwright and cypress plugins could build on top of the new pieces. I think that would be the right time to introduce this change.

bvaughn:

PR #346, opened 2024-04-03
replayio: new CLI package [continued]

bvaughn:
Builds on top of the work done in #344

UX:
- [x] Both `upload` and `remove` commands default to interactive mode (where you select a list of recordings)
- [x] Show in-progress status

Implement the following commands:
- [x] Upload recording(s)
  - [x] Direct
  - [x] Multi-part
- [ ] Process recording(s)
- [ ] Upload source-maps

Cleanup:
- [ ] Add more unit/e2e tests (and make sure CI is running them)
- [ ] Consolidate `chalk.*` color references into a `theme` module


packages/replayio/src/utils/createDeferred.ts
```diff
@@ -0,0 +1,94 @@
...
+  let rejectPromise: (error: Error) => void;
+  let resolvePromise: (value: Type | PromiseLike<Type>) => void;
+
+  const promise = new Promise<Type>((resolve, reject) => {
+    rejectPromise = reject;
+    resolvePromise = resolve;
+  });
+  promise.catch(() => {
+    // Prevent unhandled promise rejection warning.
+  });
```
Andarist, line 48: Any particular reason why we want to prevent this here? I usually consider it a good thing that it warns/crashes the process when promises are unhandled. Fail fast and fix things and all of that.

Andarist, line 48: Any particular reason why we want to prevent this here? I usually consider it a good thing that it warns/crashes the process when promises are unhandled. Fail fast and fix things and all of that.

packages/replayio/src/utils/recordings/createdDeferredAction.ts
```diff
@@ -0,0 +1,19 @@
+import { createDeferred } from "../createDeferred";
+import { debug } from "./debug";
+
+export function createdDeferredAction<Data>(data: Data, promise: Promise<void>) {
```
Andarist, line 4: this is somewhat similar to `Promise.allSettled`. WDYT about `createSettledDeferred`?

packages/replayio/src/utils/recordings/upload/uploadRecording.ts
```diff
@@ -0,0 +1,199 @@
...
+          });
+          return uploadPart(url, { recordingPath, start, end }, end - start + 1);
+        },
+        error => {
+          debug(`Failed to upload part ${index + 1}. Will be retried: %o`, error);
+        },
+        10
+      );
+    },
+    { concurrency: 10 }
```
Andarist, line 175: the legacy CLI was also setting some concurrency limits within `uploadAllRecordings`. If I'm not mistaken we lost that in the new implementation. I'm not sure what a threshold for problems is but perhaps this is something to address at some point

packages/replayio/src/utils/recordings/types.ts
```diff
@@ -13,19 +13,23 @@ export enum RECORDING_LOG_KIND {
...
+// This data primarily comes from the runtime
+// The CLI adds some entries as well, based on upload status
 export type LogEntry = {
   buildId?: string;
   data?: any;
   driverVersion?: string;
   id: string;
   kind: RECORDING_LOG_KIND;
-  metadata?: {
+  metadata?: Object & {
```
Andarist, line 26: `Object` type is *rarely* a good fit - it corresponds to the actual `Object`'s prototype. I've removed this locally to see where the CLI wouldn't compile but everything is OK without it so maybe we can just remove this?

packages/replayio/src/utils/recordings/readRecordingLogLines.ts
```diff
@@ -0,0 +1,11 @@
+import { readFileSync } from "fs-extra";
+import { recordingLogPath } from "./config";
+
+export function readRecordingLogLines() {
+  const rawText = readFileSync(recordingLogPath, "utf8");
+  return rawText
+    .replace(/\}\{/g, "}\n{")
```
Andarist, line 7: do u know when a new line might not be inserted in between rows when writing to this log file?

packages/replayio/src/utils/recordings/printRecordings.ts
```diff
@@ -13,40 +12,9 @@ export function printRecordings(
...
-            break;
-          case "unusable":
-            status = "Unusable";
-            break;
-        }
-      }
-      columns.push(status);
-
-      return columns;
+      return [id, title, date, duration, status];
```
Andarist, line 17: nice, i like how it now is easy to relate rows and headers here :)

packages/replayio/src/utils/recordings/printDeferredRecordingActions.ts
Andarist, line 1: nice work on this one 👍

packages/replayio/src/utils/protocol/ProtocolClient.ts
```diff
@@ -0,0 +1,123 @@
...
+  private deferredAuthenticated = createDeferred<boolean>();
+  private eventListeners = new Map();
+  private nextMessageId = 1;
+  private pendingMessages: Map<number, Deferred<any>> = new Map();
+  private socket: WebSocket;
+
+  constructor({ agent }: { agent?: Agent } = {}) {
+    debug("Creating WebSocket for %s with %o", replayServer, { agent });
+
+    this.socket = new WebSocket(replayServer, {
```
Andarist, line 21: it would be great to make this websocket reconnectable - it's somewhat easy right now to succeed with an upload through retries just to fail on an already failed WS connection

packages/replayio/src/utils/createDeferred.ts
```diff
@@ -0,0 +1,94 @@
+export const STATUS_NOT_FOUND = "not-found";
+export const STATUS_PENDING = "pending";
+export const STATUS_ABORTED = "aborted";
+export const STATUS_REJECTED = "rejected";
+export const STATUS_RESOLVED = "resolved";
```
Andarist, line 5: `'not-found'` and `'aborted'` aren't used anywhere - can they be removed?

packages/replayio/src/index.ts
```diff
@@ -12,3 +13,13 @@ import "./commands/upload-source-maps";
...
+
+// If the process is terminated by CTRL+C while waiting for an async function
+// avoid ERR_UNHANDLED_REJECTION from being printed to the console
+process.on("uncaughtException", async error => {
+  if (error.name !== "UnhandledPromiseRejection") {
+    console.error("uncaughtException:", error);
+  }
+
+  await exitProcess(1);
+});
```
Andarist, line 25: isn't the fact that we get `UnhandledPromiseRejection` a sign that some promise is truly dangling unhandled somewhere?

Andarist (approved):

packages/replayio/src/commands/login.ts
```diff
@@ -7,7 +8,12 @@ registerCommand("login")
   .action(login);
 
 async function login() {
-  await requireAuthentication(true);
+  let savedAccessToken = await getAccessToken();
+  if (savedAccessToken) {
+    console.log("You are already signed in!");
+  } else {
+    requireAuthentication();
```
bvaughn, line 15: Lol oops

packages/replayio/src/utils/createDeferred.ts
bvaughn: I think it's bad UX for common case– like for instance, if we're prompting you if you'd like to upload a new recording and you CTL+C quit, I don't think we should log a long error stack trace.

packages/replayio/src/utils/recordings/readRecordingLogLines.ts
bvaughn: Nope, but I've seen it happen several times. I suspect something the runtime is doing doesn't jive with what this CLI is doine.

packages/replayio/src/index.ts
bvaughn: We talked about this above, but it's _expected_ in cases like prompt/confirmation or select-a-list-of-recordings.

packages/replayio/src/utils/protocol/ProtocolClient.ts
bvaughn: Sure, but going to defer this for now -> PRO-82

packages/replayio/src/utils/recordings/upload/uploadRecording.ts
bvaughn: PRO-83

packages/replayio/src/utils/recordings/createdDeferredAction.ts
bvaughn: I don't think this is actually very similar to `allSettled` so I think the rename probably wouldn't make sense. (We can keep chatting about this though if you think I'm wrong!)

bvaughn:
Thanks a ton for the thorough review on this, @Andarist. I'm going to merge this directly into `main` (it has feedback from both #344 and this PR) and I'll continue to iterate on the new CLI in a follow up PR.

packages/replayio/src/utils/createDeferred.ts
Andarist: Yeah, I wouldn't like this to be printed to the user. It just seems that maybe the fact that it's printed is a sign that some promise stays unhandled and that it should be fixed elsewhere?

packages/replayio/src/utils/recordings/createdDeferredAction.ts
Andarist: I see the "settled" part to be quite similar here. `createDeferred` gets some mutable data, creates the promise and gives the user control over its resolution etc. This is `Promise.withResolvers` with some extra things on top of it.

The `createdDeferredAction` wraps the above with one extra thing - it eliminates the rejected state altogether from your async action. We want to be able to track failure and all but we don't want to fail hard on failures. This is quite similar to the "settled" part - it just uses a boolean to denote its result rather than `{ status: 'fulfilled', value: T } | { status: 'rejected', reason: unknown }`. We could easily use the "standardized" shape instead of a boolean here - but both are fine.

One extra thing is that this still gives the caller control over the deferred itself. This calls `.resolve`/`.reject` but both are still returned to the caller and they could, accidentally, call them and settle at some other time.

Anyway, I'm just saying where I've seen the similarity here.

packages/replayio/src/utils/recordings/types.ts
Andarist: An index signature is appropriate:
```ts
  metadata?: {
    argv?: string[];
    uri?: string;
    [k: string]: unknown;
  };
```

packages/replayio/src/utils/recordings/createdDeferredAction.ts
bvaughn: I think I understand the similarity you're describing– specifically with the resolve/reject status. That's not a core aspect of `createDeferred` though. It's just how I used it in this _one specific case_. In most cases (there are a lot within the DevTools codebase) we resolve/reject just like you would with a Promise. The key difference being that deferred also provides a way to synchronously read a rejected/resolved value (without the microtask loop).

packages/replayio/src/utils/recordings/types.ts
bvaughn: Ah, right. Nice. Thanks!

packages/replayio/src/utils/recordings/createdDeferredAction.ts
Andarist: Yeah, so I didn't mean to change anything about `createDeferred` - just this one wrapper here (`createdDeferredAction`) that has a more specific use case

bvaughn: Oh! I totally misunderstood the scope of what you were suggesting. Yeah okay let me re-think about that.

PR #347, opened 2024-04-04
Bump the compilation target in `@replayio/playwright`

Andarist:
Transforming async functions and destructuring patterns led to problems since Playwright extracts parameter names from the source text of the loaded functions. The error has been thrown here:
https://github.com/microsoft/playwright/blob/4d4308e7b3a8c870a5a44aba09dea5e0a0519435/packages/playwright/src/common/fixtures.ts#L252


ryanjduffy (approved):
I'd like to remove the hard coded value but if this resolves it for now, I'm fine with that.

PR #348, opened 2024-04-04
Stop handling non-existent `interrupted` test status

Andarist:


packages/playwright/src/reporter.ts
```diff
@@ -339,7 +339,7 @@ class ReplayPlaywrightReporter implements Reporter {
         {
           ...testMetadata,
           approximateDuration: test.results.reduce((acc, r) => acc + r.duration, 0),
-          result: status === "interrupted" ? "unknown" : status,
+          result: status,
```
Andarist, line 342: According to the playwright types here, the `'interrupted'` status doesn't exist and this prevents this package from compiling successfully

Andarist:
I might have had stale `node_modules` locally cause CI is actually complaining now that this is not handled correctly after this change

ryanjduffy:
Yep, probably. The state is newer than 1.19 so if you had dependencies installed from `main` and switch to `playwright-next`, you'll see this error until you refresh deps.

PR #349, opened 2024-04-04
Use a regular symlink to link `@replayio/cypress` into `flake`'s repo

Andarist:
It turns out that the simplest solution is the best...

For some reason, (I have no idea what has changed) `npm link` started to fail on the CI. It tried to execute `prepare` script in our `packages/cypress` (I couldn't figure out how to make it stop doing that) and failed on executing `tsc` there (it looked as if it was invoked with a wrong `cwd` or smth). 

Since we already pre-build that at the bootstrap step, I just decided to use a regular symlink to bypass npm weirdness and it just works now™


bvaughn (approved):
Nice! 👏🏼 Thanks for hunting this down.

ryanjduffy:
> bypass npm weirdness

I feel like this could be its own npm package ... or maybe a t-shirt ... or both ...

Either way, thanks for the fix!!

PR #350, opened 2024-04-05
New replayio CLI

bvaughn:
Build on top of work from #344, #346

- [x] Parallelize upload-and-process for multiple recordings.
- [x] Show new "process" type in recordings table.
- [ ] Finish implementing the following commands:
  - [ ] Process recordings
  - [ ] Upload source-maps


packages/replayio/src/utils/protocol/api/ensureProcessed.ts
```diff
@@ -0,0 +1,39 @@
...
+  await client.waitUntilAuthenticated();
+
+  let removeListeners: Function[] = [];
+
+  return new Promise<void>(async (resolve, reject) => {
+    try {
+      removeListeners.push(
+        client.listenForMessage("Recording.sessionError", (error: any) => {
+          reject(error);
+        })
```
Andarist, line 15: This should, likely, be scoped to a specific processing. From what I was able to gather here each processing gets a unique `sessionId` but we'd fail all of the pending ones based on a single `Recording.sessionError`

packages/replayio/src/utils/protocol/api/ensureProcessed.ts
```diff
@@ -0,0 +1,39 @@
...
+export async function ensureProcessed(client: ProtocolClient, sessionId: string) {
+  await client.waitUntilAuthenticated();
+
+  let removeListeners: Function[] = [];
+
+  return new Promise<void>(async (resolve, reject) => {
+    try {
+      removeListeners.push(
+        client.listenForMessage("Recording.sessionError", (error: any) => {
+          reject(error);
```
Andarist, line 14: nit: A less indented and more linear~ flow could be easier to follow. A deferred resolved here racing with that `Session.ensureProcessed` would spare you the need to construct `new Promise` manually in this function

packages/replayio/src/utils/recordings/formatRecording.ts
```diff
@@ -1,50 +1,72 @@
...
+const MAX_TITLE_LENGTH = 35;
+
 export function formatRecording(recording: LocalRecording) {
   const id = recording.id.substring(0, 8) + "…";
 
   let title;
   if (recording.metadata.host) {
-    title = chalk.blueBright.underline(recording.metadata.host);
+    if (recording.metadata.host.length > MAX_TITLE_LENGTH) {
+      title = link(recording.metadata.host.substring(0, MAX_TITLE_LENGTH) + "…");
```
Andarist, line 13: To avoid trailing whitespace followed by `...` we could do this:
```suggestion
      title = link(recording.metadata.host.substring(0, MAX_TITLE_LENGTH).trimEnd() + "…");
```

packages/replayio/src/utils/recordings/formatRecording.ts
```diff
@@ -1,50 +1,72 @@
...
-        status = "Recorded";
-        break;
-      case "recording":
-        status = "Recording";
-        break;
-      case "unusable":
-        status = "Unusable";
-        break;
+
+  switch (recording.processingStatus) {
```
Andarist, line 29: could we move this to `getRecordingStatusLabel` util or smth like that?

packages/replayio/src/utils/protocol/api/ensureProcessed.ts
bvaughn: Yeah, good call.

Maybe I should give up on trying to reuse a single protocol client for this. I was trying to be more efficient but I think it just adds too much complication.

packages/replayio/src/commands/record.ts
```diff
@@ -30,35 +28,31 @@ async function record(url: string = "about:blank") {
...
+        true
       );
 
-      printViewRecordingLinks(uploadedRecordings);
+      console.log(""); // Spacing for readability
 
-      uploadedRecordings.forEach(recording => {
-        removeFromDisk(recording.id);
-      });
+      await uploadRecordings(selectedRecordings, { processAfterUpload });
```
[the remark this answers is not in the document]
Andarist: `uploadRecordings` currently does those steps:
1. uploads recordings
2. `printDeferredRecordingActions`
3. waits for all statuses to settle
4. `printViewRecordingLinks`
5. conditionally calls `removeFromDisk`
6. closes the client

What I was getting on is that it's very CLI-oriented as it bakes in handling some of its concerns. The playwright/cypress plugins will only need 1, 3, get value out of 4, and might also include a conditional 5. It won't be a particularly hard refactor to do to get to that point. I'm just saying that a core functionality might have to be slimmed down later and other concerns might have to built on top of it closer to the CLI. 

But I see now that, in this PR, this function only got some extra stuff but it was already not completely reusable for that other use case. So it really doesn't matter rn ;p

> Edit Maybe we can file a Linear issue in the v2 milestone (the one about Test Suites support) for this

👍

packages/replayio/src/utils/protocol/api/ensureProcessed.ts
Andarist: I'd stick with reusing the single protocol client. It's quite great from the efficiency PoV - it would be a shame to lose that. It's just that WS communication isn't best suited for the typical req/res cycle so we have to - at times - be cautious about this sort of thing. From what I see `error.parameters.sessionId` should be available here so we can just spy on it and reject conditionally

packages/replayio/src/commands/record.ts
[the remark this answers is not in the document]
bvaughn: Got it. Thanks for elaborating!

Primary short-term focus is still 100% on user experience. We can def revisit this with CLI in mind in a week or two, once I've written some e2e tests and stuff too.

Andarist (approved):

PR #351, opened 2024-04-07
Annotate all pages on playwright events

ryanjduffy:
Annotate all pages opened by the browser so we have annotations for playwright scripts that use multiple pages or create pages from the browser or context instead of using the "provided" `page`.

### DIFF
--- a/packages/playwright/src/fixture.ts
+++ b/packages/playwright/src/fixture.ts
@@ -156,17 +156,21 @@ export async function replayFixture(
 
   function addAnnotation(event: string, id?: string, data?: Record<string, any>) {
     if (id) {
-      browser.contexts().forEach(context => {
-        context.pages().forEach(page => {
-          page
-            .evaluate(ReplayAddAnnotation, [
-              event,
-              id,
-              JSON.stringify({ ...data, test: testIdData }),
-            ])
-            .catch(e => warn("Failed to add annotation", e));
-        });
-      });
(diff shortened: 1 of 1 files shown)


packages/playwright/src/fixture.ts
```diff
@@ -156,9 +156,17 @@ export async function replayFixture(
 
   function addAnnotation(event: string, id?: string, data?: Record<string, any>) {
```
Andarist, line 157: Since this gets called in response to both `"step:start"` and `"step:end"`, it's probably already fine - but can we think about any race condition or an edge case when we'd miss some annotations when navigation happens in between those 2 or something like that?

ryanjduffy: Good question! We definitely miss some annotations already due to navigations. The most common case is for page navigations themselves. We could `await` all of these `evaluate` calls which would improve the chances they would all complete before a navigation but that might also slow down the overall execution because we're indiscriminately firing off these annotations to all pages even if they aren't relevant to that page.

I'll give that a look this morning though and see how it fares. I only barely tests the current iteration when I pushed this PR up.

ryanjduffy: I remember now that the current ~~hack~~ integration doesn't give us an async endpoint so we can't `await` calls here. This is about the best we can do with the current approach unfortunately.

Andarist (approved):

PR #352, opened 2024-04-07
feat: Automatic playwright setup

callingmedic911:
Closes PRO-331.

Automatically installs and setups playwright config to enable Replay.

Demo: https://www.loom.com/share/d5374c45a8444c5a837657298c35805b


packages/replayio/src/commands/setup.ts
```diff
@@ -0,0 +1,100 @@
...
+  .description("Setup Replay for Playwright test runner")
+  .action(setupPlaywright);
+setupCommand.addCommand(playwrightCommand);
+
+async function setupPlaywright({ config }: { config?: string }) {
+  const hasPackageJson = existsSync(path.join(process.cwd(), "package.json"));
+
+  if (!hasPackageJson) {
+    console.error(chalk.red("Please ensure you're in the Node.js project directory."));
+    await exitProcess(1);
```
Andarist, line 25: people often (-ish) run commands in subdirectories (they might just not be aware that they have changed their location for some reason). It would be great if this could work in those scenarios - to fix this you could recursively check parent directories to see if they have `package.json` file in them. A common pkg used for this is [`read-package-up`](https://github.com/sindresorhus/read-package-up) (or [`find-up-simple`](https://github.com/sindresorhus/find-up-simple)/[`find-up`](https://github.com/sindresorhus/find-up))

packages/replayio/src/commands/setup.ts
```diff
@@ -0,0 +1,100 @@
...
+
+  console.log(chalk.greenBright("✔"), chalk.whiteBright("Replay setup complete."));
+  console.log(chalk.whiteBright(docs.postinstall));
+
+  await exitProcess(0);
+}
+
+const docs = {
+  install: "Add @replayio/playwright as a dev dependency.",
+  config: `Update your playwright.config.ts to include Replay's Chromium browser and reporter. Here's how:
```
Andarist, line 72: Since this PR supports `.js` config files and custom paths we shouldn't hardcore the `.ts` path here
```suggestion
  config: `Update your Playwright config file to include Replay's Chromium browser and reporter. Here's how:
```

packages/replayio/src/commands/setup.ts
```diff
@@ -0,0 +1,100 @@
...
+   
+  const config: PlaywrightTestConfig = {
+    reporter: [
+      [
+        "@replayio/playwright/reporter",
+        {
+          apiKey: process.env.REPLAY_API_KEY,
+          upload: true,
+        },
+      ],
```
Andarist, line 84: at the moment we also have `createReplayReporterConfig` - it helps some nice advantages over what you have here as it automatically provides contextual types for all of the options. I think it would be good to use it here and in the jscodeshift script

packages/replayio/src/commands/setup.ts
```diff
@@ -0,0 +1,100 @@
...
+      ["line"],
+    ],
+    projects: [
+      {
+        name: "replay-chromium",
+        use: { ...(replayDevices["Replay Chromium"] as any) },
```
Andarist, line 90: I checked locally and this cast doesn't seem to be really needed and it looks suspicious. If we could avoid it, I think it would be for the better
```suggestion
        use: replayDevices["Replay Chromium"],
```

packages/replayio/src/utils/config/playwright.test.ts
```diff
@@ -0,0 +1,121 @@
...
+      name: 'replay-chromium',
+      use: { ...replayDevices["Replay Chromium"] },
+    },
+    // Other projects...
+  ],
+});
+`;
+    expect(() => setupReplayConfig(source)).toThrow(
+      "Replay imports already exist in the playwright.config.ts file."
+    );
```
Andarist, line 119: This is not the expected result, right? :p

bvaughn (changes requested):
I wasn't aware that we were thinking about adding this kind of functionality to the new CLI. Haven't had time to look at the code yet but I would like to before it lands.

My initial impression is that I'm not sure if Test Suites _Playwright_ specific logic belongs in the `replayio` CLI. I wonder if we considered a separate entrypoint (or a separate NPM package) for that stuff? I couldn't find a Linear issue where this was discussed. Maybe you could link me so I could catch up on the motivation here?

bvaughn:
Okay! Just had a chance to watch the Loom.

My high level thought is that I'd prefer this go into a separate package. Maybe even our be part of our `@replayio/cypress` and `@replayio/playwright` packages (e.g. `@replayio/playwright --init`)

My reasoning is:

The CLI is focused on helping users create recordings and upload them. This automation CLI is more of a meta command, not related to recording/uploading, and would add significant weight (in terms of dependency bundle size, and maybe also in terms of complexity– if we try to handle more complex e2e configurations).

More importantly though, this logic is specific to the test runner plugins (Playwright now– maybe Cypress or Selenium later?) and seems more appropriately maintained and versioned alongside of the respective plug-ins. For example, if we change something about the `@replayio/playwright` plug-in that required a config change, putting the initialization logic in this package would require us to also update the `replayio` with a major version bump (since it would break other versions of the `@replayio/playwright` package).

Does that make sense? Happy to chat more about it.

callingmedic911:
I agree that the current scope of replay-cli and the goals of this PR do not overlap significantly. However, I don’t want to burden the end-user with having to determine which CLI tool is appropriate for the tool. I think there is value for new users organically discovering how the CLI can assist them in getting set up with Replay. Though, purely in terms of tooling, I have no objections to keeping this in a separate package while packaged as single command.

> More importantly though, this logic is specific to the test runner plugins and seems more appropriately maintained and versioned alongside of the respective plug-ins. For example, if we change something about the @replayio/playwright plug-in that required a config change, then then --init command could easily be changed with the plug-in. Putting the initialization logic in replayio would require us to also update the replayio CLI though. (And if we got into specific Playwright+plugin version constraints, we might even need to do a breaking change version bump for the CLI.)

The setup always installs the latest version of Playwright plugin and then updates the config, so it only needs to support the latest configuration. What scenarios are you considering where we might need to deal with something other than the latest version?

bvaughn:
I feel pretty strongly about keeping this functionality separate, both in terms of code and in terms of the published package. If you disagree then we can maybe talk about it in realtime (not on a PR)

bvaughn:

PR #353, opened 2024-04-08
CLI UX improvements

bvaughn:
### [Loom summary of these changes](https://www.loom.com/share/14f17681244d4919a7a11bbc8cd0a295)

- [x] Remove the "process?" Y/N prompt and just automatically process; (we may add support for a flag to override this behavior later so that Test Suites can turn this off selectively)
- [x] Better format NodeJS recordings in `list`
- [x] When there's only one new recording, show it along with the confirmation (see below)
- [x] Don't block on processing; just kick it off (fire and forget)
- [x] Remove underline from links (to avoid confusion, since they aren't left-click clickable)
- [x] Close browser _or_ "press any key to continue" to stop recording

### Multiple new recordings:
Screenshot 2024-04-08 at 2 43 27 PM
### One new recording:
Screenshot 2024-04-08 at 2 43 43 PM


packages/replayio/src/utils/recordings/formatRecording.ts
```diff
@@ -7,58 +8,57 @@ const MAX_TITLE_LENGTH = 35;
...
+
   let title;
-  if (recording.metadata.host) {
-    if (recording.metadata.host.length > MAX_TITLE_LENGTH) {
-      title = link(recording.metadata.host.substring(0, MAX_TITLE_LENGTH).trimEnd() + "…");
-    } else {
-      title = link(recording.metadata.host);
+  switch (runtime) {
+    case "node": {
+      title = "NodeJS";
```
Andarist, line 16: Small proposal to make this one more robust 
https://github.com/replayio/replay-cli/pull/354

packages/replayio/src/commands/record.ts
```diff
@@ -30,27 +31,26 @@ async function record(url: string = "about:blank") {
   if (recordingsNew.length > 0) {
     let selectedRecordings: LocalRecording[] = [];
     if (recordingsNew.length === 1) {
-      const confirmed = await confirm("Would you like to upload the new recording?", true);
+      const confirmed = await confirm(
+        "New recording found. Would you like to upload it?",
+        true,
+        "\n" + printRecordings(recordingsNew)
```
Andarist, line 37: This one feels a little bit too much to me. As the user, I focus on the last line of the printed content and that's now a row of the single-row table. YMMV and maybe that's just my initial reaction.

At the very least columns like Date and Status feel redundant here (it's always~ Now and it's always Recorded). Btw. this is true for the other branch too. In this particular branch the Process is likely redundant too (it's always going to be `(root)`, right? 🤔 ). I could be persuaded that this one is OK here for consistency's sake.

I find it weird that we print headers here (in the single-recording branch) but we don't do that in the multi-recording branch.

packages/replayio/src/utils/browser/launchBrowser.ts
```diff
@@ -38,7 +41,31 @@ export async function launchBrowser(
...
+    onError: () => {
+      abortControllerForPrompt.abort();
+      browserClosedDeferred.resolveIfPending();
+    },
+    onExit: () => {
+      abortControllerForPrompt.abort();
+      browserClosedDeferred.resolveIfPending();
+    },
+    onSpawn: () => {
+      console.log(`Recording ${dim("(press any key to continue)")}`);
```
Andarist, line 59:
```suggestion
      console.log(`Recording ${dim("(press any key to stop recording)")}`);
```

packages/replayio/src/utils/browser/launchBrowser.ts
```diff
@@ -38,7 +41,31 @@ export async function launchBrowser(
...
+  // Wait until the user quits the browser process OR
+  // until the user presses a key to continue (in which case, we will kill the process)
+  const abortControllerForPrompt = new AbortController();
+  const browserClosedDeferred = createDeferred<void>();
 
-  await spawnProcess(runtimeExecutablePath, args, processOptions);
+  const { data: childProcess } = spawnProcess(runtimeExecutablePath, args, processOptions, {
+    onError: () => {
+      abortControllerForPrompt.abort();
+      browserClosedDeferred.resolveIfPending();
```
Andarist, line 52: q: shouldn't this be a rejection? I have no idea when our browser can error but I'd assume that usually that would mean some fatal error and that we won't get useful recordings out of this. It would also be good to at least get some insight that this has happened - do we keep this info anywhere? like in the recording log or somewhere?

packages/replayio/src/utils/recordings/upload/uploadRecording.ts
```diff
@@ -130,63 +127,26 @@ export async function uploadRecording(
...
-
-    debug("Processed recording %s", recording.id);
-  } finally {
-    debug("Releasing session %s", sessionId);
-
-    await releaseSession(client, { sessionId });
-  }
+  // Processing can take a while
+  // In some cases it's nicer to give users a URL sooner rather than waiting for processing to complete
+  processRecording(client, { recordingId: recording.id });
```
Andarist, line 149: I'm not sure what processing is yet in this context - does it make any sense to wait for it to be completed at any later point in time? Or can we truly treat it as fire-and-forget since it's all up to the backend?

Can this somehow fail/throw - in other words, should we attach a `.catch(noop)` to this or smth?

packages/replayio/src/commands/record.ts
bvaughn: > I could be persuaded that this one is OK here for consistency's sake.

That's what I was thinking. Consistency/familiarity (and a chance to verify that the host and duration look right).

> I find it weird that we print headers here (in the single-recording branch) but we don't do that in the multi-recording branch.

Fair. I can change that. Wasn't intentional.

packages/replayio/src/utils/recordings/upload/uploadRecording.ts
bvaughn: > Can this somehow fail/throw - in other words, should we attach a .catch(noop) to this or smth?

We are catching (higher up) and retrying.
```js
    try {
      await retryWithExponentialBackoff(() => processUploadedRecording(client, recording));
    } catch (error) {
      debug(`Failed to begin processing recording ${recording.id}`);

      // Processing may have failed to start, but the recording still uploaded successfully
    }
```

I can probably inline this stuff now though. It was separate before because I was also creating a session (which could itself fail). I'll do a quick cleanup pass so it's clearer.

PR #354, opened 2024-04-09
Better default recording title for node recordings

Andarist:


packages/replayio/src/utils/recordings/formatRecording.ts
```diff
@@ -5,6 +5,13 @@ import { LocalRecording } from "./types";
 
 const MAX_TITLE_LENGTH = 35;
 
+function truncateRecordingTitle(title: string) {
+  if (title.length > MAX_TITLE_LENGTH) {
+    return title.substring(0, MAX_TITLE_LENGTH).trimEnd() + "…";
+  }
+  return title;
+}
+
```
bvaughn, line 14: Maybe?
```suggestion
function truncateRecordingTitle(title: string) {
  title = title.trimEnd()
  if (title.length > MAX_TITLE_LENGTH) {
    return title.substring(0, MAX_TITLE_LENGTH) + "…";
  }
  return title;
}
```

bvaughn (approved):

packages/replayio/src/utils/recordings/formatRecording.ts
bvaughn: That's fair.

I was thinking about this being used in the context of our table formatter. (If we ever had trailing space then it seemed better to get rid of it than potentially push columns out.) It's not really important though. Fine as-is.

bvaughn:
Thanks for rebasing. I'll make sure this one lands before I NPM publish.

PR #358, opened 2024-04-09
Upload source maps when uploading a fresh recording

Andarist:
Not a lot has changed here UX-wise - beyond the added functionality (which is transparent to the user).

I thought briefly if this needs any extra work to surface to the user what's happening and all but I think it would be too technical and noisy. For what the user cares it's just part of the upload.

Note that `replayio upload-source-maps` will be implemented in the follow-up PR.

### DIFF
--- a/packages/replayio/src/utils/protocol/api/addOriginalSource.ts
+++ b/packages/replayio/src/utils/protocol/api/addOriginalSource.ts
@@ -1,0 +1,14 @@
+import { addOriginalSourceParameters, addOriginalSourceResult } from "@replayio/protocol";
+import ProtocolClient from "../ProtocolClient";
+
+export async function addOriginalSource(
+  client: ProtocolClient,
+  params: addOriginalSourceParameters
+) {
+  await client.waitUntilAuthenticated();
+
+  return await client.sendCommand<addOriginalSourceParameters, addOriginalSourceResult>({
+    method: "Record.addOriginalSource",
+    params,
+  });
+}
--- a/packages/replayio/src/utils/protocol/api/addOriginalSource.ts
+++ b/packages/replayio/src/utils/protocol/api/addOriginalSource.ts
@@ -8,7 +8,7 @@ export async function addOriginalSource(
   await client.waitUntilAuthenticated();
 
   return await client.sendCommand<addOriginalSourceParameters, addOriginalSourceResult>({
-    method: "Record.addOriginalSource",
+    method: "Recording.addOriginalSource",
     params,
   });
 }
--- a/packages/replayio/src/utils/recordings/removeFromDisk.ts
+++ b/packages/replayio/src/utils/recordings/removeFromDisk.ts
@@ -15,10 +15,14 @@ export function removeFromDisk(id?: string) {
     if (recording) {
       const { metadata, path } = recording;
 
-      metadata.sourcemaps?.forEach(path => {
-        debug("Removing recording source-map file %s", path);
+      metadata.sourcemaps.forEach(sourceMap => {
+        debug("Removing recording source-map file %s", sourceMap.path);
+        removeSync(sourceMap.path);
 
-        removeSync(path);
+        sourceMap.originalSources.forEach(source => {
+          debug("Removing recording original source file %s", source.path);
+          removeSync(source.path);
+        });
       });
 
       // Delete recording data file
@@ -34,6 +38,7 @@ export function removeFromDisk(id?: string) {
           try {
             const entry = JSON.parse(text) as LogEntry;
             switch (entry.kind) {
+              case RECORDING_LOG_KIND.originalSourceAdded:
               case RECORDING_LOG_KIND.sourcemapAdded: {
                 return entry.recordingId !== id;
               }
@@ -56,7 +61,7 @@ export function removeFromDisk(id?: string) {
 
     const files = readdirSync(recordingsPath);
     files.forEach(fileName => {
-      if (fileName.startsWith("recording-") || fileName.startsWith("sourcemap-")) {
+      if (/(recording|sourcemap|original)-/.test(fileName)) {
         removeSync(join(recordingsPath, fileName));
       }
     });
--- a/packages/replayio/src/utils/recordings/types.ts
+++ b/packages/replayio/src/utils/recordings/types.ts
@@ -37,6 +37,27 @@ export type LogEntry = {
   recordingId?: string;
   server?: string;
   timestamp: number;
+  baseURL?: string;
+  targetContentHash?: string;
+  targetURLHash?: string;
+  targetMapURLHash?: string;
+  parentId?: string;
+  parentOffset?: number;
+};
+
+export type OriginalSource = {
+  path: string;
+  parentOffset: number;
+};
+
+export type SourceMap = {
+  id: string;
+  path: string;
+  baseURL: string;
+  targetContentHash?: string;
+  targetURLHash?: string;
+  targetMapURLHash: string;
+  originalSources: OriginalSource[];
 };
 
 export type LocalRecording = {
@@ -50,7 +71,7 @@ export type LocalRecording = {
     host: string | undefined;
     processGroupId: string | undefined;
     processType: ProcessType | undefined;
-    sourcemaps: string[] | undefined;
+    sourcemaps: SourceMap[];
     uri: string | undefined;
     [key: string]: unknown;
   };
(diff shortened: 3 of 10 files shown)


packages/replayio/src/utils/recordings/upload/uploadSourceMaps.ts
```diff
@@ -0,0 +1,91 @@
...
+        },
+        {
+          concurrency: 5,
+        }
+      );
+    },
+    {
+      concurrency: 10,
+    }
+  );
```
Andarist, line 90: I reused this structure from the old CLI but I think it can be improved and I will follow up with a different queue-based take on this.

packages/replayio/src/utils/protocol/api/addOriginalSource.ts
```diff
@@ -0,0 +1,14 @@
...
+import ProtocolClient from "../ProtocolClient";
+
+export async function addOriginalSource(
+  client: ProtocolClient,
+  params: addOriginalSourceParameters
+) {
+  await client.waitUntilAuthenticated();
+
+  return await client.sendCommand<addOriginalSourceParameters, addOriginalSourceResult>({
+    method: "Record.addOriginalSource",
```
bvaughn, line 11: https://static.replay.io/protocol/tot/Recording/#method-addOriginalSource
```suggestion
    method: "Recording.addOriginalSource",
```

packages/replayio/src/utils/recordings/removeFromDisk.ts
```diff
@@ -15,10 +15,14 @@ export function removeFromDisk(id?: string) {
     if (recording) {
       const { metadata, path } = recording;
 
-      metadata.sourcemaps?.forEach(path => {
-        debug("Removing recording source-map file %s", path);
+      metadata.sourcemaps.forEach(sourceMap => {
+        debug("Removing recording source-map file %s", sourceMap.path);
+        removeSync(sourceMap.path);
```
Andarist, line 20: This will remove source maps and original sources even if their uploads failed. IIRC re-uploading the same recording isn’t a thing (please correct me if i’m wrong). So would there be any value in keeping them around in such a case?

packages/replayio/src/utils/recordings/upload/uploadRecording.ts
```diff
@@ -117,7 +118,7 @@ export async function uploadRecording(
 
   debug("Uploaded %d bytes for recording %s", size, recording.id);
 
-  // TODO [PRO-103] Upload source-maps
+  await uploadSourceMaps(client, recording);
```
bvaughn, line 121: nit: Could add some debug logs for this to potentially help ourselves later

```suggestion
  if (recording.metadata.sourcemaps.length > 0) {
    debug("Uploading %s source maps", recording.metadata.sourcemaps);

    await uploadSourceMaps(client, recording);

    debug("Uploaded %s source maps", recording.metadata.sourcemaps);
  }
```

packages/replayio/src/utils/recordings/types.ts
```diff
@@ -37,6 +37,27 @@ export type LogEntry = {
...
+};
+
+export type SourceMap = {
+  id: string;
+  path: string;
+  baseURL: string;
+  targetContentHash?: string;
+  targetURLHash?: string;
+  targetMapURLHash: string;
+  originalSources: OriginalSource[];
```
bvaughn, line 60: nit: I think it helps with merge conflicts to always alpha-sort object/type attributes (unless there's some reason _not_ to, e.g. `...rest`)

packages/replayio/src/utils/recordings/types.ts
```diff
@@ -37,6 +37,27 @@ export type LogEntry = {
   recordingId?: string;
   server?: string;
   timestamp: number;
+  baseURL?: string;
+  targetContentHash?: string;
+  targetURLHash?: string;
+  targetMapURLHash?: string;
+  parentId?: string;
+  parentOffset?: number;
+};
```
bvaughn, line 46: Man our log entries are kind of a gross grab-bag of junk 😆

packages/replayio/src/utils/recordings/removeFromDisk.ts
```diff
@@ -56,7 +61,7 @@ export function removeFromDisk(id?: string) {
 
     const files = readdirSync(recordingsPath);
     files.forEach(fileName => {
-      if (fileName.startsWith("recording-") || fileName.startsWith("sourcemap-")) {
+      if (/(recording|sourcemap|original)-/.test(fileName)) {
```
bvaughn, line 64: Nice 👍🏼

packages/replayio/src/utils/recordings/upload/uploadSourceMaps.ts
```diff
@@ -0,0 +1,91 @@
...
+import { getResourceToken } from "../../protocol/api/getResourceToken";
+import { debug } from "../debug";
+import { LocalRecording } from "../types";
+
+async function ensureResource(client: ProtocolClient, content: string) {
+  const { token } = await getResourceToken(client, { hash: `sha256:${hashValue(content)}` });
+  const resource = {
+    token,
+    saltedHash: `sha256:${hashValue(token + content)}`,
+  };
```
bvaughn, line 18: I wonder if it would be worth us adding a stronger type for [`Resource.FileHash`](https://static.replay.io/protocol/tot/Resource/#type-FileHash)?

```ts
// https://static.replay.io/protocol/tot/Resource/#type-FileHash
class FileHash {
  content: string;
  token: string;

  constructor(content: string, token: string) {
    this.content = content;
    this.token = token;
  }

  toJSON() {
    const { content, token } = this;

    return `sha256:${hashValue(token + content)}`;
  }
}
```

Maybe overkill.

bvaughn (approved):

packages/replayio/src/utils/recordings/types.ts
Andarist: yeee, i was thinking of making them a discriminated union - that could help us to maintain this (and just give us more context on which property belongs to which log entry)

packages/replayio/src/utils/recordings/upload/uploadSourceMaps.ts
Andarist: I don't mind the suggestion but both of those calls are (right now) contained in the same function so I don't see it as a problem. It's slightly more clear to me what happens with the optional `token` in the current version since it's all local and without indirection

packages/replayio/src/utils/recordings/types.ts
bvaughn: How much of that would just be false confidence (given that we don't control Chromium/Node or the data they log) 😆

packages/replayio/src/utils/recordings/upload/uploadSourceMaps.ts
bvaughn: That's fair, though we're going to introducing another instance of this for the `upload-source-maps` command.

packages/replayio/src/utils/recordings/types.ts
Andarist: I don't think it would have to be worse than what we have today. I didn't mean to remove the runtime validation that we already have (note: it's not exhaustive anyway).

packages/replayio/src/utils/recordings/upload/uploadSourceMaps.ts
Andarist: That one isn't implemented using the protocol client - the protocol required a `recordingId`. The old CLI was using `@replayio/sourcemap-upload` to do that and I have a plan to continue using it for now (let me know if you think we should pull that here).

packages/replayio/src/utils/recordings/upload/uploadRecording.ts
Andarist: I already have some debug logs here within `uploadSourceMaps`. I just added an extra one here to have a clear mark when all source maps were uploaded. I could also add a "success" log for all of the singular uploads but perhaps we could reduce the noise by logging only the start and the failure. I could see myself having a different opinion on this tomorrow 😅

packages/replayio/src/utils/recordings/types.ts
bvaughn: I don't think it would be _worse_ either. I'm just not sure it would be _better_ since we only control one side of this log.

packages/replayio/src/utils/recordings/upload/uploadSourceMaps.ts
bvaughn: Ah, gotcha. Then let's just skip this for now.

bvaughn (approved):

PR #361, opened 2024-04-10
CLI UX feedback

bvaughn:
- [x] Update default selection for the `upload` prompt to not recommend a recording that can't be uploaded (due to its status).

- [x] [PRO-133](https://linear.app/replay/issue/PRO-133): Add explicit "Enter to confirm" text to upload selection prompt.
Screenshot 2024-04-10 at 9 31 35 AM

- [x] [PRO-132](https://linear.app/replay/issue/PRO-132): Use font weight and a "→" prefix for the current row (rather than color) and replace the green/gray "✔" with "☐" for unselected and "✔" for selected.
Screenshot 2024-04-10 at 9 31 41 AM

### DIFF
--- a/packages/replayio/src/commands/upload.ts
+++ b/packages/replayio/src/commands/upload.ts
@@ -31,6 +31,8 @@ async function upload(
     selectedRecordings = findRecordingsWithShortIds(recordings, shortIds);
   } else if (all) {
     selectedRecordings = recordings;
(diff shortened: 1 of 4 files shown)


Andarist (approved):
I especially like [PRO-132](https://linear.app/replay/issue/PRO-132) :)

PR #362, opened 2024-04-10
Handle protocol authentication errors in a better way

Andarist:
I had some old token cached in one of my terminal sessions. I couldn't upload anything and accidentally my first run was with `DEBUG` on. That got me investigating why we are not printing anything to the console about failed recordings. it turned out that we print it but some of the logs were not logged because `printDeferredRecordingActions` suppressed parts of the logs in `DEBUG`. So I changed it to print the table only at the beginning and at the end - that should help to avoid confusing situations for us while debugging while it still wont interfere with the updatable CLI output.

While using this wrong token I also noticed some gnarly error logs related to the failed authentication in the launchdarkly files. While the problem was real - that path shouldn't fail fast most of the time and shouldn't introduce unfriendly logs like this. So I replaced that with some `debug` calls.

The big issue that is being fixed though is handling the authentication errors from the protocol when uploading recordings. The deferred authentication promise was always resolved with a boolean so all of the `await client.waitUntilAuthenticated()` calls were kind of useless. I'm not rejecting that deferred so we can actually catch it early and avoid going into the loop responsible for uploading recordings.

### DIFF
--- a/packages/replayio/src/utils/createDeferred.ts
+++ b/packages/replayio/src/utils/createDeferred.ts
@@ -6,8 +6,8 @@ export type StatusPending = typeof STATUS_PENDING;
 export type StatusRejected = typeof STATUS_REJECTED;
 export type StatusResolved = typeof STATUS_RESOLVED;
 
-export interface Deferred<Type, Data = void> {
-  data: Data | undefined;
+export interface Deferred<Type, Data = undefined> {
+  data: Data;
   debugLabel: string | undefined;
   promise: Promise<Type>;
   rejection: Error | undefined;
@@ -21,10 +21,15 @@ export interface Deferred<Type, Data = void> {
 
 export type Status = StatusPending | StatusRejected | StatusResolved;
 
-export function createDeferred<Type, Data = void>(
+export function createDeferred<Type, Data>(data: Data, debugLabel?: string): Deferred<Type, Data>;
+export function createDeferred<Type, Data = undefined>(
   data?: Data,
   debugLabel?: string
-): Deferred<Type, Data> {
+): Deferred<Type, Data | undefined>;
+export function createDeferred<Type, Data>(
+  data?: Data,
+  debugLabel?: string
+): Deferred<Type, Data | undefined> {
   let rejection: Error | undefined = undefined;
   let resolution: Type | undefined = undefined;
   let status: StatusPending | StatusRejected | StatusResolved = STATUS_PENDING;
@@ -46,7 +51,7 @@ export function createDeferred<Type, Data = void>(
     }
   }
 
-  const deferred: Deferred<Type, Data> = {
+  const deferred: Deferred<Type, Data | undefined> = {
     data,
     debugLabel,
 
--- a/packages/replayio/src/utils/launch-darkly/initLaunchDarklyFromAccessToken.ts
+++ b/packages/replayio/src/utils/launch-darkly/initLaunchDarklyFromAccessToken.ts
@@ -1,4 +1,5 @@
 import { readFromCache } from "../cache";
+import { GraphQLError } from "../graphql/GraphQLError";
 import { queryGraphQL } from "../graphql/queryGraphQL";
 import { cachePath } from "./config";
 import { debug } from "./debug";
@@ -24,7 +25,7 @@ export async function initLaunchDarklyFromAccessToken(accessToken: string) {
       await identifyUserProfile(id);
     }
   } catch (error) {
-    console.error(error);
+    debug("Failed to initialize LaunchDarkly profile: %o", error);
   }
 }
 
@@ -58,9 +59,7 @@ async function fetchUserIdFromGraphQLOrThrow(accessToken: string) {
   );
 
   if (errors) {
-    console.error(errors);
-
-    throw new Error("Failed to fetch auth info");
+    throw new GraphQLError("Failed to fetch auth info", errors);
   }
 
   const response = data as {
--- a/packages/replayio/src/utils/protocol/ProtocolClient.ts
+++ b/packages/replayio/src/utils/protocol/ProtocolClient.ts
@@ -1,6 +1,5 @@
 import { SessionId, sessionError } from "@replayio/protocol";
 import assert from "assert";
-import { Agent } from "http";
 import WebSocket from "ws";
 import { replayServer } from "../../config";
 import { getAccessToken } from "../authentication/getAccessToken";
@@ -11,19 +10,23 @@ import { debug } from "./debug";
 
 type Callback = (params: any) => void;
 
+type CommandData = {
+  // `command` is kept around only for debugging purposes
+  command: unknown;
+  sessionId: SessionId | undefined;
+};
+
 export default class ProtocolClient {
   private deferredAuthenticated = createDeferred<boolean>();
   private eventListeners: Map<string, Set<Callback>> = new Map();
   private nextMessageId = 1;
-  private pendingMessages: Map<number, Deferred<any, SessionId>> = new Map();
+  private pendingCommands: Map<number, Deferred<any, CommandData>> = new Map();
   private socket: WebSocket;
 
-  constructor({ agent }: { agent?: Agent } = {}) {
-    debug("Creating WebSocket for %s with %o", replayServer, { agent });
+  constructor() {
+    debug("Creating WebSocket for %s", replayServer);
 
-    this.socket = new WebSocket(replayServer, {
-      agent: agent,
-    });
+    this.socket = new WebSocket(replayServer);
 
     this.socket.on("close", this.onSocketClose);
     this.socket.on("error", this.onSocketError);
@@ -32,8 +35,8 @@ export default class ProtocolClient {
 
     this.listenForMessage("Recording.sessionError", (error: sessionError) => {
       if (error.sessionId) {
-        this.pendingMessages.forEach(deferred => {
-          if (deferred.status === STATUS_PENDING && deferred.data === error.sessionId) {
+        this.pendingCommands.forEach(deferred => {
+          if (deferred.status === STATUS_PENDING && deferred.data?.sessionId === error.sessionId) {
             deferred.reject(
               new ProtocolError({
                 code: error.code,
@@ -81,23 +84,22 @@ export default class ProtocolClient {
 
     debug("Sending command %s: %o", method, { id, params, sessionId });
 
-    this.socket.send(
-      JSON.stringify({
-        id,
-        method,
-        params,
-        sessionId,
-      }),
-      error => {
-        if (error) {
-          debug("Received socket error: %s", error);
-        }
+    const command = {
+      id,
+      method,
+      params,
+      sessionId,
+    };
+
+    this.socket.send(JSON.stringify(command), error => {
+      if (error) {
+        debug("Received socket error: %s", error);
       }
-    );
+    });
 
-    const deferred = createDeferred<ResponseType, SessionId>(sessionId);
+    const deferred = createDeferred<ResponseType, CommandData>({ sessionId, command });
 
-    this.pendingMessages.set(id, deferred);
+    this.pendingCommands.set(id, deferred);
 
     return deferred.promise;
   }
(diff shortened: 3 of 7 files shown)


packages/replayio/src/utils/createDeferred.ts
```diff
@@ -21,10 +21,15 @@ export interface Deferred<Type, Data = void> {
 
 export type Status = StatusPending | StatusRejected | StatusResolved;
 
-export function createDeferred<Type, Data = void>(
```
bvaughn, line 24: The purpose for this overloaded call was so that you didn't have to pass an explicit `undefined` to deferreds with an undefined/void data type, e.g.

```ts
createDeferred<number>("debug label"); // <- This worked before but it won't now
```

packages/replayio/src/utils/launch-darkly/initLaunchDarklyFromAccessToken.ts
```diff
@@ -24,7 +25,7 @@ export async function initLaunchDarklyFromAccessToken(accessToken: string) {
       await identifyUserProfile(id);
     }
   } catch (error) {
-    console.error(error);
+    debug("Failed to initialize LaunchDarkly profile: %o", error);
```
bvaughn, line 28: `debug` gets swallowed up by default (basically for any public user) which is why I was using `console.error` in a few places instead. Why this change?

packages/replayio/src/utils/protocol/ProtocolClient.ts
```diff
@@ -116,18 +118,18 @@ export default class ProtocolClient {
     debug("Socket error:\n", error);
 
     if (this.deferredAuthenticated.status === STATUS_PENDING) {
-      this.deferredAuthenticated.resolve(false);
+      this.deferredAuthenticated.reject(error);
```
bvaughn, line 121: Good catch.

packages/replayio/src/utils/recordings/printDeferredRecordingActions.ts
```diff
@@ -7,61 +7,69 @@ import { dim, statusFailed, statusPending, statusSuccess } from "../theme";
...
+function getRecordingStatus(recording: LocalRecording): string | undefined {
+  switch (recording.uploadStatus) {
+    case "failed":
+      return "(failed)";
+    case "uploading":
+      return "(uploading…)";
+    case "uploaded":
+      return "(uploaded)";
+  }
+}
```
bvaughn, line 29: I think we should back out the changes to this utility.

The reason this I passed some of this stuff in as parameters was to support other async operations (like processing) which would print a different status suffix. I think we may need to re-add processing (I'm getting conflicting requests about this so far).

packages/replayio/src/utils/recordings/upload/uploadRecordings.ts
```diff
@@ -30,7 +33,26 @@ export async function uploadRecordings(
...
+        message += ` Please check your ${highlight(name)}.`;
+      } else {
+        message += " Please try to `replay login` again.";
+      }
+      console.error(message);
+      await exitProcess(1);
+      return;
+    }
+    throw error;
+  }
```
bvaughn, line 55: Can you share a screenshot of what this looks like?

packages/replayio/src/utils/recordings/upload/uploadRecordings.ts
```diff
@@ -30,7 +33,26 @@ export async function uploadRecordings(
 
   const multiPartUpload = await getFeatureFlagValue<boolean>("cli-multipart-upload", false);
   const client = new ProtocolClient();
-  await client.waitUntilAuthenticated();
+  try {
+    await client.waitUntilAuthenticated();
+  } catch (error) {
+    if (
+      error instanceof ProtocolError &&
+      error.protocolCode === AUTHENTICATION_REQUIRED_ERROR_CODE
```
Andarist, line 41: We likely should handle this on more places but I want to finish some other stuff right now and I didn't have time to investigate all of the places where it could be handled similarly. The CLI uses a mix of HTTP and WS requests~ so at times this stuff just errors differently between commands

packages/replayio/src/utils/recordings/upload/uploadRecordings.ts
```diff
@@ -30,7 +33,26 @@ export async function uploadRecordings(
...
+    ) {
+      let message = `${statusFailed("✘")} Authentication failed.`;
+      if (process.env.REPLAY_API_KEY || process.env.RECORD_REPLAY_API_KEY) {
+        const name = process.env.REPLAY_API_KEY ? "REPLAY_API_KEY" : "RECORD_REPLAY_API_KEY";
+        message += ` Please check your ${highlight(name)}.`;
+      } else {
+        message += " Please try to `replay login` again.";
```
bvaughn, line 48:
```suggestion
        message += ` Please try to ${highlight("replay login")} again.`;
```

packages/replayio/src/utils/createDeferred.ts
```diff
@@ -6,8 +6,8 @@ export type StatusPending = typeof STATUS_PENDING;
 export type StatusRejected = typeof STATUS_REJECTED;
 export type StatusResolved = typeof STATUS_RESOLVED;
 
-export interface Deferred<Type, Data = void> {
```
Andarist, line 9: `void` is somewhat a weird type in TS and honestly I don't even understand it in full despite being deep into their internals. My rule of thumb is to avoid it and use `unknown`/`undefined` (depending on the context) - unless absolutely necessary to use `void` (it's sometimes handy to use it as return type of callbacks, so there is that)

packages/replayio/src/utils/createDeferred.ts
bvaughn: I don't feel _too strongly_ about this TBH. Was just kind of curious why the change.

packages/replayio/src/utils/recordings/printDeferredRecordingActions.ts
bvaughn: I don't feel _too strongly_ about this either, but it does feel like we may be making more work for ourselves in the near future by hard-coding these labels.

bvaughn: I don't feel _too strongly_ about this either, but it does feel like we may be making more work for ourselves in the near future by hard-coding these labels.

packages/replayio/src/utils/createDeferred.ts
Andarist: This particular call didn't work before:
> Argument of type 'string' is not assignable to parameter of type 'void'.ts(2345)

It's an ambiguous call at runtime because we can't know if the caller wants this to become the `.data` or the `.debugLabel`.

This call is still allowed using overloads:
```ts
createDeferred<number>();
```
So you don't have to use `undefined` explicitly here

bvaughn: > So you don't have to use undefined explicitly here

Right, only if you wanted to pass a debug label.

But you're right that... it would be ambiguous. Sigh. I think I'm pushing my brain too hard this week. Sorry. Ignore this whole chain of feedback.

packages/replayio/src/utils/launch-darkly/initLaunchDarklyFromAccessToken.ts
Andarist: Launchdarkly is not a critical piece of the CLI - so even if it fails we should be able to cope with it.

At the moment the user is greeted with this:
Errors printed in the stdout

bvaughn: Yeah that error message is kind of noisy and unhelpful. Okay this change makes sense.

packages/replayio/src/utils/recordings/printDeferredRecordingActions.ts
Andarist: I'm definitely up for moving this up if needed and letting the caller decide about it. I like the adjusted logic though as I put some time into figuring out today that some of the stuff I was expecting was swallowed completely because I used `DEBUG`. I think keeping it here will help to make sure that this is accounted for when we have to make those labels and statuses configurable for processing.

bvaughn: Yeah to be clear, I wasn't complaining about the changes to the logic, only the removal of the parameters.

packages/replayio/src/utils/recordings/upload/uploadRecordings.ts
Andarist: It isn't pretty but it doesn't step out of the comfort zone of what can already happen today since it's handled by this top-level handler:
https://github.com/replayio/replay-cli/blob/9398b58e6c92000a9a7839c7eaab520c9236023d/packages/replayio/src/index.ts#L18-L26

So it just gets logged to stderr and the process exits. We could decorate this with some "Please raise a ticket to help us fix this issue" but I think it's worth printing the error with the call stack and all. It should help us to diagnose problems when people raise issues about seeing this

bvaughn: My bad for the unclear comment. I was asking about a screenshot of the formatted prompt above (` console.error(message)`) just to make sure spacing and colors looked okay.

packages/replayio/src/utils/recordings/printDeferredRecordingActions.ts
Andarist: I brought back those parameters but I'm using render props to handle this now. Please let me know what you think.

packages/replayio/src/utils/recordings/upload/uploadRecordings.ts
bvaughn: Huh. How would auth fail _after_ you'd made a recording? We require auth before we even launch the browser.

packages/replayio/src/utils/recordings/printDeferredRecordingActions.ts
```diff
@@ -1,67 +1,65 @@
...
 
   clearInterval(interval);
-  print();
+  print(true);
   loggingDone();
 
   const failedActions = deferredActions.filter(deferred => deferred.status !== STATUS_RESOLVED);
   if (failedActions.length > 0) {
-    console.log(statusFailed(`${failedActions.length} ${failedMessage}\n`));
+    console.log(statusFailed(`${failedActions.length} recording(s) did not upload successfully\n`));
```
bvaughn, line 63: Missed a spot

packages/replayio/src/utils/recordings/upload/uploadRecordings.ts
Andarist: Yes, but we never validate the token from the env variable - we only test its presence. We could try to validate it, I guess - but then we'd be postponing the recording where *usually* it's probably fine to continue

bvaughn: Oh right. This is the API_KEY scenario.

bvaughn (approved):
This change seems good, with the exception of one spot in `printDeferred...` where it should be parameterized.

packages/replayio/src/utils/recordings/upload/uploadRecordings.ts
```diff
@@ -30,7 +33,26 @@ export async function uploadRecordings(
...
+    ) {
+      let message = `${statusFailed("✘")} Authentication failed.`;
+      if (process.env.REPLAY_API_KEY || process.env.RECORD_REPLAY_API_KEY) {
+        const name = process.env.REPLAY_API_KEY ? "REPLAY_API_KEY" : "RECORD_REPLAY_API_KEY";
+        message += ` Please check your ${highlight(name)}.`;
+      } else {
+        message += ` Please try to ${highlight("replay login")} again.`;
+      }
+      console.error(message);
+      await exitProcess(1);
```
bvaughn, line 51: I missed that we had (re)added a nested `exitProcess` call. What was the motivation over throwing/rejecting?

PR #363, opened 2024-04-10
Disable process recordings

bvaughn:
This feature is apparently too controversial (see [PRO-140](https://linear.app/replay/issue/PRO-140)) so I've just turned it off for now.

I did re-add the blocking behavior though, because we may want to re-add processing support in the future.


packages/replayio/src/utils/recordings/upload/uploadRecording.ts
```diff
@@ -138,13 +138,32 @@ export async function uploadRecording(
 
       debug(`Processing recording ${recording.id}`);
```
Andarist, line 139: this is now a dead branch, right?

Andarist (approved):

PR #366, opened 2024-04-11
Use a single promise queue for all uploads related to source maps

Andarist:
This is the follow-up cleanup that I mentioned [here](https://github.com/replayio/replay-cli/pull/358#discussion_r1557798431). It could also be seen as being part of https://linear.app/replay/issue/PRO-83

Both uploading source maps and original sources are basically the same kind of task (creating a backend resource~) and they both utilize the same protocol connection. I think it makes for both to use the same concurrency limit. The previous implementation had separate limits for both of them and thus the worst case scenario was having 50 active jobs at the same time.

I looked for an off-the-shelf solution but I couldn't find anything popular with support for "groups" so I decided to write my own here. 

- [x] tests


packages/replayio/src/utils/recordings/upload/uploadSourceMaps.ts
```diff
@@ -25,10 +25,13 @@ async function ensureResource(client: ProtocolClient, content: string) {
   return (await createResource(client, { content })).resource;
 }
 
+const queue = createPromiseQueue({ concurrency: 10 });
```
Andarist, line 28: This could live even closer to the `client` but I stayed away from doing any changes to that right now.

packages/replayio/src/utils/createPromiseQueue.ts
```diff
@@ -0,0 +1,91 @@
...
+  };
+}
+
+/**
+ * Creates a queue for adding async jobs to it. It won't run more actions at the same time than the set concurrency limit.
+ * It has support for forking a group of tasks - the consumer might `await q.waitUntilIdle()` on such a forked group or the root queue.
+ *
+ * Error handling is the responsibility of the scheduled action.
+ */
+export function createPromiseQueue({ concurrency }: { concurrency: number }): PromiseQueue {
```
bvaughn, line 47: Nice. I'll move this into the new utils/async folder once these both land :)

packages/replayio/src/utils/recordings/upload/uploadSourceMaps.ts
```diff
@@ -1,5 +1,4 @@
 import fs from "fs/promises";
-import promiseMap from "p-map";
```
bvaughn, line 2: Would be nice for us to also remove the other "p-map" reference (in `uploadRecording`) then we could remove the dependency entirely.

packages/replayio/src/utils/createPromiseQueue.ts
```diff
@@ -0,0 +1,91 @@
...
+      pendingCount++;
+      const finalize = <T>(result: T) => {
+        pendingCount--;
+        if (!pendingCount) {
+          idleDeferred?.resolve();
+          idleDeferred = undefined;
+        }
+        return result;
+      };
+      return root.add(() => action().then(finalize, finalize));
```
bvaughn, line 27: Returning `result` for the 2nd callback will cause rejected Promises to be treated as successful Promises (that returned an error instead of throwing one). That seems unintentional?

packages/replayio/src/utils/createPromiseQueue.ts
```diff
@@ -0,0 +1,91 @@
...
+interface PromiseQueueRoot {
+  add: <T>(action: () => Promise<T>) => Promise<T>;
+}
+
+interface PromiseQueue extends PromiseQueueRoot {
+  fork: () => PromiseQueue;
+  waitUntilIdle: () => Promise<undefined>;
+}
+
+function createGroup(root: PromiseQueueRoot): PromiseQueue {
```
bvaughn, line 12: This kind of code really benefits from having unit tests. I think we should either add them now or in a fast-follow commit.

packages/replayio/src/utils/createPromiseQueue.ts
```diff
@@ -0,0 +1,91 @@
...
+      return root.add(() => action().then(finalize, finalize));
+    },
+    fork: () => createGroup(root),
+    waitUntilIdle: (): Promise<undefined> => {
+      if (!pendingCount) {
+        return Promise.resolve(undefined);
+      }
+
+      idleDeferred ??= createDeferred();
+      return idleDeferred.promise;
```
bvaughn, line 36: I _think_ these semantics are right but this is another thing I would normally cover with unit tests (e.g. different combos of resolving before/between add+ waitUntilIdle)

bvaughn (approved):

packages/replayio/src/utils/recordings/upload/uploadSourceMaps.ts
Andarist: that's already done in https://github.com/replayio/replay-cli/pull/368 :)

packages/replayio/src/utils/createPromiseQueue.ts
Andarist: the `result` there is a misnomer - it's actually either a result or an error (since the same function is passed as both `onfulfilled` and `onrejected`). I'll rename it to `resultOrError` to make it clear

packages/replayio/src/utils/createPromiseQueue.ts
Andarist: Yeah, I have that as a TODO here and I want to add them asap. I tested this manually and stuff but I'm still anxious about this change. I will feel way better about landing this with those tests in. So my plan is to add them tomorrow morning and *then* land this and the one stacked on top of it.

bvaughn: I dig it.

packages/replayio/src/utils/createPromiseQueue.ts
bvaughn: Wouldn't we still be swallowing errors since (a) this method always calls `idleDeferred?.resolve()` (without checking for failure) and (b) it always wraps the passed in `action` (Promise) and returns the result instead of throwing it (which would also cause a rejected Promise to become resolved by the wrapper)

Andarist: Calling `idleDeferred?.resolve()` is intentional here. This queue isn't responsible for handling errors anyhow - it's up to each queued job.

You are right about throwing, a rejected promise is not the same as one resolved with an error 😅

bvaughn: > Calling idleDeferred?.resolve() is intentional here. This queue isn't responsible for handling errors anyhow - it's up to each queued job.

That makes sense.

Tests will help clarify the expected behavior I guess.

packages/replayio/src/utils/createPromiseQueue.ts
Andarist: I pushed out tests for this, feel free to take a look at them if you want. I'll land this once CI gets green

PR #367, opened 2024-04-11
Improve CLI poor Internet connectivity UX

bvaughn:
Work in progress

- [x] Reorganize async code into `utils/async`
- [x] Add new `logPromise` and `raceWithTimeout` helpers
- [x] Refactor async init to parallelize more (e.g. Launch Darkly init, NPM and Runtime update checks)
- [x] Log in-progress messages for async init (that requires Internet)

There's still a bit of a snag sometimes when I'm completely offline where the process hangs before exiting. I assume this is related to some kind of open socket or ref that Node's waiting to be cleaned up but I haven't been able to track it down yet. I'm going to share this with @Andarist for some initial thoughts though.


packages/replayio/src/utils/initialization/promptForNpmUpdate.ts
```diff
@@ -0,0 +1,29 @@
...
+  console.log("  Installed version:", highlight(fromVersion));
+  console.log("  New version:", highlight(toVersion));
+  console.log("");
+  console.log("To upgrade, run the following:");
+  console.log(highlight(`  npm install -g ${name}`));
+  console.log("");
+  console.log("Press any key to continue");
+  console.log("");
+
+  await prompt();
```
Andarist, line 23: One thing bothers me a little bit about this prompt - it doesn't work like the other one. it's just information and nothing happens when I confirm it. After seeing the other one I have a different expectation for this one - if I've seen one that runs the install process for me then I expect the same - familiar - process here. 

I know we might not want to run `npm install -g` for the user. They might be using the CLI through the per-project installation and they might be using an alternative package manager. Maybe we could try to detect those situations and offer the appropriate command (and actually run it)?

packages/replayio/src/utils/async/logPromise.ts
```diff
@@ -0,0 +1,71 @@
...
+import { isDebugging } from "../../config";
+import { logUpdate } from "../logUpdate";
+import { statusFailed, statusPending, statusSuccess } from "../theme";
+import { STATUS_PENDING, STATUS_REJECTED, STATUS_RESOLVED, Status } from "./createDeferred";
+
+export async function logPromise({
+  delayBeforeLoggingMs = 0,
+  messages,
+  promise,
+}: {
```
Andarist, line 11: nit: `logPromise(promise, options` could make for a good signature here

packages/replayio/src/utils/initialization/initialize.ts
```diff
@@ -0,0 +1,71 @@
...
+  const promises = Promise.all([
+    checkAuthentication(),
+    shouldCheckForRuntimeUpdate
+      ? raceWithTimeout(checkForRuntimeUpdate(), 5_000)
+      : Promise.resolve(),
+    raceWithTimeout(checkForNpmUpdate(), 5_000),
+  ]);
+
+  logPromise({
+    delayBeforeLoggingMs: 250,
```
Andarist, line 29: When I tried it locally I could see the `Initializing...` log for a split second. Maybe it would be worth delaying its dismissal (when it gets shown)?

packages/replayio/src/utils/async/timeoutAfter.ts
```diff
@@ -0,0 +1,15 @@
+type TimeoutResult = { timedOutAfter: number };
+
+export async function timeoutAfter(duration: number): Promise<TimeoutResult> {
```
Andarist, line 3: I liked the previous `wait` (the name worked especially great in situations like in the retrying functions). `timeoutAfter` is cool but my intuition was that it's supposed to throw. I think throwing would probably overcomplicate what happens in `initialize`:
https://github.com/replayio/replay-cli/blob/cli-internet-connectivity/packages/replayio/src/utils/initialization/initialize.ts#L20-L26

So I don't know if it's worth it. Or maybe we could use some `throwOnTimeout = true` option? It's totally subjective so pick whatever you feel is best here.

Andarist (approved):

packages/replayio/src/utils/async/timeoutAfter.ts
```diff
@@ -0,0 +1,15 @@
...
+  const startTime = Date.now();
+  return new Promise(resolve =>
+    setTimeout(() => {
+      const endTime = Date.now();
+      resolve({ timedOutAfter: endTime - startTime });
```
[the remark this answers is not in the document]
bvaughn: Yeah.

The reason I added this _shape_ was because `raceWithTimeout` needs to be able to differentiate between a timed-out promise (`isTimeoutResult`) and the thing it's racing.

The reason I added the _actual_ duration vs the passed one 🤷🏼  I guess it just felt slightly more correct.

packages/replayio/src/utils/launch-darkly/initLaunchDarklyFromAccessToken.ts
```diff
@@ -19,6 +22,10 @@ export async function initLaunchDarklyFromAccessToken(accessToken: string) {
       updateLaunchDarklyCache(accessToken, id);
     }
 
+    if (abortSignal.aborted) {
+      return;
+    }
```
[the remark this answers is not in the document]
bvaughn: Primarily that, and also hoping to avoid logging an error from LD init (or any internet requests it makes). Not sure we need this but it seemed safe enough to add.

packages/replayio/src/utils/launch-darkly/close.ts
```diff
@@ -1,11 +1,12 @@
+import { raceWithTimeout } from "../async/raceWithTimeout";
 import { debug } from "./debug";
 import { getLaunchDarklyClient } from "./getLaunchDarklyClient";
 
 export async function close() {
   const client = getLaunchDarklyClient(false);
   if (client) {
     try {
-      await client.close();
+      await raceWithTimeout(client.close(), 5_000);
```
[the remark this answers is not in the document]
bvaughn: No, I just added it in case 😅

packages/replayio/src/utils/initialization/promptForAuthentication.ts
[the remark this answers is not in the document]
bvaughn: Yeah, that's fair. I mainly kept it separate because of debug-logging. I'll just change it to be re-exported.

packages/replayio/src/utils/initialization/initialize.ts
```diff
@@ -0,0 +1,71 @@
...
+  requireAuthentication: boolean;
+}) {
+  // These initialization steps can run in parallel to improve startup time
+  // None of them should log anything though; that would interfere with the initialization-in-progress message
+  const promises = Promise.all([
+    checkAuthentication(),
+    shouldCheckForRuntimeUpdate
+      ? raceWithTimeout(checkForRuntimeUpdate(), 5_000)
+      : Promise.resolve(),
+    raceWithTimeout(checkForNpmUpdate(), 5_000),
```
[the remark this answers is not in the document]
bvaughn: Ooof, good call. I lost the change I had made earlier to promisify the NPM helper. Thanks!

packages/replayio/src/utils/initialization/initialize.ts
```diff
@@ -0,0 +1,71 @@
...
+  const launchDarklyAbortController = new AbortController();
+  const launchDarklyPromise = accessToken
+    ? raceWithTimeout(
+        initLaunchDarklyFromAccessToken(accessToken, launchDarklyAbortController.signal),
+        2_500,
+        launchDarklyAbortController
+      )
+    : Promise.resolve();
+
+  if (npmUpdateCheck.hasUpdate) {
```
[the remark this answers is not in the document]
bvaughn: Yes good catch.

packages/replayio/src/utils/initialization/initialize.ts
bvaughn: Yeah, I considered that but this is kind of a slipper slope. Adding delay kind of interferes with someone getting to Replay to record their important bug. I'd rather just leave this off for now.

packages/replayio/src/utils/async/timeoutAfter.ts
bvaughn: Yeah I was torn about the name change also but leaned slightly toward this one. I also considered a `throwOnTimeout` but worried the forking behavior may have been confusing. I don't mind adding it but I don't think we'll use it anywhere.

packages/replayio/src/utils/initialization/promptForNpmUpdate.ts
bvaughn: I considered that, and I like the suggestion, but I'd rather leave it out of MVP (unless you feel strongly about it). I'll file a Linear issue though!

bvaughn:
This PR has some nice improvements that I'd like to land, even before tracking down the "why do things hang sometimes on process.exit when you aren't online" question. That scenario is arguably _much more_ broken in `main` anyway

PR #368, opened 2024-04-11
Move multipart upload to the main thread

Andarist:
closes https://linear.app/replay/issue/PRO-83/set-appropriate-concurrency-limits

This moves the upload worker bit to the main thread. I believe that most of the work there was pure async IO and that should already be handled great by node itself (that was its main original selling point after all :p). This removed the overhead of creating those workers - since they were not kept in a pool. cc @callingmedic911 

The above reduces the code complexity by keeping everything in a single scope/file. It automatically ensures that the connection is type-safe (lack of type safety bit us recently, see [this PR](https://github.com/replayio/replay-cli/pull/364)).

It also keeps a constant number of concurrent uploads. We lost some concurrency limits when moving code to the new CLI and we could end up with huge number of outgoing requests. We were only limiting concurrency between parts of a single recording - disallowing to upload more than 10 at the same time. Now the limit is global - it doesn't matter if we are upload the whole recording or a part of one recording.


packages/replayio/src/utils/recordings/upload/uploadRecording.ts
```diff
@@ -198,54 +194,115 @@ async function uploadRecordingFileInParts({
...
+  const closeStream = () => stream.close();
+  stream.on("error", streamError.reject);
+
+  try {
+    const response = await Promise.race([
+      fetch(url, {
+        headers: {
+          "Content-Length": size.toString(),
+          "User-Agent": getUserAgent(),
+          Connection: "keep-alive",
```
Andarist, line 282: The previous version of the code was only using this for the multipart upload - I think it can safely be used at all times here. On top of that, the previous version of the code was creating a custom HTTP agent. It was only setting `keepAlive: true` option on it but that's already the default:
https://github.com/nodejs/node/blob/ee4fa77624f17c1958dac47fddbe7d9513b9dee8/lib/_http_agent.js#L557

packages/replayio/src/utils/recordings/upload/uploadRecording.ts
```diff
@@ -198,54 +194,115 @@ async function uploadRecordingFileInParts({
...
+          },
+          (error: any) => {
+            debug(`Failed to upload part ${index + 1}. Will be retried: %o`, error);
+            if (error.code === "ENOENT") {
+              throw error;
+            }
+          }
+        );
+      } catch (error) {
+        abortController.abort();
```
Andarist, line 229: There is probably no point in uploading other parts if one of them failed to be uploaded (note that they are retried before this handler kicks in). So I decided to add some cancellation to all of this.

packages/replayio/src/utils/recordings/upload/uploadRecording.ts
```diff
@@ -91,19 +92,22 @@ export async function uploadRecording(
...
+        retryWithExponentialBackoff(
+          () =>
+            uploadRecordingFile({
+              recordingPath: path,
+              size,
+              url: uploadLink,
+            }),
+          (error: any, attemptNumber: number) => {
+            debug(`Attempt ${attemptNumber} to upload failed:\n%j`, error);
+            if (error.code === "ENOENT") {
```
Andarist, line 105: It's quite an edge case and it shouldn't happen - but it did happen before https://github.com/replayio/replay-cli/pull/364 . Either way, it's worth handling it here. If the target file doesn't exist there is no point in retrying (throwing in this `onFail` callback prevents the next retrying attempt).

bvaughn (approved):
I dig it.

packages/replayio/src/utils/recordings/upload/uploadRecording.ts
callingmedic911: I'm not sure what node version we want to support but if I remember correctly this default was changed recently (Node 16/18 maybe?).

callingmedic911:
Love it, thanks!

packages/replayio/src/utils/recordings/upload/uploadRecording.ts
Andarist: Oh, goooood catch. Node 18 doesn't have this `keepAlive` there. I guess, I'll have to replicate the Node 20 behavior by creating a custom agent here (one that we'll be able to reuse across fetches)

packages/replayio/src/utils/recordings/upload/uploadRecording.ts
```diff
@@ -198,54 +194,115 @@ async function uploadRecordingFileInParts({
...
-        },
-        10
-      );
-    },
-    { concurrency: 10 }
-  );
+  const { size: totalSize } = await stat(recordingPath);
+  const abortController = new AbortController();
+
+  const partsUploads = partLinks.map((url: string, index: number) => {
```
callingmedic911, line 200: During my testing, if I surge with all chunks at once, most of them will timeout with a bad connection. It will either end up in more retries or complete failure (hitting max retry limit). I think we should retain concurrency control.

Andarist: This is still capped as every scheduled chunk gets put on the queue. This line here just awaits the results - but those promises are not executed eagerly (they have concurrency set to 10).

Andarist: Or are you saying that you ended up with this situation when testing this PR locally? 🤔

callingmedic911: No, I haven't checked out locally. I looked PR only superficially, for that matter. I didn't want to block PR just because I'm occupied with other stuff.

PR #369, opened 2024-04-11
Use modern JS output in the new CLI

Andarist:


packages/replayio/package.json
```diff
@@ -3,6 +3,9 @@
   "version": "0.0.2",
   "description": "CLI tool for uploading and managing recordings",
   "bin": "./replayio.js",
+  "engines": {
+    "node": ">=18"
```
Andarist, line 7: It's the current LTS - released at 19 Apr 2022

PR #372, opened 2024-04-11
Remove 'view' command

bvaughn:


packages/replayio/src/commands/view.ts
```diff
@@ -1,26 +0,0 @@
...
-    const recordings = await getRecordings();
-    const ids = findRecordingsWithShortIds(recordings, [idOrShortId]);
-    if (ids.length === 1) {
-      id = ids[0].id;
-    }
-  }
-
-  const url = `${replayAppHost}/recording/${id}`;
-
-  spawn(getSystemOpenCommand(), [url]);
```
Andarist, line 23: This was probably the only caller of `getSystemOpenCommand` so that might be removed

Andarist (approved):
I’m not a target of this feature at all but i could see it being useful to some folks if it would allow them to pick the uploaded recording from the list. Otoh, we are removing recording entries from the log after uploading it so that would require querying the API for the recording list

packages/replayio/src/commands/view.ts
bvaughn: The authenticate-by-browser helper (for `record`) also uses it.

PR #373, opened 2024-04-11
Address `process.exit` hang [draft]

bvaughn:
Work in progress...


packages/replayio/src/utils/async/overrideSetTimeout.ts
Andarist, line 1: are we sure that timeouts are the problem here?

bvaughn: It seemed that way based on outstanding launch darkly timeouts, and the fact that the behavior went away if I removed that dependency. But honestly I'm not sure.

PR #374, opened 2024-04-11
Improve package upgrade prompt

bvaughn:
TBH I'm not sure if even this is worth doing. I think everyone who uses a package manager other than NPM knows how to convert the standard "npm install" message to their own package-manager-of-choice.

### DIFF
--- a/packages/replayio/src/utils/initialization/getLatestPackageVersion.ts
+++ b/packages/replayio/src/utils/initialization/getLatestPackageVersion.ts
@@ -1,0 +1,31 @@
+import { exec } from "child_process";
+import { promisify } from "util";
+import { PackageManager } from "./types";
+
+const execAsync = promisify(exec);
+
+export async function getLatestPackageVersion(command: PackageManager, packageName: string) {
+  try {
+    switch (command) {
+      case "npm":
+      case "pnpm": {
+        const { stdout: text } = await execAsync(`${command} info ${packageName} --json`, {
+          encoding: "utf8",
+        });
+        const json = JSON.parse(text.trim());
+
+        return json.version;
+      }
+      case "yarn": {
+        const { stdout: text } = await execAsync(`yarn npm info ${packageName} --json`, {
+          encoding: "utf8",
+        });
+        const json = JSON.parse(text.trim());
+
+        return json["dist-tags"]?.latest;
+      }
+    }
+  } catch (error) {
+    // Ignore
+  }
+}
--- a/packages/replayio/src/utils/initialization/getPackageManagerCommand.ts
+++ b/packages/replayio/src/utils/initialization/getPackageManagerCommand.ts
@@ -1,0 +1,22 @@
+import { basename, sep } from "path";
+import { isPackageManagerInstalled } from "./isPackageManagerInstalled";
+import { PackageManager } from "./types";
+
+export function getPackageManagerCommand(): PackageManager | undefined {
+  // If this is a global install, it's possible we can detect the package manager that way
+  const path = basename(__filename);
+  if (path.includes("pnpm" + sep)) {
+    return "pnpm";
+  } else if (path.includes("yarn" + sep)) {
+    return "yarn";
+  }
+
+  // Otherwise let's see which package manager(s) are installed
+  if (isPackageManagerInstalled("pnpm")) {
+    return "pnpm";
+  } else if (isPackageManagerInstalled("yarn")) {
+    return "yarn";
+  } else if (isPackageManagerInstalled("npm")) {
+    return "npm";
+  }
+}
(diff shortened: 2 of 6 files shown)


packages/replayio/src/utils/initialization/getPackageManagerCommand.ts
```diff
@@ -0,0 +1,22 @@
...
+export function getPackageManagerCommand(): PackageManager | undefined {
+  // If this is a global install, it's possible we can detect the package manager that way
+  const path = basename(__filename);
+  if (path.includes("pnpm" + sep)) {
+    return "pnpm";
+  } else if (path.includes("yarn" + sep)) {
+    return "yarn";
+  }
+
+  // Otherwise let's see which package manager(s) are installed
```
Andarist, line 14: I have all of them installed. It would be better to lookup the closest lockfile

packages/replayio/src/utils/initialization/getLatestPackageVersion.ts
```diff
@@ -0,0 +1,31 @@
+import { exec } from "child_process";
+import { promisify } from "util";
+import { PackageManager } from "./types";
+
+const execAsync = promisify(exec);
+
+export async function getLatestPackageVersion(command: PackageManager, packageName: string) {
```
Andarist, line 7: I probably wouldn't even mind just using `npm` for this. Your version is better/more robust but even if a project uses an alternative package manager they will likely still have `npm` available.

But we could even go a step further - spawning an extra process for this is already a little bit wasteful. So we could optimize this by making a direct API call:
```ts
  const response = await fetch("https://registry.npmjs.org/replayio", {
    headers: {
      // request abbreviated metadata
      Accept: "application/vnd.npm.install-v1+json",
    },
  });
  const json = await response.json();
  const latestVersion = json["dist-tags"]?.latest;
```

This is documented [here](https://github.com/npm/registry/blob/1c794110badd54b9d9fb08e7489746b6089c6648/docs/responses/package-metadata.md)

packages/replayio/src/utils/initialization/promptForNpmUpdate.ts
```diff
@@ -1,29 +1,50 @@
...
-  console.log("  Installed version:", highlight(fromVersion));
-  console.log("  New version:", highlight(toVersion));
-  console.log("");
-  console.log("To upgrade, run the following:");
-  console.log(highlight(`  npm install -g ${name}`));
-  console.log("");
-  console.log("Press any key to continue");
-  console.log("");
+  const command = getPackageManagerCommand();
+  if (command) {
```
Andarist, line 14: >  I think everyone who uses a package manager other than NPM knows how to convert the standard "npm install" message to their own package-manager-of-choice.

Right. So I wasn't worried about what we print here - just that we don't execute the command when the user confirms the prompt. We execute runtime installs - that's of course much more helpful since the user doesn't have to know how to do that from the top of their head. It's just that when we are prompting for both it feels off that they both don't execute things. 

That was my literal confusion as the user here. I was first prompted about the runtime update and later about the npm update (during some subsequent run). At that point, I was just expecting that it would also perform the installation for me but it didn't.

Andarist: I think that without this auto-installation it might not be worth pulling in this PR. I understand that auto-installation isn't exactly a no-brainer with a plethora of alternative package managers on the market, use cases and setups. So staying away from it might very well be just the right thing to do to avoid edge cases and weird problems.

packages/replayio/src/utils/initialization/getPackageManagerCommand.ts
bvaughn: Hm. I dunno. I think the likelihood of you running the CLI from within a project folder that also happens to be the way you installed `replayio` seems low. I either run it from my home folder (default terminal location) or I run it from whatever project I'm working in (which may use Yarn or PNPM, but that's not how I globally installed the CLI)

packages/replayio/src/utils/initialization/getLatestPackageVersion.ts
bvaughn: That seems like an improvement.

packages/replayio/src/utils/initialization/promptForNpmUpdate.ts
bvaughn: Having looked into this, and thought more about the edge cases, I'm inclined to agree. I posted this PR for discussion but I'm happy to bail on it and just do [this one thing](https://github.com/replayio/replay-cli/pull/374#discussion_r1562227248) instead.

packages/replayio/src/utils/initialization/getLatestPackageVersion.ts
bvaughn: https://github.com/replayio/replay-cli/pull/376

PR #375, opened 2024-04-12
Add `upload-source-maps` command

Andarist:
This PR introduces `upload-sourcemaps` command. It can be used like this:
```ts
replayio upload-sourcemaps -g test1 ../devtools/.next/static/chunks
```


packages/replayio/src/commands/upload-source-maps.ts
Andarist, line 1: I changed this to `upload-sourcemaps` because that's how it was called in the old CLI. I can revert this change but it feels like we use that very inconsistently already in a bunch of places (error messages, variable names, protocol commands, command names, package names). Are they source maps or sourcemaps? How should we camelCase this? :P

packages/replayio/src/commands/upload-sourcemaps.ts
```diff
@@ -0,0 +1,63 @@
+import { uploadSourceMaps as uploadSourceMapsExternal } from "@replayio/sourcemap-upload";
+import { replayApiServer } from "../config";
+import { getAccessToken } from "../utils/authentication/getAccessToken";
+import { registerCommand } from "../utils/commander/registerCommand";
+import { exitProcess } from "../utils/exitProcess";
+import { dim } from "../utils/theme";
+import { logPromise } from "../utils/async/logPromise";
+
+registerCommand("upload-sourcemaps <paths...>", { requireAuthentication: true })
```
Andarist, line 9: I removed a couple of options from here because they felt redundant. For instance, there is no need for `--api-key` when we support `REPLAY_API_KEY` already. `batch-size` felt like something the user shouldn't care about - so why do we bother exposing it? `dry-run`, `verbose` and `quiet` could be useful but I think it's a bigger question and if anything them (or some of them) should be introduced to the whole new CLI - having them just here would be weird.

packages/replayio/src/utils/async/logPromise.ts
```diff
@@ -56,16 +61,16 @@ export async function logPromise(
...
   clearInterval(interval);
 
-  print();
-
-  logUpdate.done();
+  try {
+    print();
+  } finally {
+    logUpdate.done();
+  }
```
Andarist, line 75: I wrapped this just in case - the `messages.failed` could like to rethrow the error or something and in such a case we should still call this.

packages/replayio/src/commands/upload-sourcemaps.ts
bvaughn: Agreed with those simplifications.

packages/replayio/src/utils/async/logPromise.ts
```diff
@@ -2,30 +2,29 @@ import { dots } from "cli-spinners";
 import { isDebugging } from "../../config";
 import { logUpdate } from "../logUpdate";
 import { statusFailed, statusPending, statusSuccess } from "../theme";
-import { STATUS_PENDING, STATUS_REJECTED, STATUS_RESOLVED, Status } from "./createDeferred";
+import { STATUS_PENDING, STATUS_REJECTED, STATUS_RESOLVED, createDeferred } from "./createDeferred";
 
-export async function logPromise(
-  promise: Promise<any>,
+export async function logPromise<T>(
```
bvaughn, line 7: tiny nit: I prefer descriptive names over single letter variable or generic type names. Single letter names are sometimes (not always) vague to people who didn't write the code originally, and I prefer clarity/consistency to avoid those cases.
```suggestion
export async function logPromise<PromiseType>(
```

packages/replayio/src/utils/async/logPromise.ts
bvaughn: I don't care, since it doesn't hurt to have this, but that scenario sounds bizarre to me. This helper is just for printing; control flow should be handled separately (with the original promise)

packages/replayio/src/commands/upload-sourcemaps.ts
```diff
@@ -0,0 +1,63 @@
...
+  .description("Upload source-maps for a Workspace")
+  .requiredOption(
+    "-g, --group <name>",
+    "The name to group this sourcemap into, e.g. A commit SHA or release version."
+  )
+  .option(
+    "-x, --extensions <exts>",
+    `A comma-separated list of extensions to process; ${dim('default ".js,.map"')}`,
```
bvaughn, line 17: nit: Maybe more clear?
```suggestion
    `A comma-separated list of file extensions to process; ${dim('default ".js,.map"')}`,
```

packages/replayio/src/commands/upload-sourcemaps.ts
```diff
@@ -0,0 +1,63 @@
...
+    (value: string) => value.split(",")
+  )
+  .option(
+    "-i, --ignore <pattern>",
+    "Ignore files that match this pattern",
+    (value: string, previous: Array<string> = []) => {
+      return previous.concat([value]);
+    }
+  )
+  .option("    --root <dirname>", "The base directory to use when computing relative paths")
```
bvaughn, line 27: nit: Command doesn't require this leading space
```suggestion
  .option("--root <dirname>", "The base directory to use when computing relative paths")
```

packages/replayio/src/commands/upload-sourcemaps.ts
```diff
@@ -0,0 +1,63 @@
...
+    key: await getAccessToken(),
+    root,
+    server: replayApiServer,
+  });
+
+  await logPromise(uploadPromise, {
+    messages: {
+      pending: "Uploading sourcemaps...",
+      success: "Source maps uploaded",
+      failed: error => `Source maps upload failed:\n${error}`,
```
bvaughn, line 58: Cool. I like this change.

bvaughn (approved):

packages/replayio/src/commands/upload-source-maps.ts
Andarist: but seriously - what they are? sourcemaps or source maps? 😅 it would be great to have this consistent at least within this new code here

packages/replayio/src/utils/async/logPromise.ts
Andarist: I agree and I'm not sure what I was thinking here :P I will revert this change cause if anything the whole thing should be wrapped with this `finally` block but it would indent the whole thing and I don't think it's really worth it

packages/replayio/src/commands/upload-source-maps.ts
bvaughn: I usually think "source-maps" in my head but I agree, this is one of those terms that are inconsistent _everywhere_ and I just try not to think about it because if I do... 🤪

Andarist: I decided to rollback this and unify it where possible. Chrome, VS Code and HTTP header call them source maps and not sourcemaps

packages/replayio/src/commands/upload-source-maps.ts
```diff
@@ -1,22 +1,21 @@
+import { uploadSourceMaps as uploadSourceMapsExternal } from "@replayio/sourcemap-upload";
+import { replayApiServer } from "../config";
+import { getAccessToken } from "../utils/authentication/getAccessToken";
 import { registerCommand } from "../utils/commander/registerCommand";
 import { exitProcess } from "../utils/exitProcess";
 import { dim } from "../utils/theme";
+import { logPromise } from "../utils/async/logPromise";
 
 registerCommand("upload-source-maps <paths...>", { requireAuthentication: true })
+  .alias("upload-sourcemaps")
```
bvaughn, line 10: I think we should remove this.

bvaughn: I don't think it adds much value and I think it slightly worsens the first impression of the library:
Screenshot 2024-04-12 at 10 04 16 AM

Andarist: It might help people to migrate to the new CLI - since the command name was changed. What do you think about dropping aliases from this table? I didn't realize that it would be displayed here.

Note that without this we get a good suggestion from `commander`:
```
error: unknown command 'upload-sourcemaps'
(Did you mean upload-source-maps?)
```
That said, I would consider this to be a likely error so I thought it would be nice to handle this gracefully and deal with it. Up to you though - I can make any requested changes.

PR #376, opened 2024-04-12
Remove dependency on NPM command for version checks

bvaughn:
Alternative to #374, discussed here: https://github.com/replayio/replay-cli/pull/374#discussion_r1562227248

### DIFF
--- a/packages/replayio/src/utils/initialization/checkForNpmUpdate.ts
+++ b/packages/replayio/src/utils/initialization/checkForNpmUpdate.ts
@@ -1,20 +1,20 @@
-import { exec } from "child_process";
-import { promisify } from "util";
-import { version as currentVersion, name } from "../../../package.json";
+import { version as currentVersion, name as packageName } from "../../../package.json";
 import { shouldPrompt } from "../prompt/shouldPrompt";
 import { debug } from "./debug";
 import { UpdateCheck } from "./types";
-
-const execAsync = promisify(exec);
 
 const PROMPT_ID = "npm-update";
 
 export async function checkForNpmUpdate(): Promise<UpdateCheck<string>> {
   try {
-    const { stdout: text } = await execAsync(`npm info ${name} --json`, {
-      encoding: "utf8",
+    // https://github.com/npm/registry/blob/master/docs/responses/package-metadata.md#abbreviated-metadata-format
+    const response = await fetch(`https://registry.npmjs.org/${packageName}`, {
+      headers: {
+        Accept: "application/vnd.npm.install-v1+json",
+      },
     });
-    const { version: latestVersion } = JSON.parse(text.trim());
+    const json = (await response.json()) as any;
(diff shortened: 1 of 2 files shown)


packages/replayio/src/utils/initialization/checkForNpmUpdate.ts
```diff
@@ -1,20 +1,20 @@
...
-    const { stdout: text } = await execAsync(`npm info ${name} --json`, {
-      encoding: "utf8",
+    // https://github.com/npm/registry/blob/master/docs/responses/package-metadata.md#abbreviated-metadata-format
+    const response = await fetch(`https://registry.npmjs.org/${packageName}`, {
+      headers: {
+        Accept: "application/vnd.npm.install-v1+json",
+      },
     });
-    const { version: latestVersion } = JSON.parse(text.trim());
+    const json = (await response.json()) as any;
```
Andarist, line 16: nit: it's already `any` ;p
```suggestion
    const json = await response.json();
```
unless you wanted to be explicit about it - in such a case i'd probably go with a type annotation over a cast:
```suggestion
    const json: any = await response.json();
```

packages/replayio/src/utils/initialization/checkForNpmUpdate.ts
```diff
@@ -1,20 +1,20 @@
...
+    const latestVersion = json["dist-tags"]?.latest;
```
Andarist, line 17: I'd say that we should either trust this response or not trust it all. `?.` gives me a vibe that we don't trust it whereas `hasUpdate: currentVersion !== latestVersion` gives me a vibe that we *do* trust it (but in such a case it's inconsistent with the optional chain - and if we think that optional chain is good here then i'd probably add an early return here for `typeof latestVersion !== 'string'`)

Andarist (approved):

packages/replayio/src/utils/initialization/checkForNpmUpdate.ts
Andarist: Huh, that's weird - I *did* check that in the IDE before posting. Rechecking now... you are right. Maybe somehow my IDE pulled in `@types/node-fetch` as that has this typed as `Promise<any>`: https://unpkg.com/browse/@types/node-fetch@2.6.11/index.d.ts#L133

Note though that this file doesn't use `node-fetch`. It's reaching for a global - perhaps not intentionally.

bvaughn: That's fair...hmm. TSC warned without this `: any` so I added it 🤷🏼  I dunno man.

Andarist: The modern version of this lib type this as `Promise<unknown>` (see [here](https://unpkg.com/browse/node-fetch@3.3.2/@types/index.d.ts#L139)) - so that weirdness was probably just me jumping between branches or something. All good :p

PR #377, opened 2024-04-12
Add README and bump package to 0.0.6

bvaughn:
- [x] Remove `upload-sourcemaps` alias (subjective change)
- [x] Add basic `README` and `Contributing` files
- [x] Bump package version to 0.0.6

### DIFF
--- a/packages/replayio/src/commands/upload-source-maps.ts
+++ b/packages/replayio/src/commands/upload-source-maps.ts
@@ -7,7 +7,6 @@ import { dim } from "../utils/theme";
 import { logPromise } from "../utils/async/logPromise";
 
 registerCommand("upload-source-maps <paths...>", { requireAuthentication: true })
-  .alias("upload-sourcemaps")
   .description("Upload source-maps for a Workspace")
   .requiredOption(
     "-g, --group <name>",
--- a/packages/replayio/Contributing.md
+++ b/packages/replayio/Contributing.md
@@ -1,0 +1,25 @@
+This project uses the [Yarn package manager](https://yarnpkg.com/). To install project dependencies:
+```bash
+# Workspace root
+yarn
+yarn run bootstrap
+```
+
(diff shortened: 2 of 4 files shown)


packages/replayio/src/commands/upload-source-maps.ts
```diff
@@ -7,7 +7,6 @@ import { dim } from "../utils/theme";
 import { logPromise } from "../utils/async/logPromise";
 
 registerCommand("upload-source-maps <paths...>", { requireAuthentication: true })
-  .alias("upload-sourcemaps")
```
bvaughn, line 10: Subjective change, I guess, but this...
Screenshot 2024-04-12 at 10 05 55 AM

...is a better first impression than this...
Screenshot 2024-04-12 at 10 04 16 AM

This isn't a command people will be running often, so I think we can just ask them to type it correctly :)

Andarist: The only alternative that comes to mind is to drop the aliases from this table altogether (see my comment [here](https://github.com/replayio/replay-cli/pull/375#discussion_r1562613521)) - that is if we want to keep it being available to people.

Andarist (approved):

packages/replayio/src/commands/upload-source-maps.ts
bvaughn: I think we should just do a clean break. We've already done this with so many things (`ls` -> `list`, `rm` -> `remove` and `rm-all` -> `remove --all`, etc.) I don't see a reason to carry the old alias forward :)

Andarist: My thinking here was a mix of keeping the old name and that people (us included :P) use that spelling interchangeably - it's actually the latter that made me add the alias. I'm fine with removing it though 👍

PR #379, opened 2024-04-12
Add `replayio info` command

bvaughn:
I think this will be helpful when debugging potential stale NPM cache issues (e.g. with `npx`) and may also be useful when reporting bugs with either Replay CLI or browser. Example output:

Screenshot 2024-04-12 at 12 55 57 PM

### DIFF
--- a/packages/replayio/src/commands/info.ts
+++ b/packages/replayio/src/commands/info.ts
@@ -1,0 +1,29 @@
+import { name as packageName, version as packageVersion } from "../../package.json";
+import { registerCommand } from "../utils/commander/registerCommand";
+import { exitProcess } from "../utils/exitProcess";
+import { getCurrentRuntimeMetadata } from "../utils/initialization/getCurrentRuntimeMetadata";
+import { parseBuildId } from "../utils/installation/parseBuildId";
+import { dim, highlight } from "../utils/theme";
+
+registerCommand("info", { requireAuthentication: false })
+  .description("Display info for installed Replay dependencies")
+  .action(info);
+
+async function info() {
+  console.log(`Currently using ${highlight(`${packageName}@${packageVersion}`)}`);
+
+  const metadata = getCurrentRuntimeMetadata("chromium");
+  if (metadata) {
+    const { buildId, nativeVersion } = metadata;
+
+    const { releaseDate } = parseBuildId(buildId);
+
+    console.log("\nReplay Chromium");
+    console.log(`• Release date: ${highlight(releaseDate.toLocaleDateString())}`);
+    if (nativeVersion) {
+      console.log(`• Native version: ${highlight(nativeVersion)}`);
+    }
+  }
+
+  await exitProcess(0);
+}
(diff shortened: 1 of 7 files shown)


packages/replayio/src/utils/installation/installLatestRelease.ts
```diff
@@ -44,14 +44,17 @@ export async function installLatestRelease() {
...
 
   // Write version metadata to disk so we can compare against the latest release and prompt to update
   debug("Saving release metadata to %s", metadataPath);
   writeToCache<MetadataJSON>(metadataPath, {
     chromium: {
       buildId: latestBuildId,
       installDate: new Date().toISOString(),
+      nativeVersion: latestVersion,
     },
+    node: undefined,
```
bvaughn, line 57: We're mostly only supporting Chromium for now but I don't see the harm in thinking a little bit ahead for other runtimes.

bvaughn:
Tagging a couple of folks in hopes of getting a stamp for this since there's no auto-approve in this repo.

packages/replayio/src/utils/installation/installLatestRelease.ts
Andarist: sure but this will get lost immediately anyway :D `JSON.stringify` will strip it away

bvaughn: Yup! TS complains without it though bc of the metadata type.

bvaughn: Maybe if I wrap a `Partial<...>` around the `Record<...>` it wouldn't?

packages/replayio/src/commands/info.ts
```diff
@@ -0,0 +1,29 @@
...
+  const metadata = getCurrentRuntimeMetadata("chromium");
+  if (metadata) {
+    const { buildId, nativeVersion } = metadata;
+
+    const { releaseDate } = parseBuildId(buildId);
+
+    console.log("\nReplay Chromium");
+    console.log(`• Release date: ${highlight(releaseDate.toLocaleDateString())}`);
+    if (nativeVersion) {
+      console.log(`• Native version: ${highlight(nativeVersion)}`);
```
Andarist, line 24: q: why do we call this a "native" version?

Andarist (approved):

packages/replayio/src/commands/info.ts
bvaughn: I couldn't think of a better name for it. Original version? Base version? Forked version?

packages/replayio/src/utils/installation/types.ts
```diff
@@ -16,9 +16,14 @@ export type Release = {
...
+  Record<
+    Runtime,
+    | {
+        buildId: string;
+        installDate: string;
+        nativeVersion: string | null;
+      }
+    | undefined
+  >
+>;
```
Andarist, line 29: This one could help a little bit with nesting but ofc both are fine:
```suggestion
export type MetadataJSON = {
  [K in Runtime]?: {
    buildId: string;
    installDate: string;
    nativeVersion: string | null;
  }
}
```

packages/replayio/src/commands/info.ts
Andarist: A fork version could work - it makes sense and its readable (I assume that this is like a version from which this particular chromium fork is based on). A base version is OK too. I guess this is mostly useful for us anyway so it doesn't matter all that much.

PR #384, opened 2024-04-12
Automatically upload crash data

bvaughn:

### DIFF
--- a/packages/replayio/src/commands/record.ts
+++ b/packages/replayio/src/commands/record.ts
@@ -9,6 +9,7 @@ import { printRecordings } from "../utils/recordings/printRecordings";
 import { selectRecordings } from "../utils/recordings/selectRecordings";
 import { LocalRecording } from "../utils/recordings/types";
 import { uploadRecordings } from "../utils/recordings/upload/uploadRecordings";
+import { logPromise } from "../utils/async/logPromise";
 
 registerCommand("record", { checkForRuntimeUpdate: true, requireAuthentication: true })
   .argument("[url]", `URL to open (default: "about:blank")`)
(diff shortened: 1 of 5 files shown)


packages/replayio/src/utils/recordings/formatRecording.ts
```diff
@@ -59,6 +59,7 @@ export function formatRecording(recording: LocalRecording) {
     switch (recordingStatus) {
       case "crashed":
         status = "Crashed";
+        break;
```
bvaughn, line 62: This was an oversight from before.

packages/replayio/src/utils/recordings/upload/uploadCrashData.ts
```diff
@@ -14,7 +14,7 @@ export async function uploadCrashedData(client: ProtocolClient, recording: Local
     recordingId: recording.id,
   });
 
-  await Promise.all(crashData.map(async data => reportCrash(client, data)));
+  await Promise.all(crashData.map(async data => reportCrash(client, { data })));
```
bvaughn, line 17: The [docs](https://static.replay.io/protocol/tot/Internal/#method-reportCrash) are not very clear, but this was wrong before.

packages/replayio/src/commands/record.ts
```diff
@@ -9,44 +9,79 @@ import { printRecordings } from "../utils/recordings/printRecordings";
...
+
+    const promise = uploadRecordings(nextCrashedRecordings, {
+      processAfterUpload: false,
+      silent: true,
+    });
+
+    logPromise(promise, {
+      messages: {
+        pending: "Uploading crash data...",
+        success: "Crash data uploaded successfully",
```
Andarist, line 55: if you intentionally dont want to render anything here when it fails then maybe a comment here would be helpful explianing? it's tempting to add that error message here

packages/replayio/src/commands/record.ts
```diff
@@ -9,44 +9,79 @@ import { printRecordings } from "../utils/recordings/printRecordings";
...
+      },
+    });
+
+    await promise;
```
Andarist, line 59: if this fails - do we want to halt everything?

Andarist (approved):

packages/replayio/src/commands/record.ts
bvaughn: I thought about that but I'm not sure what we would tell the user.

At that point, the runtime has crashed _and_ we failed to upload crash data. It looks pretty clowny and there isn't any action the user take (beyond maybe reporting this to us via a ticket). Which maybe that is what we should suggest, I dunno.

bvaughn: Going to land this as-is but happy to iterate on the messaging if you have suggestions.

packages/replayio/src/commands/record.ts
Andarist: > Other types of errors would get caught by createSettledDeferred and resolved with a status of false (rather than throwing).

Right. So that means that it can't ever fail here. And since it's configured with `silent: true` we end up with a very broken situation in which we report `✔ Crash data uploaded successfully` when the upload of the crashed data has failed.

packages/replayio/src/commands/record.ts
Andarist: > It looks pretty clowny and there isn't any action the user take (beyond maybe reporting this to us via a ticket).

Even if the user can't take any action in this situation, I think it's still reasonable to let them know what happened. It's about expectations. Like, they would see that we were uploading the crashed data and they would assume that it went well if we won't tell them otherwise. So they could come back to us after a while to ask if the issue got resolved or something. At that point, we'd try to find the data associated with that user and we wouldn't find any. It's hypothetical and I don't know if it's even possible to look for those crashes in such a way - but maybe it will be in the future? Either way, being honest about it here could save both parties from wasting some time on assuming that the operation has succeeded.

packages/replayio/src/commands/record.ts
bvaughn: > Right. So that means that it can't ever fail here. And since it's configured with silent: true we end up with a very broken situation in which we report ✔ Crash data uploaded successfully when the upload of the crashed data has failed.

Hmm. I tested the failure case for this PR before posting it, and we didn't incorrectly report upload success.

I'm having a bit of trouble keeping up with bouncing back and forth between PRs though. If you think this is broken, can you post a PR with a proposed change and/or demonstrate it failing? (I believe you. It just seems more efficient that way than chatting on an already merged PR.)

PR #385, opened 2024-04-13
Improve UX for runtime update prompt

bvaughn:
- [x] Change prompt text to remove superfluous details
- [x] Use `logPromise` helper to improve the download-in-progress messaging

### Successful update

https://github.com/replayio/replay-cli/assets/29597/f7787e1b-feca-4fbb-9a6a-19dc4245b956

### Successful after retry

https://github.com/replayio/replay-cli/assets/29597/c21f0ff2-8ccb-42d5-b413-4f4d6bb9ca45

### Failed install

https://github.com/replayio/replay-cli/assets/29597/429508de-b2c3-4226-88f5-fcfe55b47707

### Failed update

https://github.com/replayio/replay-cli/assets/29597/f567186a-97ad-40ff-ace7-abfa74627cb2

### DIFF
--- a/packages/replayio/src/commands/record.ts
+++ b/packages/replayio/src/commands/record.ts
@@ -19,7 +19,11 @@ registerCommand("record", { checkForRuntimeUpdate: true, requireAuthentication: 
 async function record(url: string = "about:blank") {
   const prevRecordings = await getRecordings();
 
-  await launchBrowser(url, { processGroupId: uuid() });
+  const didLaunchBrowser = await launchBrowser(url, { processGroupId: uuid() });
+  if (!didLaunchBrowser) {
+    // Browser is missing or somehow corrupted
+    await exitProcess(1);
+  }
 
   const recordingsAfter = await getRecordings();
 
--- a/packages/replayio/src/commands/record.ts
+++ b/packages/replayio/src/commands/record.ts
@@ -21,7 +21,8 @@ async function record(url: string = "about:blank") {
 
   const didLaunchBrowser = await launchBrowser(url, { processGroupId: uuid() });
   if (!didLaunchBrowser) {
-    // Browser is missing or somehow corrupted
+    console.error("Something went wrong trying to launch the Replay browser.");
+
     await exitProcess(1);
   }
 
--- a/packages/replayio/src/utils/installation/installLatestRelease.ts
+++ b/packages/replayio/src/utils/installation/installLatestRelease.ts
@@ -2,9 +2,11 @@ import { spawnSync } from "child_process";
 import { ensureDirSync, renameSync, rmSync, unlinkSync, writeFileSync } from "fs-extra";
 import { get } from "https";
 import { join } from "path";
+import { logPromise } from "../async/logPromise";
+import { timeoutAfter } from "../async/timeoutAfter";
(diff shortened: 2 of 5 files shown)


packages/replayio/src/utils/installation/installLatestRelease.ts
```diff
@@ -65,16 +74,10 @@ async function downloadReplayFile() {
   };
 
   for (let i = 0; i < 5; i++) {
-    console.log(
-      `Downloading ${highlight(runtimeMetadata.runtime)} from ${link("replay.io")}${
-        i ? ` ${dim(`(retry ${i} of 4)`)}` : ""
-      }`
-    );
-
```
bvaughn, line 73: We lose a little bit of info here but I think in the overwhelming common case, the `logPromise` approach is a better UX.

packages/replayio/src/utils/installation/installLatestRelease.ts
```diff
@@ -15,11 +17,33 @@ export async function installLatestRelease() {
...
+        return `${prefix}${suffix}...`;
+      },
+      success: "Replay browser has been updated.",
+    },
+  });
+
+  const buffers = await downloadPromise;
+
   debug("Removing previous installation at %s", runtimePath);
   rmSync(runtimePath, { force: true, recursive: true });
```
bvaughn, line 45: We shouldn't remove the old installation until we've successfully downloaded the new one, else a failed update will make it so you can't even use Replay anymore.

jasonLaster:
Looks great

packages/replayio/src/commands/record.ts
```diff
@@ -19,7 +19,11 @@ registerCommand("record", { checkForRuntimeUpdate: true, requireAuthentication:
 async function record(url: string = "about:blank") {
   const prevRecordings = await getRecordings();
 
-  await launchBrowser(url, { processGroupId: uuid() });
+  const didLaunchBrowser = await launchBrowser(url, { processGroupId: uuid() });
+  if (!didLaunchBrowser) {
+    // Browser is missing or somehow corrupted
+    await exitProcess(1);
```
Andarist, line 25: This should be accompanied by some log - even if it's just "Something whent terribly wrong". Right now the process exits and the terminal stays blank - leaving me confused as to what has happened.

Andarist (approved):

packages/replayio/src/commands/record.ts
Andarist: I addressed this as part of https://github.com/replayio/replay-cli/pull/386 (that temporarily targets your PR)

bvaughn: I'm guessing you just made the launch command throw to test this?

I tested it, and noticed that we would have already printed an error when the upgrade failed (which we would have required, if the install was missing). But I guess you're right that maybe the file could be there but fail to launch for some other reason.

I've added another explicit message here.

Andarist: > I'm guessing you just made the launch command throw to test this?

Yes. I temporarily replaced the condition to always enter that branch

bvaughn:
Going to merge this. We can merge your other branch once the review is done.

PR #386, opened 2024-04-15
Ensure that runtime crashes are reported to the console

Andarist:


packages/replayio/src/utils/spawnProcess.ts
```diff
@@ -1,18 +1,23 @@
...
+  return buffers;
+}
+
 export function spawnProcess(
   executablePath: string,
   args: string[] = [],
   options: SpawnOptions = {},
-  callbacks: {
-    onError?: (error: Error) => void;
-    onExit?: () => void;
```
Andarist, line 10: We were already creating a deferred in `spawnProcess` (but never actually using it). So those 2 felt redundant and I decided to simplify this and lean on the returned `deferred.promise` in the caller

packages/replayio/src/utils/spawnProcess.ts
```diff
@@ -29,30 +34,28 @@ export function spawnProcess(
     // github.com/replayio/replay-cli/pull/344#discussion_r1553258356
     spawned.unref();
   } else {
+    const stderr = collectData(spawned.stderr);
```
Andarist, line 37: this was lost as we were never listening to `stderr` so we had no chance to surface what was put there by the launched browser

packages/replayio/src/utils/spawnProcess.ts
```diff
@@ -29,30 +34,28 @@ export function spawnProcess(
...
 
     spawned.on("spawn", () => {
-      if (onSpawn) {
-        onSpawn();
-      }
+      onSpawn?.();
     });
 
     spawned.on("exit", (code, signal) => {
-      if (code || signal) {
```
Andarist, line 47: According to the node docs we were basically always rejecting this deferred here:
> The 'exit' event is emitted after the child process ends. If the process exited, code is the final exit code of the process, otherwise null. If the process terminated due to receipt of a signal, signal is the string name of the signal, otherwise null. One of the two will always be non-null.

I wonder if that's the reason why we have the code that silences the uncaught promises (this one was always uncaught!)

packages/replayio/src/utils/prompt/prompt.ts
```diff
@@ -39,9 +40,9 @@ export async function prompt({
...
 
     stdin.on("data", onData);
-
-    if (abortSignal) {
-      abortSignal.addEventListener("abort", destroy);
-    }
+    signal?.addEventListener("abort", () => {
+      destroy();
+      resolve(false);
+    });
```
bvaughn, line 46: You're now adding something different than what you're removing above.

packages/replayio/src/utils/spawnProcess.ts
bvaughn: Kind of a nit but why a separate method for this? It's only ever referenced once, and the other process event handlers are inline methods.

packages/replayio/src/utils/spawnProcess.ts
bvaughn: > I wonder if that's the reason why we have the code that silences the uncaught promises (this one was always uncaught!)

No. We were calling "rejectIfPending" and "resolveIfPending" which are no-ops if the deferred is not pending.

The reason we had _that_ code was in case you CTRL+C while a _prompt_ (like the Y/n prompts or select-recordings list) was active (bc those are promises)

packages/replayio/src/utils/prompt/prompt.ts
Andarist: 🤦 good catch!

packages/replayio/src/utils/spawnProcess.ts
```diff
@@ -29,30 +34,28 @@ export function spawnProcess(
...
-      if (onError) {
-        onError(error);
-      }
     });
 
     spawned.on("spawn", () => {
-      if (onSpawn) {
-        onSpawn();
-      }
+      onSpawn?.();
```
bvaughn, line 44: Totally subjective and fine to leave this but oof I hate this syntax 😆

packages/replayio/src/utils/spawnProcess.ts
```diff
@@ -29,30 +34,28 @@ export function spawnProcess(
...
+        const buffered = Buffer.concat(stderr).toString();
+
+        let message = `Process failed (code: ${code})`;
+        if (buffered.length) {
+          message += `:\n${buffered}`;
+        }
+        deferred.rejectIfPending(new Error(message));
+        return;
       }
+      deferred.resolveIfPending();
```
bvaughn, line 58: Nit (feel free to ignore)
```suggestion
      } else {
        deferred.resolveIfPending();
      }
```

bvaughn (approved):
Left some nits but this looks good

packages/replayio/src/utils/spawnProcess.ts
Andarist: Right, but this deferred was pending at the time when this got called. 

So the scenario I'm thinking about goes like this:
1. we got ENTER to confirm the prompt
2. that killed the browser process
3. since it got killed we got `signal` here and we called `deferred.rejectIfPending`
4. nothing was `await`ing the `deferred` returned from this function
5. that could result in an unhandled rejection

Note that I really didn't test it out but the steps above do check out logically

packages/replayio/src/utils/spawnProcess.ts
Andarist: no particular reason, it could be reused to collect `stdout` at some point but we don't need it rn

packages/replayio/src/utils/spawnProcess.ts
Andarist: This guy is sad to hear that 😢 
?.(); put vertically, question mark at the top, semicolon at the bottom. It resembles a guy with an open mouth

packages/replayio/src/utils/spawnProcess.ts
bvaughn: Hmm, I just tested that theory (by removing the conditional in `uncaughtException`) and it didn't print anything so I dunno.

packages/replayio/src/utils/spawnProcess.ts
bvaughn: ༼ ༎ຶ ෴ ༎ຶ༽

packages/replayio/src/utils/spawnProcess.ts
Andarist: I couldn't make an actual unhandledrejection error logged into the console - but after removing `process.on("uncaughtException", ...)` handler and a noop `.catch` in `createDeferred` I managed to make the process exit (with an error logged into the console).

To trigger this situation in combination with `spawnProcess` I had to force this deferred to reject on exit... cause even when it's `.kill`ed we still get `code === 0` from Replay Chromium (which doesn't match my expectation cause it should re-raise the received signal or something like that).

Anyway, I'm just posting this as interesting trivia and nothing more right now 😉

PR #387, opened 2024-04-15
Log information about failed crash uploads

Andarist:
Follow up to the discussion [here](https://github.com/replayio/replay-cli/pull/384#discussion_r1565421735)

The situation can be mocked with this temp patch:
<details>
<summary>git diff</summary>

```diff
diff --git a/packages/replayio/src/utils/recordings/upload/uploadRecordings.ts b/packages/replayio/src/utils/recordings/upload/uploadRecordings.ts
index 04e14c9..7256a48 100644
--- a/packages/replayio/src/utils/recordings/upload/uploadRecordings.ts
+++ b/packages/replayio/src/utils/recordings/upload/uploadRecordings.ts
@@ -57,7 +57,17 @@ export async function uploadRecordings(

   const deferredActions = recordings.map(recording => {
     if (recording.recordingStatus === "crashed") {
-      return createSettledDeferred<LocalRecording>(recording, uploadCrashedData(client, recording));
+      // return createSettledDeferred<LocalRecording>(recording, uploadCrashedData(client, recording));
+      return createSettledDeferred<LocalRecording>(
+        recording,
+        new Promise((_, reject) => {
+          setTimeout(() => {
+            recording.uploadStatus = "failed";
+            reject(new Error("Couldnt upload it"));
+          }, 1000);
+        })
+      );
     } else {
       return createSettledDeferred<LocalRecording>(
         recording,
```
</details>

Currently, if the crash upload fails we still see this output:
```
> node ./packages/replayio/replayio.js record https://app.replay.io/team/me/recordings
Recording (press any key to stop recording)

It looks like something went wrong with this recording. Please hold while we upload crash data.
✔ Crash data uploaded successfully
```

With this PR we replace that with information that some of the crash upload data couldn't be uploaded

### DIFF
--- a/packages/replayio/src/commands/record.ts
+++ b/packages/replayio/src/commands/record.ts
@@ -10,6 +10,7 @@ import { selectRecordings } from "../utils/recordings/selectRecordings";
 import { LocalRecording } from "../utils/recordings/types";
 import { uploadRecordings } from "../utils/recordings/upload/uploadRecordings";
 import { logPromise } from "../utils/async/logPromise";
+import { canUpload } from "../utils/recordings/canUpload";
 
 registerCommand("record", { checkForRuntimeUpdate: true, requireAuthentication: true })
   .argument("[url]", `URL to open (default: "about:blank")`)
@@ -44,6 +45,8 @@ async function record(url: string = "about:blank") {
       "It looks like something went wrong with this recording. Please hold while we upload crash data."
     );
 
+    const uploadableCrashes = nextCrashedRecordings.filter(canUpload);
+
     const promise = uploadRecordings(nextCrashedRecordings, {
       processAfterUpload: false,
       silent: true,
@@ -52,7 +55,12 @@ async function record(url: string = "about:blank") {
     logPromise(promise, {
       messages: {
         pending: "Uploading crash data...",
-        success: "Crash data uploaded successfully",
+        success: () => {
+          if (uploadableCrashes.some(recording => recording.uploadStatus === "failed")) {
+            return "Some of the crash data couldn't be uploaded";
+          }
+          return "Crash data uploaded successfully";
+        },
       },
     });
 
--- a/packages/replayio/src/commands/record.ts
+++ b/packages/replayio/src/commands/record.ts
@@ -57,7 +57,7 @@ async function record(url: string = "about:blank") {
         pending: "Uploading crash data...",
         success: () => {
           if (uploadableCrashes.some(recording => recording.uploadStatus === "failed")) {
-            return "Some of the crash data couldn't be uploaded";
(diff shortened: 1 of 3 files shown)


packages/replayio/src/commands/record.ts
```diff
@@ -52,7 +55,12 @@ async function record(url: string = "about:blank") {
     logPromise(promise, {
       messages: {
         pending: "Uploading crash data...",
-        success: "Crash data uploaded successfully",
+        success: () => {
+          if (uploadableCrashes.some(recording => recording.uploadStatus === "failed")) {
+            return "Some of the crash data couldn't be uploaded";
+          }
+          return "Crash data uploaded successfully";
+        },
```
bvaughn, line 63: I kind of love how useful the little `logPromise` helper is proving to be 😁

Andarist, line 63: This isn't exactly beautiful and I want to rewrite this somehow. 

Atm I'm a little bit brain-farted and can't figure out how to juggle this code better - I might consider making the `uploadRecodings` a deferred action to get synchronous access to the "return value". However, even that won't fix everything - the problem is that `uploadRecordings` "settles" everything so the final result is always a success (and that comes with a green tick in the generated message). I want to avoid adding some clunky rendering options.

bvaughn (approved):
Awesome. Thanks for the follow up!

packages/replayio/src/commands/record.ts
bvaughn: > Atm I'm a little bit brain-farted and can't figure out how to juggle this code better 

Lemme know if I can help any!

packages/replayio/src/commands/record.ts
```diff
@@ -52,7 +55,12 @@ async function record(url: string = "about:blank") {
     logPromise(promise, {
       messages: {
         pending: "Uploading crash data...",
-        success: "Crash data uploaded successfully",
+        success: () => {
+          if (uploadableCrashes.some(recording => recording.uploadStatus === "failed")) {
+            return "Some of the crash data couldn't be uploaded";
```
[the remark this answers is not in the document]
Andarist: "✔ Crash data partially uploaded" - this doesn't like a failure to me so it could go a little bit unnoticed. Maybe something like "Crash data could only be partially uploaded"?

Either way, I'm fine with either - you have a better sense of English than I do :D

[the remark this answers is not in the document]
bvaughn: Sure. I don't feel strongly about either one.

packages/replayio/src/commands/record.ts
Andarist: Useful stuff :)

PR #388, opened 2024-04-15
Start processing uploaded recordings

bvaughn:
Preserve the option to block on processing (in case we want to re-add it later) but also add the option to start the processing operation without blocking.

### DIFF
--- a/packages/replayio/src/utils/recordings/upload/uploadRecording.ts
+++ b/packages/replayio/src/utils/recordings/upload/uploadRecording.ts
@@ -140,6 +140,9 @@ export async function uploadRecording(
     case "start-processing": {
       debug("Start processing recording %s ...", recording.id);
 
+      // In this code path, we intentionally don't update the "processingStatus" nor the recording log
+      // because this would interfere with how the recordings are printed when the upload has finished
+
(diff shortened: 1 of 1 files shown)


packages/replayio/src/utils/recordings/upload/uploadRecording.ts
```diff
@@ -135,40 +136,51 @@ export async function uploadRecording(
...
 
-  if (processAfterUpload) {
-    debug("Begin processing recording %s ...", recording.id);
+  switch (processingBehavior) {
+    case "start-processing": {
+      debug("Start processing recording %s ...", recording.id);
 
-    try {
-      await client.waitUntilAuthenticated();
+      processRecording(client, { recordingId: recording.id }).catch(error => {
```
Andarist, line 143: it might be a good idea to add a comment here that we don't want to set processing statuses here as that would hurt the renderered results (they could get stuck in unfinished states etc as we are only waiting for uploads there)

Andarist (approved):

packages/replayio/src/utils/recordings/upload/uploadRecording.ts
bvaughn: Yeah good call.

PR #391, opened 2024-04-15
Decouple `postinstall` execution from the package manager

Andarist:
I tested this with npm, yarn, pnpm and bun. Bun won't run postinstall scripts in our package but it sets `process.env.npm_node_execpath` correctly~ (from what I can tell it sets it to node and not bun unless you pass some extra options).

This isn't a particularly popular thing to reach for `$` variables within `package.json`... but it's standard and supported. Nevertheless, I feel a little bit weird proposing this 😅

### DIFF
--- a/packages/cypress/package.json
+++ b/packages/cypress/package.json
@@ -17,9 +17,7 @@
     "*.d.ts"
   ],
   "scripts": {
-    "if:exists": "node -e \"require('fs').existsSync(process.argv[1]) ? require('child_process').spawnSync(process.argv[2], process.argv.slice(3), {shell: true, stdio: 'inherit'}) : 0\"",
-    "if:dist": "yarn run if:exists -- dist",
-    "postinstall": "yarn run if:dist -- node ./bin.js first-run",
+    "postinstall": "\"$npm_node_execpath\" ./postinstall.js",
     "prepare": "npm run build",
     "build": "rm -rf dist/ tsconfig.tsbuildinfo && tsc",
     "test": "echo \"Error: no test specified\"",
--- a/packages/cypress/postinstall.js
+++ b/packages/cypress/postinstall.js
@@ -1,0 +1,11 @@
+const { existsSync } = require("fs");
+if (existsSync("dist")) {
+  require("child_process").spawnSync(
+    `"${process.env.npm_node_execpath}"`,
+    ["./bin.js", "first-run"],
+    {
+      shell: true,
+      stdio: "inherit",
+    }
+  );
+}
--- a/packages/playwright/package.json
+++ b/packages/playwright/package.json
@@ -17,9 +17,7 @@
     "*.d.ts"
   ],
   "scripts": {
-    "if:exists": "node -e \"require('fs').existsSync(process.argv[1]) ? require('child_process').spawnSync(process.argv[2], process.argv.slice(3), {shell: true, stdio: 'inherit'}) : 0\"",
-    "if:dist": "yarn run if:exists -- dist",
-    "postinstall": "yarn run if:dist -- node ./bin.js first-run",
+    "postinstall": "\"$npm_node_execpath\" ./postinstall.js",
     "prepare": "yarn run build",
     "build": "rm -rf dist/ tsconfig.tsbuildinfo && tsc",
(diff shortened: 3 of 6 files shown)


ryanjduffy (approved):
I'm fine with this but had a couple comments:

* do we need to use the `$` variable or is `node` already expected to be in the path when running an npm script?
* the `first-run` scripts are only called during post-install so we could refactor that code into `postinstall.js` and make it a straightforward node script rather than spawning a new process just to invoke `first-run`.

Andarist:
> do we need to use the $ variable or is node already expected to be in the path when running an npm script?

They are not guaranteed to match. It's better to use the `$` variable because this way we "inherit" the node version that was used for execution etc. Think about `replay-node` :) With this we'd call `replay-node` again - which is IMHO desirable.

> the first-run scripts are only called during post-install so we could refactor that code into postinstall.js and make it a straightforward node script rather than spawning a new process just to invoke first-run.

Goot shout 👍 I refactored this now - could you take another look?

There is also a relevant question here... should we even be running this as postinstall script? We are getting flagged by Socket.io and other tools as a potential threat. Postinstall scripts don't have a good rep. It would be a breaking change to change this right now though - so it's definitely not something that we can change here.

Andarist:
Argh, I see that the CI isn't particularly happy about the last change - I need to take a look at this.

ryanjduffy:
> They are not guaranteed to match. It's better to use the `$` variable because this way we "inherit" the node version that was used for execution etc. Think about `replay-node` :) With this we'd call `replay-node` again - which is IMHO desirable.

Makes sense!

> There is also a relevant question here... should we even be running this as postinstall script? We are getting flagged by Socket.io and other tools as a potential threat. Postinstall scripts don't have a good rep. It would be a breaking change to change this right now though - so it's definitely not something that we can change here.

I did this to follow playwright's lead (at the time, at least) to ensure that the browsers were installed when the tests ran. I think this is something we should consider changing in the next major release.

crutchcorn:
Hey all, I noticed that this change doesn't seem to work on Windows.

```shell
↳  pnpm install @replayio/playwright@alpha -D
../../angular/array                      |  WARN  deprecated reflect-metadata@0.2.1
../../..                                 |  +27 +++
../../..                                 | Progress: resolved 1667, reused 1535, downloaded 20, added 27, done
../../../node_modules/.pnpm/@replayio+playwright@3.0.0-alpha.10_@playwright+test@1.44.0/node_modules/@replayio/playwright: Running postinstall script, failed in 27ms
.../node_modules/@replayio/playwright postinstall$ "$npm_node_execpath" ./first-run.js
│ '$npm_node_execpath' is not recognized as an internal or external command,
│ operable program or batch file.
└─ Failed in 27ms at C:\Users\crutchcorn\git\TanStack\form\node_modules\.pnpm\@replayio+playwright@3.0.0-alpha.10_@playwright+test@1.44.0\node_modules\@replayio\playwright
 ELIFECYCLE  Command failed with exit code 1.
 ```

**Comment by Andarist** [comment]
Do u happen to know how to invoke an executable located at a path kept in env var on Windows? I think we should remove those postinstall scripts anyway but it’s a breaking change that I tried to avoid here

**Comment by crutchcorn** [comment]
I do:

```shell
"postinstall": "run-script-os",
"postinstall:default": "\"$npm_node_execpath\" ./first-run.js",
"postinstall:win32": "\"%npm_node_execpath%\" ./first-run.js",
```

But the bigger problem is that those env vars don't appear on my system at all, even.

Andarist:
Those postinstall scripts are currently published in the latest versions too. 

> But the bigger problem is that those env vars don't appear on my system at all, even. 

Hm, maybe it's related to how you run this on Windows? IIRC we don't even have Chromium builds for Windows so you are likely to run this in some emulated environment.

This is a standard env var that should be set by the package manager - but well, it's on the verge of being rarely depended on etc so I guess we are discovering edge cases now 😢

crutchcorn:
I don't think it is related to how I run on Windows. I could be wrong but I think I am running a native Windows build of Replay Chromium. Or at least I'm not running the CLI through WSL2

Either way, yeah this is rough

PR #393, opened 2024-04-16
Remove authentication requirement from the `update` command

Andarist:
Authentication requirement makes it harder to use on the CI. We don't really need this as nothing in this command doesn't truly depend on the authenticated user.


bvaughn (approved):
Yeah that's fair.

PR #395, opened 2024-04-16
Fixed base dir of the renamed location on Linux

Andarist:
It currently throws at Linux with:
> Error: ENOENT: no such file or directory, rename '/home/runner/.replay/runtimes/chrome-linux/replay-chromium' -> '/home/runner/.replay/runtimes/chrome-linux/chrome-linux'

We might notice here that `chrome-linux` is included both in the source and the destination whereas this renaming is conceptually meant to create `chrome-linux` directory


bvaughn (approved):
Ooof, silly of me.

PR #398, opened 2024-04-16
Introduce `logProgress` for improved control over rendering pending tasks

Andarist:
this PR addresses what I brain-farted on yesterday (see https://github.com/replayio/replay-cli/pull/387#discussion_r1565682435 )

When some crash data fails to upload we should now log this:
```diff
-✔ Crash data could only be partially uploaded
+✘ Crash data could only be partially uploaded
```

In addition to that, I noticed today that we are displaying this
```
✔ Replay browser has been updated.
```
before we even get to the end of `installLatestRelease`. So I reused the `logProgress` infra to fix this too. This success message is now displayed at the very end of `installLatestRelease` and after we successfully download the browser archive we update the progress message to:
```
Processing downloaded browser archive
```

I'm opening this as a draft because I'm wrapping up for today and I didn't have the time to test this properly yet. Feedback is welcome though :)


packages/replayio/src/utils/async/logProgress.ts
```diff
@@ -0,0 +1,78 @@
...
+  };
+
+  return {
+    setFailed: (message: string) => {
+      assertPending();
+      status = STATUS_REJECTED;
+      displayedMessage = message;
+      finalize();
+    },
+    setProgress: (message: string) => {
```
bvaughn, line 66: I think this should be "pending" to stay inline with other deferred statuses.
```suggestion
    setPending: (message: string) => {
```

IMO "progress" is a broader term that could also encompass "done" states

packages/replayio/src/utils/installation/installLatestRelease.ts
```diff
@@ -11,82 +11,91 @@ import { debug } from "./debug";
 import { getLatestRelease } from "./getLatestReleases";
 import { MetadataJSON } from "./types";
 
+function getPendingDownloadMessage(attempt: number) {
```
bvaughn, line 14: This method should probably also accept the `maxAttempts` param. Else it would be easy to change the 5 below and leave these two values in conflict.

packages/replayio/src/utils/installation/installLatestRelease.ts
```diff
@@ -11,82 +11,91 @@ import { debug } from "./debug";
...
+  const prefix = `Downloading from ${link("static.replay.io")}`;
+  let suffix = "";
+  if (attempt > 0) {
+    suffix = `${dim(` (retry ${attempt} of 4)`)}`;
+  }
+
+  return `${prefix}${suffix}...`;
```
bvaughn, line 21: How do you feel about?
```suggestion
    suffix = dim(`(retry ${attempt} of 4)`);
  }

  return `${prefix}... ${suffix}`;
```

That way the message wouldn't "jump" after retry.

packages/replayio/src/utils/installation/installLatestRelease.ts
```diff
@@ -11,82 +11,91 @@ import { debug } from "./debug";
...
-  logPromise(downloadPromise, {
-    messages: {
-      failed: "Something went wrong installing the Replay browser. Please try again later.",
-      pending: () => {
-        const prefix = `Downloading from ${link("static.replay.io")}`;
-        let suffix = "";
-        if (metadata.attemptNumber > 1) {
-          suffix = `${dim(` (retry ${metadata.attemptNumber - 1} of 4)`)}`;
-        }
+    progress.setProgress("Processing downloaded browser archive");
```
bvaughn, line 38: This is nice 👍🏼

packages/replayio/src/utils/async/logProgress.ts
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  const finalize = () => {
+    clearInterval(interval);
+    print();
+    logUpdate.done();
+  };
+
+  return {
+    setFailed: (message: string) => {
```
bvaughn, line 60: Should we add an explicit `clear` method?

I think calling any of the above methods with an empty string _would_ have the same effect but may not be intuitive.

bvaughn (approved):
I like this change.

packages/replayio/src/utils/installation/installLatestRelease.ts
Andarist: good idea 👍

bvaughn:
Since you're out for a few days, I'm going to apply my own PR feedback and land this. (I did some testing locally too.)

PR #405, opened 2024-04-17
Bring back logging stderr in case of a runtime crash

Andarist:
I only made some basic tests with this - please test this more before merging ;p

### DIFF
--- a/packages/replayio/src/utils/spawnProcess.ts
+++ b/packages/replayio/src/utils/spawnProcess.ts
@@ -39,19 +39,25 @@ export function spawnProcess(
       onSpawn?.();
     });
 
-    if (printStderr) {
-      spawned.stderr?.setEncoding("utf8");
-      spawned.stderr?.on("data", printStderr);
-    }
+    let stderr = "";
+    spawned.stderr?.setEncoding("utf8");
+    spawned.stderr?.on("data", (data: string) => {
+      stderr += data;
+      printStderr?.(data);
+    });
 
     if (printStdout) {
       spawned.stdout?.setEncoding("utf8");
       spawned.stdout?.on("data", printStdout);
     }
 
-    spawned.on("exit", code => {
-      if (code) {
-        deferred.rejectIfPending(new Error(`Process failed (code: ${code})`));
+    spawned.on("exit", (code, signal) => {
+      if (code || signal) {
+        let message = `Process failed (${code ? `code: ${code}` : `signal: ${signal}`})`;
+        if (stderr.length) {
+          message += `:\n${stderr}`;
+        }
+        deferred.rejectIfPending(new Error(message));
       } else {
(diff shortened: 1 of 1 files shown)


packages/replayio/src/utils/spawnProcess.ts
```diff
@@ -39,19 +39,25 @@ export function spawnProcess(
...
     }
 
-    spawned.on("exit", code => {
-      if (code) {
-        deferred.rejectIfPending(new Error(`Process failed (code: ${code})`));
+    spawned.on("exit", (code, signal) => {
+      if (code || signal) {
+        let message = `Process failed (${code ? `code: ${code}` : `signal: ${signal}`})`;
+        if (stderr.length) {
+          message += `:\n${stderr}`;
```
bvaughn, line 58: I don't love this approach. IMO an error message should be concise. This will make it potentially huge and it could contain a lot of unrelated junk (based on what I see our browser logging to stderr at times).

bvaughn: I wonder if it would be nicer for us to keep the error message short and print the output of `stderr` to a log (which we could also tell the user about).

Andarist: > and print the output of stderr to a log (which we could also tell the user about).

i like this approach - how do u clean up those files though? Personally i dont mind the current approach, it might be noisy but at least it doesnt require extra steps from me. Having to look in a separate file feels often like a chore when i encounter this flow in other tools

bvaughn: > how do u clean up those files though?

Perhaps the CLI could always check and delete log files older than some value (or even just all of them) whenever it runs.

> but at least it doesnt require extra steps from me

That's fair. Lemme sit on this another few minutes. I'm not totally against it, it was just a preference.

Andarist: I agree that the runtime spits out some junk too - but that could be addressed on their side

bvaughn: It seems like they (intentionally) only log to `stderr` – so the stuff there might not even be error related at all.

bvaughn: I wonder if the `--verbose` flag (#407) takes the pressure off of this a bit

bvaughn: It seems like they (intentionally) only log to `stderr` – so the stuff there might not even be error related at all.

I wonder if the addition of a `--verbose` flag (#407) makes this change any less important (or if we'd still want to do it regardless). I don't love a really noisy error message. I do think logging the info _somewhere_ is useful, but I think a (temporary) log file might be a more standard way of doing this, given how noisy the `stderr` text might be.

PR #408, opened 2024-04-18
Log recorder crash info to disk

bvaughn:
Proposed alternative to #405

Screenshot 2024-04-18 at 8 34 13 AM

### DIFF
--- a/packages/replayio/src/commands/record.ts
+++ b/packages/replayio/src/commands/record.ts
@@ -1,9 +1,12 @@
+import { writeFileSync } from "fs-extra";
 import { v4 as uuid } from "uuid";
+import { ProcessError } from "../utils/ProcessError";
 import { logAsyncOperation } from "../utils/async/logAsyncOperation";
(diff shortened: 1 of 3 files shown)


packages/replayio/src/commands/record.ts
```diff
@@ -20,7 +24,26 @@ registerCommand("record", { checkForRuntimeUpdate: true, requireAuthentication:
...
   const prevRecordings = await getRecordings();
 
-  await launchBrowser(url, { processGroupId: uuid() });
+  try {
+    await launchBrowser(url, { processGroupId: uuid() });
+  } catch (error) {
+    if (error instanceof ProcessError) {
+      console.log("\nSomething went wrong while recording. Try again.");
+
+      // TODO [PRO-235] Upload recorder crash data somewhere
```
bvaughn, line 33: Filed PRO-235 for follow up here because I think it's worth doing eventually/soon.

jazzdan (approved):

PR #410, opened 2024-04-18
Better handle when Replay browser is already running

bvaughn:
### [Loom explanation](https://www.loom.com/share/427b41ba94694e41a76525d94cdc4cdf)

Screenshot 2024-04-18 at 8 34 13 AM

### DIFF
--- a/packages/replayio/src/utils/browser/launchBrowser.ts
+++ b/packages/replayio/src/utils/browser/launchBrowser.ts
@@ -1,10 +1,13 @@
+import findProcess from "find-process";
 import { ensureDirSync, existsSync } from "fs-extra";
 import { join } from "path";
+import { retryWithLinearBackoff } from "../async/retry";
+import { timeoutAfter } from "../async/timeoutAfter";
 import { getReplayPath } from "../getReplayPath";
 import { runtimeMetadata, runtimePath } from "../installation/config";
 import { prompt } from "../prompt/prompt";
 import { spawnProcess } from "../spawnProcess";
-import { dim, stderrPrefix, stdoutPrefix } from "../theme";
+import { dim, highlight, stderrPrefix, stdoutPrefix } from "../theme";
 import { debug } from "./debug";
 import { getBrowserPath } from "./getBrowserPath";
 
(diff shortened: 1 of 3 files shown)


packages/replayio/src/utils/browser/launchBrowser.ts
```diff
@@ -42,42 +45,67 @@ export async function launchBrowser(
...
-      debug(stderrPrefix("stderr"), text);
-    },
-    printStdout: (text: string) => {
-      debug(stdoutPrefix("stdout"), text);
-    },
-  });
+    // Ask the browser to open a new tab
+    spawnProcess(browserExecutablePath, args, processOptions);
+
+    // The best we can do in this case is to regularly poll to see when the process exits
```
Andarist, line 59: Currently we check the recording log before launching the browser and then compare that against the log that we get after it. I'm worried that this will lead to problems because both CLI jobs will print overlapping recordings - and then u could already pick a recording A to be uploaded in the first CLI and then pick the same in the second CLI. I'd expect that situation to not be handled well today.

I agree that if we want to proceed... this is probably the best we can do. But I also wonder if it's worth proceeding in this case.

packages/replayio/src/utils/browser/launchBrowser.ts
```diff
@@ -42,42 +45,67 @@ export async function launchBrowser(
...
+      async () => {
+        await timeoutAfter(1_000);
+        const processes = await findProcess("name", browserExecutablePath);
+        if (processes.length > 0) {
+          throw Error("Still running");
+        }
+      },
+      undefined,
+      Number.POSITIVE_INFINITY
+    );
```
Andarist, line 70: nit: I'd probably prefer a simple loop here instead of throwing an error to trigger the retry. I feel like with a simple loop there would be less moving pieces and ultimately it would be simpler and easier to understand.

packages/replayio/src/utils/browser/launchBrowser.ts
bvaughn: I talk about this in the Loom. I don't think there's any way around it.

Andarist: What are your thoughts on exiting the CLI with an error when the user tries to do this? It feels like an edge case that the user tries to call the CLI more than once before they close the browser - is it worth introducing more edge cases on our side to accommodate this situation? 🤔

bvaughn: @Domiii and @jcmorrow asked for this change, so I feel like you'd probably want to run this by them more than me.

I think we'd need to keep _most of_ this code in place either way, because we'll still need to detect that the runtime is already open.

bvaughn: btw if you haven't already watched the Loom attached to this PR, it's worth giving a watch

Andarist: > I think we'd need to keep most of this code in place either way, because we'll still need to detect that the runtime is already open.

Totally 👍 

> btw if you haven't already watched the Loom attached to this PR, it's worth giving a watch

I've already watched it after you first responded to me - I was stalling it to avoid the risk of waking up my wife ;p 

> and then u could already pick a recording A to be uploaded in the first CLI and then pick the same in the second CLI.

What would happen today in this PR if somebody did this?

bvaughn: If they tried to upload the same recording twice?

Likely the 2nd attempt would fail and we'd print an error message. TBH I don't think that's a big deal.

Andarist: > If they tried to upload the same recording twice?

Yes.

> Likely the 2nd attempt would fail and we'd print an error message. TBH I don't think that's a big deal.

If that's the case then I'm fine with it - i would just love to make sure that the error would be reported to the user one way or another.

bvaughn: Yeah it looks like we just show the duplicate as failed.

I tested in both directions.

Screenshot 2024-04-18 at 7 08 26 PM

Screenshot 2024-04-18 at 7 08 47 PM

bvaughn:
Made the `while` loop change you suggested. (I agree, that's nicer.) Otherwise, back to you for a check or an x.

Andarist (approved):

Andarist:
I’m still on the fence if we should proceed in this scenario but i dont want to block this

bvaughn:
I think it's slightly better to have this than not to have it, but I also feel a bit torn about it.

If you'd like to talk about it, we can.

PR #411, opened 2024-04-18
Report a better uploading errors to the user

ryanjduffy:
Passing `strict` to `uploadRecording` will cause it to throw on any error rather than returning `null` (which is a legacy behavior that should eventually change in a future breaking release). This allows us to show more detailed errors to the user like when they are being rate-limited.

### DIFF
--- a/packages/test-utils/src/reporter.ts
+++ b/packages/test-utils/src/reporter.ts
@@ -591,6 +591,9 @@ class ReplayReporter<TRecordingMetadata extends UnstructuredMetadata = Unstructu
     try {
       const result = await uploadRecording(recording.id, {
         apiKey: this.apiKey,
+        // Per TT-941, we want to throw on any error so it can be caught below
+        // and reported back to the user rather than just returning null
+        strict: true,
       });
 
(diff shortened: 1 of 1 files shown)


packages/test-utils/src/reporter.ts
```diff
@@ -591,6 +591,9 @@ class ReplayReporter<TRecordingMetadata extends UnstructuredMetadata = Unstructu
     try {
       const result = await uploadRecording(recording.id, {
         apiKey: this.apiKey,
+        // Per TT-941, we want to throw on any error so it can be caught below
+        // and reported back to the user rather than just returning null
+        strict: true,
```
Andarist, line 596: does it remove the need for `if (result === null)` that is below this?

miriambudayr:
Thank you so much 🙏

packages/test-utils/src/reporter.ts
ryanjduffy: Appears so. I've removed that as well.

Andarist (approved):

PR #417, opened 2024-04-29
Fix compatibility with new Playwright versions

Andarist:
I tested this on https://github.com/replayio-public/overboard using both old and new Playwright versions. My context on all of this is still limited so please be careful when reviewing 😉


packages/playwright/src/fixture.ts
```diff
@@ -255,26 +276,45 @@ export async function replayFixture(
...
+            error: step.error ? parseError(step.error) : null,
+          },
         });
       }
     }
   );
 
   const csiListener: ClientInstrumentationListener = {
-    onApiCallBegin: (apiName, params, stackTrace, _wallTime) => {
+    onApiCallBegin: (apiName, params, stackTraceOrFrames, _wallTime) => {
```
ryanjduffy, line 292: A utility fn that returns the appropriate `csiListener` for the current version of playwright might this more understandable vs growing the adapter for each internal API change

ryanjduffy (approved):
Looks reasonable based on my read of the code. I think we might consider a more resilient way to manage different playwright versions (not that the cypress plugin does this very well ...) but if the path forward isn't obvious, I think we could do this in a later milestone.

packages/playwright/src/fixture.ts
Andarist: I agree but this is a simple change and I think the code is still OK. It's also not completely apparent how to check the playwright version that is being used by the user. The `playwright` object that gets passed to `replayFixture` doesn't contain this information

PR #421, opened 2024-04-30
Export pollForAuthentication

hbenl:
I'm writing an e2e test for external browser auth and don't want to duplicate the code for fetching the refresh token from the backend.

### DIFF
--- a/packages/replayio/src/utils/authentication/authenticateByBrowser.ts
+++ b/packages/replayio/src/utils/authentication/authenticateByBrowser.ts
@@ -75,7 +75,7 @@ async function fetchRefreshTokenFromGraphQLOrThrow(key: string) {
   return data.closeAuthRequest.token as string;
 }
 
-async function pollForAuthentication(key: string) {
+export async function pollForAuthentication(key: string) {
   let refreshToken: string | undefined = undefined;
   while (!refreshToken) {
     try {


packages/replayio/src/utils/authentication/authenticateByBrowser.ts
```diff
@@ -75,7 +75,7 @@ async function fetchRefreshTokenFromGraphQLOrThrow(key: string) {
   return data.closeAuthRequest.token as string;
 }
 
-async function pollForAuthentication(key: string) {
+export async function pollForAuthentication(key: string) {
```
Andarist, line 78: this won't help you because you can't reach into this package for arbitrary files, we only allow importing from the specified entrypoints:
https://github.com/replayio/replay-cli/blob/03b8db0c7b5d04d21e7f7ac98822011961190027/packages/replayio/package.json#L6-L9

to make it work you'd have to export it from the index here:
https://github.com/replayio/replay-cli/blob/03b8db0c7b5d04d21e7f7ac98822011961190027/packages/replayio/src/index.ts

But honestly, I'm rather hesitant to add any exports to this package. The CLI files often might perform some initialization-time side-effects and might assume global/singleton context. It's quite easy to forget that utils might get imported by non-CLI consumers and we had a bunch of problems/regressions related to this (but in the old `@replayio/replay` package).

I would feel much better if we'd move reusable pieces to dedicated package(s) so the boundary between utils and the CLI itself would be much more apparent to people touching the relevant files.

hbenl: I think the `exports` field doesn't restrict what can be imported, it only specifies what can be imported from `"replayio"` but I can still import from `"replayio/utils/authentication/authenticateByBrowser"`.

> I'm rather hesitant to add any exports to this package

I also wouldn't want this export to show up in the `"replayio"` exports, but exporting it in this way is OK in my opinion (and many other utils are already exposed like that).

bvaughn: I'm also hesitant to add one-off exports. Whatever we do export becomes part of our public API so it should be well-defined and tested. I don't think `pollForAuthentication` feels like it fits that in its current state.

bvaughn: > I think the `exports` field doesn't restrict what can be imported, it only specifies what can be imported from `"replayio"` but I can still import from `"replayio/utils/authentication/authenticateByBrowser"`.

IMO we shouldn't do deep imports like that.

Andarist: > I think the exports field doesn't restrict what can be imported, it only specifies what can be imported from "replayio" but I can still import from "replayio/utils/authentication/authenticateByBrowser".

That's not how it works. The `exports` manifest limits everything. For that reason we list `./package.json` there so `require('replayio/package.json')` can still work. Everything else essentially stays hidden as private. 

With `exports` you can even rewrite paths like:
```json
"exports": {
  "./lib/*/stuff": "./dist/*/blabla/deep/wow/awesome/yet/weird.js"
}
```

It acts as a **complete** manifest of available specifiers within the package ([docs](https://nodejs.org/api/packages.html#subpath-exports)):
> Now only the defined subpath in ["exports"](https://nodejs.org/api/packages.html#exports) can be imported by a consumer:

PR #422, opened 2024-04-30
Bump replayio to 1.0.6

hbenl:


packages/replayio/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "replayio",
-  "version": "1.0.5",
+  "version": "1.0.6",
```
bvaughn, line 3: The version bump is fine but the other change in this PR shouldn't land.

PR #423, opened 2024-04-30
Make fixture the only way to use the Playwright reporter

Andarist:


packages/playwright/src/reporter.ts
```diff
@@ -108,6 +108,7 @@ class ReplayPlaywrightReporter implements Reporter {
   > = {};
 
   constructor() {
+    addReplayFixture();
```
Andarist, line 111: this seems to work when called from here but the question is - was there a specific reason why it wasn't called from here?

Andarist: Cool, it's good to at least have some confirmation that there is no obvious blocker to move it here 👍

PR #424, opened 2024-05-01
Filter empty recordings

bvaughn:
Temporary workaround for TT-1036

### DIFF
--- a/packages/replayio/src/utils/recordings/getRecordings.ts
+++ b/packages/replayio/src/utils/recordings/getRecordings.ts
@@ -246,19 +246,26 @@ export function getRecordings(): LocalRecording[] {
 
   debug("Found %s recordings:\n%o", recordings.length, recordings);
 
-  // Sort recordings in reverse chronological order
-  // but group related recordings so that "root" recordings are always listed first
-  recordings.sort((a, b) => {
-    if (a.metadata.processGroupId === b.metadata.processGroupId) {
-      if (a.metadata.processType === "root") {
-        return -1;
-      } else if (b.metadata.processType === "root") {
-        return 1;
+  return recordings
+    .filter(recording => {
+      if (!recording.metadata.host) {
+        // Ignore new tab recordings (see TT-1036)
+        return false;
       }
-    }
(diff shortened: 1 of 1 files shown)


packages/replayio/src/utils/recordings/getRecordings.ts
```diff
@@ -246,19 +246,26 @@ export function getRecordings(): LocalRecording[] {
...
-  recordings.sort((a, b) => {
-    if (a.metadata.processGroupId === b.metadata.processGroupId) {
-      if (a.metadata.processType === "root") {
-        return -1;
-      } else if (b.metadata.processType === "root") {
-        return 1;
+  return recordings
+    .filter(recording => {
+      if (!recording.metadata.host) {
+        // Ignore new tab recordings (see TT-1036)
```
Andarist, line 252: I wonder if this shouldn't be combined with an extra check for the root like:
```suggestion
      if (!recording.metadata.host && recording.metadata.processType === "root") {
        // Ignore new tab recordings (see TT-1036)
```

Rendering src-less `<iframe />`s is a thing and I wouldn't like to accidentally filter recordings made there. Although, maybe those just don't create extra recordings since they don't need process isolation anyway? Even if that's the case the proposed check is probably erring this on the safer side.

Andarist (approved):

packages/replayio/src/utils/recordings/getRecordings.ts
bvaughn: That seems...reasonable. Yeah.

bvaughn: Actually, looking at this– we end up with a lot of empty iframe recordings from new tabs too, because of how Chrome loads the default content in this tabs. So that isn't a great UX.

bvaughn: Going to make a judgement call here to filter all "empty" recordings, regardless of their type. This seems like the least bad UX option.

To handle this right requires Chromium changes (TT-1036).

PR #426, opened 2024-05-02
Patch Playwright fixtures in the main entry point

Andarist:


Andarist:
@bvaughn I know that you don't have a lot of context on this but if u'd stamp this I could release a new version and try it out properly in our projects.

bvaughn (approved):

PR #427, opened 2024-05-02
Add error message when uploading fails due to missing apiKey

hbenl:


packages/test-utils/src/reporter.ts
```diff
@@ -594,6 +594,12 @@ class ReplayReporter<TRecordingMetadata extends UnstructuredMetadata = Unstructu
   async uploadRecording(recording: RecordingEntry): Promise<UploadPendingWork> {
     debug("Starting upload of %s", recording.id);
 
+    if (!this.apiKey) {
```
Andarist, line 597: Would it be possible to raise this early? Right now it seems like we’d still attempt to upload N recordings and we would just fail each one of those uploads immediately. 

Maybe we could check this in the constructor when `upload: true`?

hbenl: I don't think we can cancel the test run from the reporter (if we throw, Playwright will catch and log the error and then continue as if nothing happened), so it will still run the tests and upload their results to Replay.
I've changed the code to only emit one error message though (instead of one per recording).

packages/test-utils/src/reporter.ts
```diff
@@ -913,6 +913,13 @@ class ReplayReporter<TRecordingMetadata extends UnstructuredMetadata = Unstructu
       });
     }
 
+    if (this.recordingsToUpload.length && !this.apiKey) {
+      output.push(`\n❌ Failed to upload ${this.recordingsToUpload.length} recordings:\n`);
+      output.push(
+        "   Can't upload recordings without an API key - please set the REPLAY_API_KEY environment variable"
```
ryanjduffy, line 919:
```suggestion
        "   Can't upload recordings without an API key. Either pass a value to the apiKey plugin configuration or set the REPLAY_API_KEY environment variable"
```

ryanjduffy (approved):

packages/test-utils/src/reporter.ts
Andarist: I just tested this and throwing in our `constructor` goes straight to this `catch` handler and exits the whole process:
https://github.com/microsoft/playwright/blob/aa0020878ad51cf5894dd97527b74d456b1135fa/packages/playwright/src/program.ts#L47-L50

PR #428, opened 2024-05-02
Fixed scope mismatch between the fixture and the reporter

Andarist:


packages/playwright/src/fixture.ts
```diff
@@ -165,7 +165,11 @@ export async function replayFixture(
     attempt: testInfo.retry + 1,
     source: {
       title: testInfo.title,
-      scope: testInfo.titlePath.slice(3, -1),
+      // this one only drops the filename (first segment) and the test title (last segment)
+      // it's different from the one in the reporter, since the "root" suites are just the file suites created here:
+      // https://github.com/microsoft/playwright/blob/a6488c4a2879a22e8da0f6708114ef7b9f4d253f/packages/playwright/src/common/testLoader.ts#L36
+      // in this context they are not attached to the root and project suites
+      scope: testInfo.titlePath.slice(1, -1),
```
Andarist, line 172: I went back to Playwright 1.19 to verify that the behavior was the same even back then. The previous version worked only OK when there were no `describe` blocks

hbenl (approved):

PR #431, opened 2024-05-03
[wip] Add honeycomb event

jasonLaster:
This adds a `sendEvent` method for tracking errors in honeycomb. This is lifted from [recordData](https://github.com/search?type=code&q=org%3Areplayio+recordData+repo%3Areplayio%2Fdevtools) in DevTools.

Marking it as a WIP because i haven't tested it.

### DIFF
--- a/packages/replayio/src/commands/record.ts
+++ b/packages/replayio/src/commands/record.ts
@@ -16,6 +16,7 @@ import { selectRecordings } from "../utils/recordings/selectRecordings";
 import { LocalRecording } from "../utils/recordings/types";
 import { uploadRecordings } from "../utils/recordings/upload/uploadRecordings";
 import { dim } from "../utils/theme";
+import { sendEvent } from "../utils/telemetry";
 
 registerCommand("record", { checkForRuntimeUpdate: true, requireAuthentication: true })
   .argument("[url]", `URL to open (default: "about:blank")`)
@@ -46,6 +47,10 @@ async function record(url: string = "about:blank") {
         const errorLogPath = getReplayPath("recorder-crash.log");
 
         writeFileSync(errorLogPath, stderr, "utf8");
+
+        sendEvent("cli-recorder-crash", {
+          message: stderr,
+        });
 
         console.log(dim(`More information can be found in ${errorLogPath}`));
       }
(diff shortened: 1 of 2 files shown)


bvaughn (changes requested):
Let's talk on Monday. I'd rather get consensus from the TT team that this is the right place to send the data, and the right data to be sending (`stderr` may contain a lot of noise, and it may not contain useful info that's in a crashpad file)

jasonLaster:
Yep - will defer to @jcmorrow and TT

packages/replayio/src/utils/telemetry.ts
```diff
@@ -0,0 +1,47 @@
...
+): asserts assertion {
+  if (!assertion) {
+    sendEvent(type, {
+      message,
+      ...tags,
+    });
+  }
+
+  return assert(assertion, message);
+}
```
bvaughn, line 47: This method isn't being used:
```suggestion
```

packages/replayio/src/utils/telemetry.ts
```diff
@@ -0,0 +1,47 @@
+import assert from "assert";
+
+let defaultTags = {};
+export function setDefaultTags(tags: Object) {
+  defaultTags = tags;
+}
```
bvaughn, line 6: This method isn't being used:
```suggestion
```

packages/replayio/src/commands/record.ts
```diff
@@ -47,6 +48,10 @@ async function record(url: string = "about:blank") {
 
         writeFileSync(errorLogPath, stderr, "utf8");
 
+        sendEvent("cli-recorder-crash", {
+          message: stderr,
+        });
```
bvaughn, line 53: I think we'd probably want to send more info along with such a crash– the runtime build id for example, and potentially the CLI package version as well.

There may also be [crash dump files](https://chromium.googlesource.com/crashpad/crashpad) with data that isn't present in `stderr`. Chromium puts them [here](https://www.chromium.org/developers/crash-reports/#generating-crash-dumps-locally) by default but I'm not sure where Replay writes them. (This is something I left a question about on PRO-235/TT-983.)

bvaughn:

PR #432, opened 2024-05-04
Disable animated log output for non-TTY instances

bvaughn:
Maybe this is a good idea to always do?

### DIFF
--- a/packages/replayio/src/config.ts
+++ b/packages/replayio/src/config.ts
@@ -5,5 +5,6 @@ export const replayWsServer =
 
 const isCI = !!process.env.CI;
 const isDebugging = !!process.env.DEBUG;
+const isTTY = process.stdout.isTTY;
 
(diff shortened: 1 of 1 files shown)


packages/replayio/src/config.ts
```diff
@@ -5,5 +5,6 @@ export const replayWsServer =
 
 const isCI = !!process.env.CI;
 const isDebugging = !!process.env.DEBUG;
+const isTTY = process.stdout.isTTY;
 
-export const disableAnimatedLog = isCI || isDebugging;
+export const disableAnimatedLog = isCI || isDebugging || isTTY;
```
Andarist, line 10: When it would not be disabled now? `isTTY` is our standard mode, right? 🤔

bvaughn: 🤪 🔫 

Yeah. I did the exact opposite of what I intended to do with that commit.

Andarist (approved):

PR #433, opened 2024-05-05
fix: replayio login now works on Windows

crutchcorn:
This PR fixes an issue with `replayio login` on Windows caused by an `ENOENT` when using `start`.

Turns out that windows requires a `cmd /c start https://google.com` command prefix.

While I could have hardcoded a solution for this, I instead opted to use a broadly utilized package (`open`) from a known author (Sindre) instead.

> Side note, we should migrate this package to use ESM and upgrade this + the `node-fetch` API.

image


bvaughn (approved):
Nice debugging. I like this change.

PR #434, opened 2024-05-05
fix: no errors should occur when pressing a button to halt recording

crutchcorn:
When using the keyboard in the CLI to halt the Replay browser, I was getting the following error:

image

Without a relevant stacktrace/error log file.

This is because `process.kill` (triggered by the `prompt`) would send a `signal: "SIGTERM"`, which would be seen by this code as a failure.

This code edgecases `SIGTERM` to sidestep this problem.

### DIFF
--- a/packages/replayio/src/utils/spawnProcess.ts
+++ b/packages/replayio/src/utils/spawnProcess.ts
@@ -54,6 +54,11 @@ export function spawnProcess(
 
     spawned.on("exit", (code, signal) => {
       if (code || signal) {
+        // Don't fail on manual closing
+        if (signal === "SIGTERM") {
+          deferred.resolveIfPending();
+          return;
+        }
         const message = `Process failed (${code ? `code: ${code}` : `signal: ${signal}`})`;
 
         deferred.rejectIfPending(new ProcessError(message, stderr));


bvaughn (approved):

packages/replayio/src/utils/spawnProcess.ts
```diff
@@ -54,6 +54,11 @@ export function spawnProcess(
 
     spawned.on("exit", (code, signal) => {
       if (code || signal) {
+        // Don't fail on manual closing
+        if (signal === "SIGTERM") {
```
Andarist, line 58: I don't get `SIGTERM` re-raised by our Chromium here. When I close the browser by pressing any key in the terminal I get `code === 0 && signal === null` here. This should only depend on the Chromium's behavior... so I wonder why it's inconsistent 🤔

crutchcorn: That's interesting. FWIW it makes sense to me that it throws a SIGTERM given that the `kill` command reads as:

> The signal to send, either as a string or number. Default: 'SIGTERM'.

https://nodejs.org/api/process.html#processkillpid-signal

TBH I have only tested this on Windows, not macOS

bvaughn: > I don't get `SIGTERM` re-raised by our Chromium here. When I close the browser by pressing any key in the terminal I get `code === 0 && signal === null` here. This should only depend on the Chromium's behavior... so I wonder why it's inconsistent 🤔

Same, FWIW.

PR #437, opened 2024-05-06
Fail fast in reporters when there is no API key

Andarist:
kind of an alternative to https://github.com/replayio/replay-cli/pull/427

### DIFF
--- a/packages/playwright/src/reporter.ts
+++ b/packages/playwright/src/reporter.ts
@@ -78,34 +78,35 @@ export interface ReplayPlaywrightConfig
 
 class ReplayPlaywrightReporter implements Reporter {
   reporter?: ReplayReporter<ReplayPlaywrightRecordingMetadata>;
-  captureTestFile = ["1", "true"].includes(
-    process.env.PLAYWRIGHT_REPLAY_CAPTURE_TEST_FILE?.toLowerCase() || "true"
-  );
+  captureTestFile: boolean;
+  config: ReplayPlaywrightConfig;
+
+  constructor(config: ReplayPlaywrightConfig) {
+    if (!config || typeof config !== "object") {
+      throw new Error(
+        `Expected an object for @replayio/playwright/reporter configuration but received: ${config}`
+      );
+    }
+
+    this.config = {
+      ...config,
+      apiKey: config.apiKey || process.env.REPLAY_API_KEY || process.env.RECORD_REPLAY_API_KEY,
+    };
+    if (!this.config.apiKey) {
+      throw new Error(
+        "`@replayio/playwright/reporter` requires an API key. Either pass a value to the apiKey plugin configuration or set the REPLAY_API_KEY environment variable"
+      );
+    }
+    this.captureTestFile =
+      "captureTestFile" in config
+        ? !!config.captureTestFile
+        : ["1", "true"].includes(
+            process.env.PLAYWRIGHT_REPLAY_CAPTURE_TEST_FILE?.toLowerCase() || "true"
+          );
+  }
 
   getTestId(test: TestCase) {
     return test.titlePath().join("-");
-  }
-
-  parseConfig(config: FullConfig) {
-    let cfg: ReplayPlaywrightConfig = {};
-    config.reporter.forEach(r => {
-      // the reporter is imported from the root reporter.js which imports this
-      // file so we compare the base directory to see if this is our config
-      if (r[0].startsWith(path.resolve(__dirname, ".."))) {
-        if (r[1]) {
-          if (typeof r[1] === "object") {
-            cfg = r[1];
-          } else {
-            console.warn(
-              "Expected an object for @replayio/playwright/reporter configuration but received",
-              typeof r[1]
-            );
-          }
-        }
-      }
-    });
-
-    return cfg;
   }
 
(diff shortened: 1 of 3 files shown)


packages/playwright/src/reporter.ts
```diff
@@ -78,9 +78,35 @@ export interface ReplayPlaywrightConfig
 
 class ReplayPlaywrightReporter implements Reporter {
   reporter?: ReplayReporter<ReplayPlaywrightRecordingMetadata>;
-  captureTestFile = ["1", "true"].includes(
-    process.env.PLAYWRIGHT_REPLAY_CAPTURE_TEST_FILE?.toLowerCase() || "true"
-  );
+  captureTestFile: boolean;
+  config: ReplayPlaywrightConfig;
+
+  constructor(config: ReplayPlaywrightConfig) {
```
Andarist, line 84: Playwright introduced Playwright Test (and thus reporters) in 1.12.0 and those options were passed to constructors there right from the beginning:
https://github.com/microsoft/playwright/blob/3b669e11eb6ccc05f8987cae71213577394e760e/src/test/runner.ts#L68-L69

I'm not sure why we were getting this in `onBegin` using `parseConfig`.

packages/test-utils/src/reporter.ts
```diff
@@ -277,8 +277,8 @@ class ReplayReporter<TRecordingMetadata extends UnstructuredMetadata = Unstructu
       process.env.RECORD_REPLAY_METADATA_TEST_RUN_TITLE ||
       config.runTitle;
 
-    this.apiKey = process.env.REPLAY_API_KEY || process.env.RECORD_REPLAY_API_KEY || config.apiKey;
-    this.upload = !!process.env.REPLAY_UPLOAD || !!config.upload;
+    this.apiKey = config.apiKey || process.env.REPLAY_API_KEY || process.env.RECORD_REPLAY_API_KEY;
+    this.upload = "upload" in config ? !!config.upload : !!process.env.REPLAY_UPLOAD;
```
Andarist, line 281: I think it always makes sense to prefer `config` options over env variables as they are more specific. This is a rule of thumb that I have always been using and it has never failed me so far ;p

packages/test-utils/src/reporter.ts
```diff
@@ -277,8 +277,8 @@ class ReplayReporter<TRecordingMetadata extends UnstructuredMetadata = Unstructu
       process.env.RECORD_REPLAY_METADATA_TEST_RUN_TITLE ||
       config.runTitle;
 
-    this.apiKey = process.env.REPLAY_API_KEY || process.env.RECORD_REPLAY_API_KEY || config.apiKey;
-    this.upload = !!process.env.REPLAY_UPLOAD || !!config.upload;
+    this.apiKey = config.apiKey || process.env.REPLAY_API_KEY || process.env.RECORD_REPLAY_API_KEY;
```
Andarist, line 280: maybe this should use `'apiKey' in config` too?  🤔

packages/playwright/src/reporter.ts
```diff
@@ -78,34 +78,35 @@ export interface ReplayPlaywrightConfig
...
 
-  getTestId(test: TestCase) {
-    return test.titlePath().join("-");
+    this.config = {
+      ...config,
+      apiKey: config.apiKey || process.env.REPLAY_API_KEY || process.env.RECORD_REPLAY_API_KEY,
+    };
+    if (!this.config.apiKey) {
+      throw new Error(
+        "`@replayio/playwright/reporter` requires an API key. Either pass a value to the apiKey plugin configuration or set the REPLAY_API_KEY environment variable"
```
hbenl, line 97: `config.apiKey` is only required if `config.upload` is `true`.

packages/playwright/src/reporter.ts
```diff
@@ -78,34 +78,35 @@ export interface ReplayPlaywrightConfig
...
+      throw new Error(
+        `Expected an object for @replayio/playwright/reporter configuration but received: ${config}`
+      );
+    }
 
-  getTestId(test: TestCase) {
-    return test.titlePath().join("-");
+    this.config = {
+      ...config,
+      apiKey: config.apiKey || process.env.REPLAY_API_KEY || process.env.RECORD_REPLAY_API_KEY,
```
hbenl, line 93: I don't like how this is duplicating the config parsing logic.

packages/playwright/src/reporter.ts
Andarist: That crossed my mind but then I noticed that we also need `apiKey` for various graphql calls. I see now that they are guarded and we don't perform them when there is no API key.

How does it work when those calls are not performed? What the user might be missing in such a case? When it comes to the metadata, I see that it's stored in our `recordings.log` - so at least that is not completely lost.

hbenl: I don't know how much sense it makes to run the tests with our reporter but without an API key. But since we've been offering that option until now I guess we should keep supporting it.

Andarist:
@hbenl i addressed both concerns, could you take another look? The code isn't the prettiest at times but it's structured this way to avoid breaking changes between different versions of cypress/playwright packages and test-utils

hbenl (approved):

PR #438, opened 2024-05-06
Make sure that the fixture doesn't interfere a test run without our reporter

Andarist:


packages/playwright/src/fixture.ts
```diff
@@ -150,15 +154,24 @@ export async function replayFixture(
...
+    await new Promise<void>((resolve, reject) => {
+      ws.on("open", () => resolve());
+      ws.on("error", error => reject(error));
+    });
+  } catch (error) {
+    if (isErrorWithCode(error, "ECONNREFUSED")) {
+      // the reporter didn't end up being used and thus the server is not running
+      await use();
+      return;
+    }
```
bvaughn, line 173: Should we be re-throwing in an `else` here?

bvaughn (approved):

PR #439, opened 2024-05-06
Show reason for unusable recording

bvaughn:
Resolves PRO-533

When a recording is marked as _unusable_, the CLI should report that to the user.

| Scenario | Screenshot |
| :--- | :--- |
| 1 unusable recording | Screenshot 2024-06-05 at 3 48 29 PM |
| 1 unusable recording and 1 usable recording | Screenshot 2024-06-05 at 3 48 37 PM |
| 1 usable recording | Screenshot 2024-06-05 at 3 48 45 PM |
| multiple usable recordings | Screenshot 2024-06-05 at 4 02 48 PM |
| no new recordings | Screenshot 2024-06-05 at 4 04 46 PM |

### DIFF
--- a/packages/replayio/src/commands/record.ts
+++ b/packages/replayio/src/commands/record.ts
@@ -11,12 +11,13 @@ import { exitProcess } from "../utils/exitProcess";
 import { killProcess } from "../utils/killProcess";
 import { trackEvent } from "../utils/mixpanel/trackEvent";
 import { canUpload } from "../utils/recordings/canUpload";
+import { getRecordingUnusableReason } from "../utils/recordings/getRecordingUnusableReason";
 import { getRecordings } from "../utils/recordings/getRecordings";
 import { printRecordings } from "../utils/recordings/printRecordings";
 import { selectRecordings } from "../utils/recordings/selectRecordings";
 import { LocalRecording } from "../utils/recordings/types";
 import { uploadRecordings } from "../utils/recordings/upload/uploadRecordings";
-import { dim } from "../utils/theme";
+import { dim, statusFailed } from "../utils/theme";
 
 registerCommand("record", { checkForRuntimeUpdate: true, requireAuthentication: true })
   .argument("[url]", `URL to open (default: "about:blank")`)
@@ -55,39 +56,45 @@ async function record(url: string = "about:blank") {
   } catch (error) {
     if (error instanceof ProcessError) {
       const { errorLogPath, uploaded } = await reportBrowserCrash(error.stderr);
+
       console.log("\nSomething went wrong while recording. Try again.");
-      console.log(dim(`More information can be found in ${errorLogPath}`));
+      console.log(dim(`\nMore information can be found in ${errorLogPath}`));
       if (uploaded) {
         console.log(dim(`The crash was reported to the Replay team`));
       }
+
       await exitProcess(1);
     }
   }
 
-  const recordingsAfter = getRecordings(processGroupId);
+  const crashedRecordings: LocalRecording[] = [];
+  const finishedRecordings: LocalRecording[] = [];
+  const unusableRecordings: LocalRecording[] = [];
 
-  const nextCrashedRecordings: LocalRecording[] = [];
-  const nextRecordings: LocalRecording[] = [];
-
-  recordingsAfter.filter(recording => {
-    if (recording.recordingStatus === "crashed") {
-      if (canUpload(recording)) {
-        nextCrashedRecordings.push(recording);
-      }
-    } else {
-      nextRecordings.push(recording);
+  getRecordings(processGroupId).forEach(recording => {
+    switch (recording.recordingStatus) {
+      case "crashed":
+        if (canUpload(recording)) {
+          crashedRecordings.push(recording);
+        }
+        break;
+      case "unusable":
+        unusableRecordings.push(recording);
+        break;
+      default:
+        finishedRecordings.push(recording);
     }
   });
 
   console.log(""); // Spacing for readability
 
   // First check for any new crashes; these we should upload automatically
-  if (nextCrashedRecordings.length > 0) {
+  if (crashedRecordings.length > 0) {
     console.log(
-      "It looks like something went wrong with this recording. Please hold while we upload crash data."
+      "It looks like something went wrong while recording. Please hold while we upload crash data."
     );
 
-    const promise = uploadRecordings(nextCrashedRecordings, {
+    const promise = uploadRecordings(crashedRecordings, {
       processingBehavior: "do-not-process",
       silent: true,
     });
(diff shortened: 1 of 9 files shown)


packages/replayio/src/utils/recordings/getRecordingUnusableReason.ts
```diff
@@ -0,0 +1,20 @@
...
+  });
+
+  if (newCrashedRecordings) {
+    // Look for the most recent crashed recording; that is most likely to be related
+    for (let index = newCrashedRecordings.length - 1; index >= 0; index--) {
+      const recording = newCrashedRecordings[index];
+      if (recording.unusableReason) {
+        return recording.unusableReason;
+      }
+    }
```
Andarist, line 18: nit:
```suggestion
    return newCrashedRecordings.findLast(recording => recording.unusableReason)?.unusableReason;
```

Andarist:
I like the I like the premise of this PR - some information is better than no information. It can both point the user and us in the right direction at times

packages/replayio/src/utils/recordings/getRecordingUnusableReason.ts
Andarist: In a sense, we don't target ES2021 here but node@18 and `findLast` is there. According to https://node.green/ there is nothing substantial in node@18 that would be missing from ES2022 and ES2023 - so we could bump the compilation target to ES2023.

I was changing that compilation target already in https://github.com/replayio/replay-cli/pull/369 and I think I just picked the tsconfig values by dates and I didn't consult the compatibility table to see if I could bump it any higher.

I'm fine either way though

bvaughn: https://github.com/replayio/replay-cli/pull/441

bvaughn: Can't get to this with the ES2022 bump but oh well :)

Andarist: It works on my machine now 
findLast being successfully resolved by VS Code

And it only works in `packages/replayio` (which is what we want). Other packages don't see this.

1. Could you recheck what TS version you use in your IDE? I always open command palette in projects and "Select TypeScript version..." to use the one provided by the workspace (otherwise VS Code sticks to its builtin version)
2. Does `yarn boostrap` in the root work OK with this change?

bvaughn: My editor wasn't the one complaining, it was `yarn build` – but now it seems to be working 🤷🏼  Maybe I forgot to `yarn install` again after switching branches.

Andarist (approved):

packages/replayio/src/utils/recordings/getRecordings.ts
```diff
@@ -265,7 +264,7 @@ export function getRecordings(processGroupIdFilter?: string): LocalRecording[] {
     }
   }
 
-  debug("Found %s recordings:\n%o", recordings.length, recordings);
+  debug("Found %s recordings:\n%o\n%o", recordings.length, JSON.stringify(recordings, null, 2));
```
Andarist, line 267: You have too many formatters here (3 formatters but only 2 values) and `%o` is used while you stringify that object urself. You could use multi-line object formatter like this:
```suggestion
  debug("Found %s recordings:\n%O", recordings.length, recordings);
```

packages/replayio/src/utils/recordings/getRecordings.ts
```diff
@@ -274,11 +273,20 @@ export function getRecordings(processGroupIdFilter?: string): LocalRecording[] {
           return false;
         }
 
-        if (!recording.metadata.host) {
-          // Ignore new/empty tab recordings (see TT-1036)
-          // Note that we filter all "empty" recordings, not just root recordings,
-          // because Chrome loads its default new tab content via an <iframe>
-          return false;
+        switch (recording.recordingStatus) {
```
Andarist, line 276: nit: this looks a little bit verbose right now since this whole switch has only one case and its body has an additional condition. It would make sense to combine this in a single if statement (or even into `return condition`)

packages/replayio/src/utils/recordings/getRecordingUnusableReason.ts
```diff
@@ -0,0 +1,8 @@
+import { getRecordings } from "./getRecordings";
+
+export function getRecordingUnusableReason(processGroupIdFilter?: string) {
```
Andarist, line 3: do we want to keep this parameter optional? it's currently supplied at all times, right?

packages/replayio/src/commands/record.ts
```diff
@@ -102,11 +109,18 @@ async function record(url: string = "about:blank") {
     }
 
     console.log(""); // Spacing for readability
+  } else if (unusableRecordings.length > 0) {
+    // If there were unusable recordings we should provide explicit messaging about them
+    const reason = getRecordingUnusableReason(processGroupId);
```
Andarist, line 114: q: why can't we just grab `unusableRecordings.map(r => r.unusableReason)`?

Andarist (approved):

packages/replayio/src/utils/recordings/getRecordings.ts
bvaughn: Agreed it's verbose. When I write code that does something based on an enum, I lean towards `switch` even if it's only a single case because the pattern stands out more clearly to me (but I know this is just subjective).

packages/replayio/src/utils/recordings/getRecordingUnusableReason.ts
bvaughn: Yes, because the `replayio list` command grabs all recordings. It's only the `replayio record` command that filters by the uid.

packages/replayio/src/commands/record.ts
bvaughn: That's fair I guess. When I first wrote this method I thought I might have to read logs or do something more involved. I'll inline it.

packages/replayio/src/commands/record.ts
```diff
@@ -102,11 +108,20 @@ async function record(url: string = "about:blank") {
     }
 
     console.log(""); // Spacing for readability
+  } else if (unusableRecordings.length > 0) {
+    // If there were unusable recordings we should provide explicit messaging about them
+    const reason = unusableRecordings.findLast(
+      recording => recording.recordingStatus === "unusable" && recording.unusableReason
+    )?.unusableReason;
```
Andarist, line 115: We are already working on the filtered list of unusable recordings so rechecking that `recordingStatus` is redundant.

```suggestion
    const reason = unusableRecordings.findLast(recording => !!recording.unusableReason)?.unusableReason;
```

PR #442, opened 2024-05-06
Don't try to group related recordings; just sort reverse chronologically

bvaughn:
Trying to use metadata to group "related" recordings is *flawed* and may result in more confusion for the end user (see the discussion thread on TT-1051). This commit removes that logic. I think the right thing to do, for now, is to remove that workaround and wait until the runtime supports it natively.

Note that CLI output isn't as nice as before in the multi-recording "happy path" scenario:

Screenshot 2024-05-06 at 2 18 04 PM

Screenshot 2024-05-06 at 2 18 16 PM

cc @jonbell-lot23, @jasonLaster We can talk about this during one of our upcoming 1:1s if you'd like


packages/replayio/src/commands/record.ts
```diff
@@ -144,10 +140,8 @@ async function record(url: string = "about:blank") {
 
         console.log(""); // Spacing for readability
       } else {
-        const defaultRecording = findMostRecentPrimaryRecording(nextRecordings);
-
         selectedRecordings = await selectRecordings(nextRecordings, {
-          defaultSelected: recording => recording === defaultRecording,
+          defaultSelected: recording => recording.metadata.processType === "root",
```
bvaughn, line 144: Default select all "root" recordings seems like the right thing to do anyway

jasonLaster: That's fine.

jasonLaster:
Showing all the iframe processes is potentially super noisy now. 

I hope we can find a way to hide non-root frames, but improve the UI so that it's easier to toggle them. The CLI that was best at this was Jest which had that bottom panel for watch commands like `Press p to filter tests`. Not saying we do that per se, simply adding another option like `f` to hide show all frames would work too, but at some point that one line description becomes difficult to read too.


image

bvaughn:
It's noisier but I don't think I've seen a better proposal that wouldn't require addition feature work or UX exploration (nudging us closer toward the Interactive Mode milestone).

bvaughn:
It's noisier but I don't think I've seen a better proposal that wouldn't require addition feature work or UX exploration (nudging us closer toward the Interactive Mode milestone).

We can't just show _only roots_ because the other types can be important too. It's also kind of an open question if people will really understand what "root" recordings are to be honest.

If you'd like to chat about this feature, I'm down to do it over video chat, but I think longer UX discussions on GitHub might be less productive.

jasonLaster:
That's fair. Feel free to move forward. No need to block on this UX discussion.

jonbell-lot23:
Hi @bvaughn ! Just read up on this and I totally agree with the "let's move this out of PR discussion" sentiment.

My proposal is to maybe file a DES ticket for further discussion/exploration? That lets us keep talking but without it getting lost in GitHub or Discord. Open to other approaches too!

packages/replayio/src/commands/record.ts
```diff
@@ -10,14 +10,12 @@ import { exitProcess } from "../utils/exitProcess";
 import { getReplayPath } from "../utils/getReplayPath";
 import { trackEvent } from "../utils/mixpanel/trackEvent";
 import { canUpload } from "../utils/recordings/canUpload";
-import { findMostRecentPrimaryRecording } from "../utils/recordings/findMostRecentPrimaryRecording";
```
Andarist, line 13: if we don't end up using this we should remove the associated file too

packages/replayio/src/commands/record.ts
```diff
@@ -56,20 +54,18 @@ async function record(url: string = "about:blank") {
     }
   }
 
-  const recordingsAfter = await getRecordings();
+  const recordingsAfter = await getRecordings(processGroupId);
```
Andarist, line 57: oooh, i thought before those groups were generated by Chromium and not by us. So this is more like a recording session id or smth like that. It makes sense that we want to get rid of this for now 👍

Andarist (approved):

packages/replayio/src/commands/record.ts
bvaughn: Yup, yup.

PR #443, opened 2024-05-06
Fix test metrics tracking of recorded tests

ryanjduffy:
We were mistakenly always passing `true` for `recorded` where we should have been passing `false` when we failed to find any recordings for a given test.


packages/test-utils/src/reporter.ts
```diff
@@ -777,7 +777,7 @@ class ReplayReporter<TRecordingMetadata extends UnstructuredMetadata = Unstructu
           id: testRun.source.path + "#" + testRun.source.title,
           source: testRun.source,
           approximateDuration: testRun.approximateDuration,
-          recorded: true,
+          recorded: firstRecording != null,
```
Andarist, line 780: nit: I don't think we have to tolerate `null`s here as this comes from an element access
```suggestion
          recorded: firstRecording !== undefined,
```

PR #445, opened 2024-05-07
Fixed `isErrorWithCode` check

Andarist:


packages/playwright/src/fixture.ts
```diff
@@ -14,7 +14,7 @@ import {
 import { getServerPort } from "./server";
 
 function isErrorWithCode<T extends string>(error: unknown, code: T): error is { code: T } {
-  return !!error && typeof error === "object" && "code" in error && typeof error.code === code;
+  return !!error && typeof error === "object" && "code" in error && error.code === code;
```
Andarist, line 17: I'm a 🥔

bvaughn (approved):
Guess I'm a 🥔 too for not spotting it lol

PR #451, opened 2024-05-09
Take release architecture in to account when getting latest release

jazzdan:
The releases API now returns an architecture field. We should make sure to respect that. However, for old releases it will be null and that should be OK.

Fixes TT-1146

### DIFF
--- a/packages/replayio/src/utils/installation/getLatestReleases.test.ts
+++ b/packages/replayio/src/utils/installation/getLatestReleases.test.ts
@@ -11,6 +11,7 @@ jest.mock("./config", () => ({
   runtimeMetadata: {
     platform: "linux",
     runtime: "chromium",
+    architecture: "x86_64",
   },
 }));
 
@@ -28,10 +29,11 @@ describe("getLatestRelease", () => {
     expect(mockedFetch).toHaveBeenCalledWith(expectedUrl);
   });
 
-  it("should return the latest release", async () => {
+  it("should return the latest release and handle null architectures", async () => {
     const expectedUrl = `${replayAppHost}/api/releases`;
     const release: Release = {
       platform: "linux",
+      architecture: null,
       runtime: "chromium",
       releaseFile: "mocked-release-file",
       buildFile: "mocked-build-file",
@@ -41,6 +43,7 @@ describe("getLatestRelease", () => {
     };
     const oldRelease: Release = {
       platform: "linux",
+      architecture: null,
       runtime: "chromium",
       releaseFile: "mocked-release-file",
       buildFile: "mocked-build-file",
@@ -51,6 +54,48 @@ describe("getLatestRelease", () => {
     mockedFetch.mockResolvedValueOnce({ json: async () => [release, oldRelease] } as any);
 
     await expect(getLatestRelease()).resolves.toEqual(release);
+
+    expect(mockedFetch).toHaveBeenCalledWith(expectedUrl);
+  });
+
+  it("should return the latest release for the current platform + architecture", async () => {
+    const expectedUrl = `${replayAppHost}/api/releases`;
+    const now = Date.now().toString();
+    const x86Release: Release = {
+      platform: "linux",
+      architecture: "x86_64",
+      runtime: "chromium",
+      releaseFile: "mocked-release-file",
+      buildFile: "mocked-build-file",
+      buildId: "mocked-build-id",
+      version: "mocked-version",
+      time: now,
+    };
+    const armRelease: Release = {
+      platform: "linux",
+      architecture: "arm",
+      runtime: "chromium",
+      releaseFile: "mocked-release-file",
+      buildFile: "mocked-build-file",
+      buildId: "mocked-build-id",
+      version: "mocked-version",
+      time: now,
+    };
+    const oldRelease: Release = {
+      platform: "linux",
+      architecture: "x86_64",
+      runtime: "chromium",
+      releaseFile: "mocked-release-file",
+      buildFile: "mocked-build-file",
+      buildId: "mocked-build-id",
+      version: "mocked-version",
+      time: (Date.now() - 1000).toString(),
+    };
+    mockedFetch.mockResolvedValueOnce({
+      json: async () => [armRelease, x86Release, oldRelease],
+    } as any);
+
+    await expect(getLatestRelease()).resolves.toEqual(x86Release);
 
     expect(mockedFetch).toHaveBeenCalledWith(expectedUrl);
   });
--- a/packages/replayio/src/utils/installation/getLatestReleases.test.ts
+++ b/packages/replayio/src/utils/installation/getLatestReleases.test.ts
@@ -99,4 +99,40 @@ describe("getLatestRelease", () => {
 
     expect(mockedFetch).toHaveBeenCalledWith(expectedUrl);
   });
+
+  it("should handle a mix of null and unknown architectures", async () => {
+    const now = Date.now();
+    const nowString = now.toString();
+    const unknownRelease: Release = {
+      architecture: "unknown",
+      buildFile: "mocked-build-file",
+      buildId: "mocked-build-id",
+      platform: "linux",
+      releaseFile: "mocked-release-file",
+      runtime: "chromium",
+      time: (now - 500).toString(),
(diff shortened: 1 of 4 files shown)


packages/replayio/src/utils/installation/getLatestReleases.ts
```diff
@@ -5,24 +5,24 @@ import { runtimeMetadata } from "./config";
...
+    release =>
+      release.platform === platform &&
+      release.runtime === runtime &&
+      (release.architecture === architecture || release.architecture === null)
+  );
 
-  assert(latestRelease, `No release found for ${platform}:${runtime}`);
+  debug("Latest release", latestRelease);
+  if (!latestRelease) {
+    throw new Error(`No release found for ${platform}:${runtime}`);
```
jazzdan, line 24: This seemed easier than dealing with `assert` in a unit test, but open to other ideas.

packages/replayio/src/utils/installation/getLatestReleases.ts
```diff
@@ -5,24 +5,24 @@ import { runtimeMetadata } from "./config";
...
+  }
```
Andarist, line 25: `assert` still throws an error - so [this test case](https://github.com/replayio/replay-cli/pull/451/files#diff-ee1a1757b4a83febf95bb5c25c7780d53a29311a86aebde889dde4e5195af90aR23) passes even with an `assert`:
```suggestion
  assert(latestRelease, `No release found for ${platform}:${runtime}`);
```

packages/replayio/src/utils/installation/getLatestReleases.ts
```diff
@@ -5,24 +5,24 @@ import { runtimeMetadata } from "./config";
...
-        return true;
-      }
-    }
-  });
+  const latestRelease = json.find(
+    release =>
+      release.platform === platform &&
+      release.runtime === runtime &&
+      (release.architecture === architecture || release.architecture === null)
+  );
```
Andarist, line 20: could we say that it will be possible to remove `|| release.architecture === null` from here once the new releases with the architecture field get published?

Andarist (approved):

packages/replayio/src/utils/installation/getLatestReleases.ts
bvaughn: Yes. The CLI only checks for the _latest_ release. So once the new releases go out then we could remove this.

packages/replayio/src/utils/installation/getLatestReleases.ts
```diff
@@ -5,24 +5,24 @@ import { runtimeMetadata } from "./config";
...
-      } else {
-        return true;
-      }
-    }
-  });
+  const latestRelease = json.find(
+    release =>
+      release.platform === platform &&
+      release.runtime === runtime &&
+      (release.architecture === architecture || release.architecture === null)
```
bvaughn, line 19: That being said, this change would temporarily break Mac clients, no?

Right now the only release for 20240508 is an x86 release and that release has `"architecture": null`

I think this change would make a Mac with silicon incorrectly match that release.

packages/replayio/src/utils/installation/getLatestReleases.ts
```diff
@@ -5,24 +5,24 @@ import { runtimeMetadata } from "./config";
 import { debug } from "./debug";
 import { Release } from "./types";
 
-const { platform, runtime } = runtimeMetadata;
+const { platform, runtime, architecture } = runtimeMetadata;
```
bvaughn, line 8: nit: Alpha sort props (everywhere)

packages/replayio/src/utils/installation/getLatestReleases.test.ts
```diff
@@ -0,0 +1,102 @@
...
+    };
+    mockedFetch.mockResolvedValueOnce({
+      json: async () => [armRelease, x86Release, oldRelease],
+    } as any);
+
+    await expect(getLatestRelease()).resolves.toEqual(x86Release);
+
+    expect(mockedFetch).toHaveBeenCalledWith(expectedUrl);
+  });
+});
```
bvaughn, line 102: Nice tests!

packages/replayio/src/utils/installation/getLatestReleases.ts
```diff
@@ -16,13 +16,12 @@ export async function getLatestRelease() {
     release =>
       release.platform === platform &&
       release.runtime === runtime &&
+      // TODO(dmiller): remove this check once we have done a release with the architecture field
```
Andarist, line 19: nit: 
```suggestion
      // TODO(dmiller): remove this null check once we have done a release with the architecture field
```

bvaughn (changes requested):
I think this comment needs to be addressed at least before we should land this:
https://github.com/replayio/replay-cli/pull/451/files#r1596702960

packages/replayio/src/utils/installation/getLatestReleases.ts
jazzdan: Agreed, but we did just do a release that now includes entries for both arm and x86 on mac:

```json
[
  {
    "buildId": "linux-chromium-20240510-6bb238250bf2-99bab3984009",
    "runtime": "chromium",
    "version": "108.0.0",
    "platform": "linux",
    "architecture": "x86_64",
    "time": "2024-05-10T21:50:51.972Z",
    "buildFile": "linux-chromium-20240510-6bb238250bf2-99bab3984009.tar.xz",
    "releaseFile": "linux-replay-chromium.tar.xz"
  },
  {
    "buildId": "macOS-chromium-20240510-6bb238250bf2-99bab3984009",
    "runtime": "chromium",
    "version": "108.0.0",
    "platform": "macOS",
    "architecture": "x86_64",
    "time": "2024-05-10T21:47:49.956Z",
    "buildFile": "macOS-chromium-20240510-6bb238250bf2-99bab3984009.dmg",
    "releaseFile": "Replay-Chromium.dmg"
  },
  {
    "buildId": "macOS-chromium-20240510-6bb238250bf2-99bab3984009",
    "runtime": "chromium",
    "version": "108.0.0",
    "platform": "macOS",
    "architecture": "arm",
    "time": "2024-05-10T21:47:19.557Z",
    "buildFile": "macOS-chromium-20240510-6bb238250bf2-99bab3984009-arm.dmg",
    "releaseFile": "Replay-Chromium-arm.dmg"
  },
  // ...
]
```

I think that resolves this issue?

bvaughn: Right! At the time I left this comment that wasn't the case, and our release process is somewhat broken (particularly with respect to Mac) where it wasn't clear that the next release would address both arm and x86 platforms, so this could have broken things.

Looking at the current releases (from yesterday) we have x86 for MacOS and Linux and arm for Mac.

So yeah, I think this should be okay now. I think we can address your TODO comment and remove the `null` case now too since it looks like the API now sends "unknown" for the older releases?

packages/replayio/src/utils/installation/getLatestReleases.ts
```diff
@@ -5,23 +5,24 @@ import { runtimeMetadata } from "./config";
...
-    }
-  });
+  const latestRelease = json.find(
+    release =>
+      release.platform === platform &&
+      release.runtime === runtime &&
+      // TODO(dmiller): remove this null check once we have done a release with the architecture field
+      (release.architecture === architecture ||
+        release.architecture === "unknown" ||
+        release.architecture === null)
```
bvaughn, line 22: I think we can remove this now if you'd like.

packages/replayio/src/utils/installation/getLatestReleases.ts
```diff
@@ -5,23 +5,24 @@ import { runtimeMetadata } from "./config";
...
+  );
 
+  debug("Latest release", latestRelease);
   assert(latestRelease, `No release found for ${platform}:${runtime}`);
 
   debug("Latest release on", new Date(latestRelease.time).toLocaleDateString());
```
bvaughn, line 28: nit: This debug message is a subset of the new one above so we can remove it.
```suggestion
```

bvaughn (approved):

bvaughn (approved):

PR #455, opened 2024-05-10
Annotate tests with related recordings

Andarist:
Those annotations are automatically rendered by the Playwright's builtin junit reporter:
https://github.com/microsoft/playwright/blob/00bf7ea815684cc429824e01749cfa7f20be7f48/packages/playwright/src/reporters/junit.ts#L161-L170

### DIFF
--- a/packages/playwright/src/reporter.ts
+++ b/packages/playwright/src/reporter.ts
@@ -289,23 +289,37 @@ class ReplayPlaywrightReporter implements Reporter {
       };
     }
 
+    const tests = [
+      {
+        ...testMetadata,
+        approximateDuration: test.results.reduce((acc, r) => acc + r.duration, 0),
+        result: status === "interrupted" ? ("unknown" as const) : status,
+        error: result.error
+          ? {
+              name: "Error",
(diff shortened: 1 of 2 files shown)


Andarist:
Hm, ye - I noticed that the Cypress thing was *also* putting that in the `<failure>` elements but I didn't know why that was useful. Since Playwright has this neat way of adding annotations, I think that it's just not worth meddling with the XML manually - even if that's not going to pop up in certain places as you have mentioned (unless this is like a critical feature). Likely there are other ways to add this kind of information to CI annotations.

ryanjduffy:
Adding them to the `failure` element was critical to the customer that requested this so I think some more investigation is warranted. Glad you figured out this part though and it's much simpler than the cypress implementation!

Andarist:
There is a chance that we could also attach text attachments with a body (kinda duplicating those properties here...) and that they would be printed by `<failure>` nodes here: 
https://github.com/microsoft/playwright/blob/e728e90944174fa8d8f7016a9b20d324543dac2e/packages/playwright/src/reporters/base.ts#L307-L308

PR #456, opened 2024-05-13
Add playwright plugin integration test

hbenl:
The test runs the existing playwright example, uploads the recording of the test and checks that the recording contains test metadata with at least one step and at least one playwright annotation.


package.json
```diff
@@ -5,16 +5,17 @@
   "workspaces": [
     "packages/*",
     "packages/*/example",
-    "scripts/*"
+    "scripts/*",
+    "examples/*"
```
hbenl, line 9: I added the example to the workspaces so that it uses the locally built `@replayio/playwright` and `@replayio/replay` packages by default.

package.json
```diff
@@ -5,16 +5,17 @@
...
   ],
   "directories": {
     "example": "examples"
   },
   "scripts": {
-    "bootstrap": "yarn workspaces foreach --all --exclude=. --exclude=packages/cypress/example --topological run build",
+    "bootstrap": "yarn workspaces foreach --all --exclude=. --exclude=examples/create-react-app-typescript --exclude=packages/cypress/example --exclude=packages/node --topological run build",
```
hbenl, line 15: I got confused by the `build` script in `packages/node/package.json` because it only prints "Error: no build specified", so I removed it here.

examples/create-react-app-typescript/test/upload-and-check-recording.js
```diff
@@ -0,0 +1,66 @@
+const assert = require("assert");
+const { WebSocket } = require("ws");
+const { SimpleProtocolClient } = require("@replayio/protocol");
+const replay = require("@replayio/replay");
+
+(async () => {
+  try {
+    const apiKey = process.env.REPLAY_API_KEY;
```
bvaughn, line 8: nit: Worth `assert`ing there's an API key here and failing early?

Otherwise this value gets passed through a few layers of abstraction that try to read cached tokens or other environment variables and eventually throws, but it would be clearer in this case to throw here (since we know this is the only expected entrypoint for an api key)

examples/create-react-app-typescript/test/upload-and-check-recording.js
```diff
@@ -0,0 +1,66 @@
...
+    const recordings = replay.listAllRecordings();
+    assert(recordings.length === 1, `Expected 1 recording but found ${recordings.length}`);
+    const recordingId = recordings[0].id;
+    console.log(`Uploading recording ${recordingId}`);
+    await replay.uploadRecording(recordingId, { apiKey });
+    console.log("Checking metadata");
+    const metadata = await getTestMetadata(recordingId, apiKey);
+    assert(metadata.test.tests[0].events.main.length > 0, "No test events found in metadata");
```
bvaughn, line 16: Should we be checking this too?
```suggestion
    assert(
      metadata.test.tests.length > 0 &&
      metadata.test.tests[0].events.main.length > 0,
       "No test events found in metadata"
     );
```

I guess it would fail either way but with a less meaningful error message.

bvaughn (approved):

PR #461, opened 2024-05-14
Remove assets of test recordings after all uploads complete

Andarist:


packages/test-utils/src/reporter.ts
```diff
@@ -937,6 +945,11 @@ class ReplayReporter<TRecordingMetadata extends UnstructuredMetadata = Unstructu
     }
 
     if (uploads.length > 0) {
+      const recordingIds = uploads.map(u => u.recordingId).filter(isNonNullable);
```
bvaughn, line 948: RE `isNonNullable `: No complaint really but it's not obvious to me why this wouldn't be fine?
```suggestion
const recordingIds = uploads.map(u => u.recordingId).filter(value => value != null);
```

TBH why not just
```suggestion
const recordingIds = uploads.map(u => u.recordingId).filter(Boolean);
```

bvaughn (approved):

packages/test-utils/src/reporter.ts
Andarist: The answer to both: TypeScript 😅 I want `string[]` back but at the moment TS won't understand that I'm doing any kind of narrowing here (this is going to change with [inferred type predicates](https://devblogs.microsoft.com/typescript/announcing-typescript-5-5-beta/#inferred-type-predicates) in 5.5, this has to be one of the best improvements in TS for years and I love it :p).

`Boolean` won't work as a type predicate though because type predicates have to be exhaustive. Type predicates are used for narrowing and that can also help with narrowing "negative branches". But we can't go from `string | null` to `string` in the true branch and `null` in the false branch by using `Boolean`. `Boolean` coerces falsy primitives like `''`, 0, etc - and for that reason it won't work as the type predicate. If it did we'd assume `null` in the false branch but in reality it could still be an empty string.

The inline version that you have proposed will work in TS 5.5 though 🚀

PR #462, opened 2024-05-14
Mark the canceled uploads as failed

Andarist:
Just a quick thing to address the weirdness reported as PRO-433

### DIFF
--- a/packages/replayio/src/utils/recordings/getRecordings.ts
+++ b/packages/replayio/src/utils/recordings/getRecordings.ts
@@ -211,7 +211,10 @@ export function getRecordings(processGroupIdFilter?: string): LocalRecording[] {
 
             const recording = idToRecording[id];
             assert(recording, `Recording with ID "${id}" not found`);
-            recording.uploadStatus = "uploading";
+            // if we can only find the "uploadStarted" log then we have to assume that this upload has failed or has been canceled
+            // we can't retry those as "Internal.beginRecordingUpload" has already been called before and it fails on the second attempt
+            // recordings line create a timeline so "uploadFinished" still has a chance to update this status to "uploaded"
+            recording.uploadStatus = "failed";
             break;
           }
           case RECORDING_LOG_KIND.writeFinished: {


packages/replayio/src/utils/recordings/getRecordings.ts
```diff
@@ -211,7 +211,10 @@ export function getRecordings(processGroupIdFilter?: string): LocalRecording[] {
 
             const recording = idToRecording[id];
             assert(recording, `Recording with ID "${id}" not found`);
-            recording.uploadStatus = "uploading";
+            // if we can only find the "uploadStarted" log then we have to assume that this upload has failed or has been canceled
+            // we can't retry those as "Internal.beginRecordingUpload" has already been called before and it fails on the second attempt
+            // recordings line create a timeline so "uploadFinished" still has a chance to update this status to "uploaded"
+            recording.uploadStatus = "failed";
```
bvaughn, line 217: This breaks in-progress monitoring.

bvaughn (changes requested):

packages/replayio/src/utils/recordings/getRecordings.ts
Andarist: Could you describe the exact scenario so I could test it out later? Based on that my canceled uploads are just marked as failed at the start and I looked through all of the calls to `getRecordings` - this is only used to load up the state at the start of the CLI (and sometimes after we are done with our actions, so we can compare the one that we had before the action and the one that we have after it).

I just ran through a happy-path upload flow and I didn't notice any weirdness

bvaughn: There are a few scenarios I'm considering:
* The CLI prints in-progress messages based on upload status (see `uploadRecordings` -> `printDeferredRecordingActions`)
  * This will also be true for the upcoming interactive mode which shows a list of recordings along with their statuses
* One CLI process could be uploading things while you run `list` in another

My main concern here is that we shouldn't show _failed_ for recording that is actively being uploaded.

bvaughn: Maybe the in-progress message for the current CLI wouldn't break because the list is already in-memory, but the upcoming interactive mode would (and the multi-CLI scenario would)

Andarist: > The CLI prints in-progress messages based on upload status (see uploadRecordings -> printDeferredRecordingActions)

Yeah, but in here we are only setting the *initial* value of the `uploadStatus` - so it shouldn't interfere with the ongoing uploads etc

> One CLI process could be uploading things while you run list in another

This is a fair concern and it crossed my mind. I think it's an edge case though and that we don't handle it too well today *anyway*. 

> Maybe the in-progress message for the current CLI wouldn't break because the list is already in-memory,

This ☝️ 

> but the upcoming interactive mode

We could revisit it then - atm I just don't know how it will work so it's hard to predict how exactly it would impact this scenario and how potentially we could mitigate it in the future when we get to it.

bvaughn: TBH I'd really rather us not have data in a broken state to gloss over an unknown upload issue with the backend. Even if it only affects edge case or future cases.

bvaughn: I'd prefer we just leave this behavior as-is for now until we prioritize working on the CLI again.

PR #463, opened 2024-05-14
Avoid redundant logs when annotations fail to be added

Andarist:


packages/playwright/src/reporter.ts
```diff
@@ -279,7 +278,7 @@ class ReplayPlaywrightReporter implements Reporter {
               try {
                 return [filename, readFileSync(filename, "utf8")];
               } catch (e) {
-                warn(`Failed to read playwright test source for: ${filename}`, e);
+                debug(`Failed to read playwright test source for: ${filename}`, e);
```
hbenl, line 281: This surprised me: do we expect this to happen often enough that we want to hide those warnings by default? I understand that we want to hide the annotation warnings by default because that happens so often but I've never seen this warning.

hbenl (approved):

packages/playwright/src/reporter.ts
Andarist: This shouldn't happen often. However, we get those filenames from the call stacks - I'm not quite sure how reliable this is (especially considering source mapped sources etc). 

It could be good to raise those as errors somewhere... but perhaps this should be done by aggregating those errors and surfacing this after the whole run? Outputting this when it happens is pretty random and interferes with whatever reporters u have enabled. WDYT?

hbenl: OK, I wasn't aware that it can interfere with other reporters - in that case it makes sense to hide those messages by default.
We could raise these errors somewhere else but if we don't have a mechanism for doing so already I guess that would be a separate issue.

hbenl:
Somehow all the test steps in the recording were grouped under `beforeEach`...

Andarist:
Even before this PR steps are grouped in a wrong way.

<details>
<summary>Relevant part of the metadata on the base branch</summary>

```json
[
  {
    "id": 0,
    "error": null,
    "events": {
      "main": [
        {
          "data": {
            "id": "expect@12",
            "error": null,
            "scope": [],
            "command": { "name": "expect", "arguments": [] },
            "category": "assertion",
            "parentId": null
          }
        }
      ],
      "afterAll": [],
      "afterEach": [],
      "beforeAll": [],
      "beforeEach": [
        {
          "data": {
            "id": "hook@1",
            "error": null,
            "scope": [],
            "command": { "name": "browser.newContext", "arguments": [] },
            "category": "command",
            "parentId": null
          }
        },
        {
          "data": {
            "id": "hook@1",
            "error": null,
            "scope": [],
            "command": { "name": "browserContext.newPage", "arguments": [] },
            "category": "command",
            "parentId": null
          }
        },
        {
          "data": {
            "id": "hook@1",
            "error": null,
            "scope": [],
            "command": {
              "name": "page.goto",
              "arguments": ["http://localhost:3000/"]
            },
            "category": "command",
            "parentId": null
          }
        }
      ]
    },
    "result": "passed",
    "source": { "scope": [], "title": "basic test" },
    "attempt": 1,
    "approximateDuration": 4819
  }
]
```
</details>

The `page.goto` is grouped in the `beforeEach` but it should be within `main` group. From what I gathered so far this way of obtaining the current step isn't reliable: 
https://github.com/replayio/replay-cli/blob/e3d7e8f8190bbe51cde55815db5421e7f3cab9c0/packages/playwright/src/fixture.ts#L316

PR #469, opened 2024-05-17
Test multiple playwright versions

Andarist:


.github/workflows/playwright-tests.yml
```diff
@@ -1,21 +1,38 @@
...
 
 jobs:
   playwright-tests:
     name: Playwright tests
     runs-on: ubuntu-latest
+    timeout-minutes: 20
+    strategy:
+      fail-fast: false
+      matrix:
+        playwright-version: [1.34.x, 1.38.x, 1.44.x]
```
Andarist, line 15: I feel like this is a good start, we can add more versions here but it feels like an overkill to list *every* minor version between 1.34 and 1.44. On the other hand... none of them guarantee stability for the stuff that we integrate with so maybe we'll have to resort to that?

bvaughn (approved):

PR #472, opened 2024-05-20
Don't require an empty config when creating a reporter

Andarist:


bvaughn (approved):

packages/playwright/src/index.ts
```diff
@@ -53,7 +53,7 @@ export const devices = {
   },
 };
 
-export function createReplayReporterConfig(config: ReplayPlaywrightConfig) {
+export function createReplayReporterConfig(config: ReplayPlaywrightConfig = {}) {
```
bvaughn, line 56: Didn't you rename this fella

Andarist: Yeah, I did - I have to rebase this PR in the morning 😉

PR #473, opened 2024-05-20
Export less verbose `replayReporter` factory

Andarist:


bvaughn (approved):
I'm fine with this but I'm curious what motivated this change? Is it more inline with other plug-in naming conventions or something?

Andarist:
Jason mentioned that the old name was quite verbose - with which I agree. Before opening the PR, I looked through all the 3rd party reporters listed in the Playwright's docs. Most of them don't have anything like that.

It's convenient though because it introduces contextual typing for our options - with it TS is able to verify that the user is using correct options/types/etc. It also automatically types `filter`'s parameter. An alternative, would be to export a config type (which we probably already do) and make people to use `satisfies` to achieve the same/similar result. While that option is cool, I don't think it's that ergonomic. 

I only found a single reporter that has a factory function like this and it uses the same naming scheme for its factory function so I went with it. Based on that single reporter It's hard to call it a convention though :P

bvaughn:
That's fair. I don't disagree with this change. I just wanted more context :)

PR #474, opened 2024-05-20
CLI no longer supports Windows+Chrome

bvaughn:
This behavior can be overridden via the `REPLAY_WINDOWS_CHROMIUM_OVERRIDE` flag.

> [!NOTE]  
> I know, I know. Another env flag. It seemed the best way to allow Replay engineers to work around this like Jason requested. A command flag seems no better as we wouldn't want it to be user-visible.

Open to suggestions on the wording
Screenshot 2024-05-20 at 3 30 34 PM

### DIFF
--- a/packages/replayio/src/utils/installation/config.ts
+++ b/packages/replayio/src/utils/installation/config.ts
@@ -1,4 +1,5 @@
 import { getReplayPath } from "../getReplayPath";
+import { emphasize } from "../theme";
 import { Architecture, Platform, Runtime } from "./types";
 
 type Metadata = {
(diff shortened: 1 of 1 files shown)


packages/replayio/src/utils/installation/config.ts
```diff
@@ -44,16 +45,24 @@ switch (process.platform) {
...
+        path: ["replay-chromium", "chrome.exe"],
+        platform: "windows",
+        runtime: "chromium",
+        sourceName: "replay-chromium",
+      };
+    } else {
+      console.log("");
+      console.log(emphasize("Replay does not support Windows at this time."));
+      console.log("Please use the Windows Subsystem for Linux (WSL) instead.");
+      process.exit(0);
```
Andarist, line 64: I'd probably expect a non-zero code in this situation as effectively this isn't a successful run overall.

Andarist (approved):

packages/replayio/src/utils/installation/config.ts
bvaughn: That's fair. I had been on the fence about that. Happy to change it.

PR #479, opened 2024-05-21
Add `install` as an alias for `update`

Andarist:


packages/replayio/src/commands/update.ts
```diff
@@ -11,6 +11,7 @@ registerCommand("update", {
   checkForNpmUpdate: false,
 })
   .description("Update Replay")
+  .alias("install")
```
Andarist, line 14: This gets now rendered by help. We could hide aliases in `formatOutput` or similar. Or we could implement this using an extra command and add `hidden: true` to it.

I briefly looked through the CLI output produced when running this command and I don't think we need any specific changes to it. The current output is fine even when running `replayio install`.

bvaughn: I don't love the way this looks

bvaughn: I think it would make sense to hide the "install" alias from normal `replayio` usage because it doesn't really make semantic sense there. It exists only for the Playwright plug-in scenario really.

Andarist: I pushed out a change to drop aliases from this output so it should now look the same as before. Could you take another look?

bvaughn (approved):
Looks good

PR #486, opened 2024-05-23
Fixed `getExecutablePath` description in the README

Andarist:


bvaughn (approved):
Thanks for the follow up

PR #487, opened 2024-05-23
Setup Changesets

Andarist:
This PR sets up [Changesets](https://github.com/changesets/changesets). This is a tool helping version monorepos and releasing the contained packages on CI.

The idea is that each PR should come with a special changeset file (just a small markdown file with a frontmatter). That file describes a changelog text, a list of relevant packages, and desired "bump" types (patch/minor/major).

an example changeset file

You can use `yarn changeset` to create those:

a selectable list of potentially affected packages in the CLI

There is no 1 PR 1 file rule. You can create as many changeset files as you want. The same change can be described differently for all affected packages, depending on the abstraction layer etc. Similarly, you can create multiple changeset files in a PR if it introduces multiple (unrelated~) changes. Please do what makes the most sense to the reader - commit message !== changelog entry (although often they can be the same).

Advantages:
- we document things as we go
- we can add/remove/change those files whenever we want, the changeset files are the source of truth
- this becomes part of the code review process

Whenver new changeset files land on `main` a GitHub action creates (or updates) a PR like [this one](https://github.com/changesets/changesets/pull/1358):
an example Version Packages PR

And when we merge this PR the release process gets kicked off and packages are published. 

Advantages:
- this is pretty sweet because installation, fresh builds, and all happen on CI. There is a much smaller surface for related mistakes
- it's easy to review what is meant to be published with each release
- it's very convenient because you don't have to switch between branches to make the release. The release is just a single click of a button away :)


bvaughn (approved):
I like this change a lot. I think the one part of it that I feel a little conflicted over is the lack of manually pre-release testing on the built artifact(s) but I think the benefits of the added process outweigh that 👍🏼

Andarist:
> I think the one part of it that I feel a little conflicted over is the lack of manually pre-release testing on the built artifact(s)

What kind of tests are you thinking about here? We could even create "snapshot" releases using this flow (with `0.0.0-hash-date-whatever-i-dont-recall-the-pattern`). So those artifacts could be manually tested and all. In reality, this probably wouldn't be exercised too often ;p so I'm not sure if there is a huge value in setting this up

PR #494, opened 2024-05-27
Make it possible to enable recorder's verbose logging using `DEBUG` variable

Andarist:


packages/replayio/src/utils/browser/launchBrowser.ts
```diff
@@ -34,7 +33,7 @@ export async function launchBrowser(
       RECORD_ALL_CONTENT: "1",
       RECORD_REPLAY_DIRECTORY: getReplayPath(),
       RECORD_REPLAY_METADATA: JSON.stringify({ processGroupId }),
-      RECORD_REPLAY_VERBOSE: verbose ? "1" : undefined,
+      RECORD_REPLAY_VERBOSE: "1",
```
Andarist, line 36: This is a quick idea, the core of my change is in the playwright's package. It just seems that perhaps the same idea can just be used here - we enable verbose logs at all times **but** we still gate them behind `DEBUG`.

Note that I didn't test it out with the CLI yet. Rn I'm heading out but I wanted to open this to get the initial feedback.

packages/playwright/src/index.ts
```diff
@@ -10,6 +10,8 @@ function getDeviceConfig() {
   const env: Record<string, any> = {
     ...process.env,
     RECORD_ALL_CONTENT: 1,
+    // it doesn't log anything eagerly but it makes it possible to enable verbose logs with DEBUG=pw:browser
+    RECORD_REPLAY_VERBOSE: 1,
```
bvaughn, line 14: What's the added overhead from this added logging

bvaughn (commented):
I don't understand the motivation for this change

packages/playwright/src/index.ts
Andarist: I doubt it's high because there are really not that many logs produced by this verbose output from what I can tell. Writing to stdout shouldn't incur a significant overhead and the data logged there is probably available anyway. Just to be sure, we can ask @Domiii what he thinks about this.

Andarist:
> I don't understand the motivation for this change

The main motivation is to make debugging playwright tests easier. Right now to get those logs you have to make a change to your config like:
```diff
diff --git a/examples/create-react-app-typescript/playwright.config.ts b/examples/create-react-app-typescript/playwright.config.ts
index 2f48cff..dcaccd4 100644
--- a/examples/create-react-app-typescript/playwright.config.ts
+++ b/examples/create-react-app-typescript/playwright.config.ts
@@ -23,6 +23,13 @@ export default defineConfig({
       name: "replay-chromium",
       use: {
         ...replayDevices["Replay Chromium"],
+        launchOptions: {
+          ...replayDevices["Replay Chromium"].launchOptions,
+          env: {
+            ...replayDevices["Replay Chromium"].launchOptions.env,
+            RECORD_REPLAY_VERBOSE: "1",
+          },
+        },
       },
     },
   ],
```

And on top of this, you need to run your tests with `DEBUG=pw:browser`. So it's not straightforward and I'd like for it to work with just `DEBUG=pw:browser`. 

With that proposed change in place, it would make sense for the CLI to behave the same way. To get the most complete Replay-related output I could just use `DEBUG=replayio:* replayio record  http://localhost:3000` instead of `DEBUG=replayio:* replayio record  http://localhost:3000 --verbose`. Having 2 ways of enabling some of those logs feels like a chore to me and it's harder to document and explain to people when we ask them to enable them for troubleshooting purposes.

packages/playwright/src/index.ts
Domiii: I don't have any numbers on log overhead, but I would intuitively agree that this should not be too bad. Having numbers would be better though 😅

Domiii:
> The main motivation is to make debugging playwright tests easier.

Is there no way to add a new `verbose` flag through the playwright plugin that works just like the CLI `--verbose` flag?

Andarist:
> Is there no way to add a new verbose flag through the playwright plugin that works just like the CLI --verbose flag?

It's not our script that runs tests - it's Playwright, we just supply some plugins. I feel like CLI flags belong to the script itself and it would feel out of place to interpret them on our side. They don't support `--verbose` right now but they could in the future and the meaning could be slightly different than what we'd define for ourselves. 

In addition to that, there could always be some level of indirection between the actual Playwright call and the script that is executed (we even do it ourselves in our e2e tests [here](https://github.com/replayio/devtools/blob/e19b3a3112c14556e233a2cd29e8e303844813c8/packages/e2e-tests/scripts/record-playwright.ts), although this one isn't indirect enough for `--verbose` to get lose). Environment variables are commonly passed along so it's a good fit.

I also find it weird-ish that right now with `--verbose` I get the browser logs but that's about it. `DEBUG` gives you more control because you can dump everything or try to be selective with the defined "scopes". 

Note that this PR doesn't remove `--verbose`, it should still work the same way. The key difference is that now `DEBUG=replayio:browser` and `DEBUG=replayio:*` should spit out those logs without having to use `--verbose` too.

bvaughn:
> With that proposed change in place, it would make sense for the CLI to behave the same way. To get the most complete Replay-related output I could just use `DEBUG=replayio:* replayio record http://localhost:3000` instead of `DEBUG=replayio:* replayio record http://localhost:3000 --verbose`. Having 2 ways of enabling some of those logs feels like a chore to me and it's harder to document and explain to people when we ask them to enable them for troubleshooting purposes.

The `--verbose` flag already enabled a specific subgroup of debug logging (`"replayio:browser"`) and "_just append `--verbose` to the command_" seems easier to explain to someone than the debug var IMO.

That being said, I think I agree that if we're hoping for arguments to be inherited by nested things, env vars probably seem preferable.

I think we should probably just remove `--verbose` entirely if we're going in this direction though. (Speaking of avoiding having "_two ways to do something_".) It was never a public flag so it should be fine to remove in a minor.

Andarist:
@Domiii WDYT about removing `--verbose` is favor of `DEBUG=replayio:browser`? I agree with @bvaughn that we don't need multiple ways of doing the same thing here but I also don't want to disturb your workflow too much

Domiii:
> WDYT about removing `--verbose` is favor of `DEBUG=replayio:browser`?

I'd prefer an easier variant in all cases.

I think `DEBUG=replayio:browser`...
1. Is harder to remember.
2. Is _not cross-platform_.
3. Makes customer support a tad more difficult.

But from the sound of it, there are too many other much bigger problems for you guys to attend to, so I don't want to stand in the way!

Andarist:
@Domiii how you'd feel about `--verbose` enabling `DEBUG=replayio:*` (not just the browser's verbose logs)?

bvaughn:
> @Domiii how you'd feel about `--verbose` enabling `DEBUG=replayio:*`

Should we just enable * in that case? (Enabling _only_ "replayio:*" would not include the `pw:browser` group?)

Andarist:
I'm talking about `replayio --verbose` here. `pw:browser` is a DEBUG scope created by Playwright. 

We can't even even try to handle `--verbose` passed to Playwright (like in `playwright test --project replay-chromium --verbose`) since they just error with:
> error: unknown option '--verbose'

bvaughn:
Got it. I was confused about what was logging with `pw:browser` and I thought it was Replay browser for some reason.

Domiii:
> how you'd feel about --verbose enabling DEBUG=replayio:* (not just the browser's verbose logs)? Note that this would only be something we could support in the CLI. In Playwright you'd have to do DEBUG=replayio:*,pw:browser

I'm not sure what the difference is, so I don't know how to answer that. Its probably fine. 😆 

Last Q: Do we have a wrapper around playwright that people can use that uses a cross-platform approach for env var management or somesuch? Or do we have to rely on them just modifying the playwright config correctly? (I'm also asking because it affects our new "opt-in Asserts" for which we need a reliable leaver to toggle, and that is also an env var.)

bvaughn:
> Last Q: Do we have a wrapper around playwright that people can use that uses a cross-platform approach for env var management or somesuch?

No. (We've talked about it and there's some disagreeing about whether it would be a good idea or not, but I think it would be complex.)

> Or do we have to rely on them just modifying the playwright config correctly?

Yes. Likely they will have already had to sort out how to do this.

Andarist:
> I'm not sure what the difference is, so I don't know how to answer that. Its probably fine. 😆

You can check out what that would do by running this today: `DEBUG=replayio:* replayio record  http://localhost:3000 --verbose`. Right now `--verbose` enables one of our "scopes" (the logs coming from the browser) whereas `DEBUG=replayio:*` enables **all of them**. So the difference is that with the proposed change you will get a lot more logs than today, that's it. To get only the ones that you care about granularly you'll be able to use `DEBUG` variable with different values to control what gets spit out to stdout.

> Yes. Likely they will have already had to sort out how to do this.

I think this should make it at least a little bit easier: https://github.com/replayio/replay-cli/pull/501 . Requiring user to change `launchOptions` in their config is too error-prone

bvaughn (approved):
I defer to @Andarist's judgement on this!

PR #496, opened 2024-05-28
Require Replay browser to exit before starting a new recording

bvaughn:
cc @cweld510


.changeset/nice-seahorses-study.md
```diff
@@ -0,0 +1,5 @@
+---
+"replayio": minor
```
bvaughn, line 2: This is debatable I guess.

PR #500, opened 2024-05-28
Close the launched browser on Ctrl+C when recording

Andarist:


packages/replayio/src/utils/prompt/prompt.ts
```diff
@@ -38,6 +40,7 @@ export async function prompt({
         case "\x03":
           // \x03 is Ctrl+C (aka "End of text")
           // https://donsnotes.com/tech/charsets/ascii.html
+          onExit?.();
           process.exit(0);
```
bvaughn, line 44: It's kind of gross the `prompt` is listening for this to begin with IMO. I wasn't sure of a better way to _not_ break that functionality. (Although I should probably have used the `exitProcess` util here rather than calling `process.exit` directly.)

packages/replayio/src/utils/prompt/prompt.ts
```diff
@@ -38,6 +40,7 @@ export async function prompt({
         case "\x03":
           // \x03 is Ctrl+C (aka "End of text")
           // https://donsnotes.com/tech/charsets/ascii.html
+          onExit?.();
```
bvaughn, line 43: I wonder if we can handle this in a broader way by listening for "SIGINT" ?

I think what you have here _works_ but it's also kind of narrowly scoped.

packages/replayio/src/utils/prompt/prompt.ts
Andarist: > It's kind of gross the prompt is listening for this to begin with IMO.

I don't think we can escape this easily because we are in the raw mode here. One could have some script that could just do anything on CTRL+C in this mode - it loses its default behavior altogether.

> Although I should probably have used the exitProcess util here rather than calling process.exit directly.

I thought about using it here and decided against it as the user expects an immediate~ exit. That function waits for LaunchDarkly to exit etc.

packages/replayio/src/utils/prompt/prompt.ts
Andarist: > I wonder if we can handle this in a broader way by listening for "SIGINT" ?

I wanted to do that at first but in raw mode SIGINT is simply not dispatched here. The process gets closed **only** because we call `process.exit()` on our own from within this handler~.

bvaughn: > I wanted to do that at first but in raw mode SIGINT is simply not dispatched here.

Ah, ok.

packages/replayio/src/utils/prompt/prompt.ts
bvaughn: Okay. I don't feel strongly about what the "right" thing is to do here. I'll defer!

bvaughn (approved):

PR #503, opened 2024-06-03
Upload data related to browser crashes

Andarist:
This PR adds auto-reporting of the crash-related data. This includes:
- stderr of the browser process
- crashpad files (only on Mac)

In the process, I changed the `fetch` implementation that we are using to `undici`. `node-fetch`'s new version was ESM-only and that didn't play well with our CJS modules here (I didn't want to resort to using a dynamic import with it). 

`undici` is used within node itself. That being said, I had to use v5 because v6 imports `File` from node. That in node 18 comes with a warning that I couldn't get rid of:
https://github.com/nodejs/node/blob/64903b1ca04dbe182b0abc36deb7437a49bf1881/lib/internal/file.js#L40


packages/replayio/src/utils/recordings/upload/uploadRecording.ts
```diff
@@ -292,7 +291,6 @@ async function uploadRecordingReadStream(
   try {
     const response = await Promise.race([
       fetch(url, {
-        agent: getKeepAliveAgent,
```
Andarist, line 295: from what I understand, keep alive is already the default in `undici`:
https://github.com/nodejs/undici/blob/985b3816708512bafefa1544def183cc6d1536be/lib/client.js#L273-L274

packages/replayio/src/utils/graphql/fetchUserIdFromGraphQLOrThrow.ts
```diff
@@ -25,9 +25,7 @@ export async function fetchUserIdFromGraphQLOrThrow(accessToken: string) {
           }
         }
           `,
-    {
-      key: accessToken,
-    },
```
Andarist, line 30: `accessToken` is not a variable but it's part of the header, we were passing it in both places here

packages/replayio/src/commands/record.ts
```diff
@@ -176,3 +171,57 @@ async function record(url: string = "about:blank") {
...
+
+  formData.append(
+    "log",
+    new File([await readFile(errorLogPath)], "log.txt", { type: "text/plain" })
+  );
+
+  const latestCrashpad =
+    runtimeMetadata.crashpadDirectory &&
+    (await findMostRecentFile(runtimeMetadata.crashpadDirectory, fileName =>
+      fileName.endsWith(".dmp")
```
Andarist, line 204: this directory really should only have those `.dmp` files, I'm using this predicate just as an extra precaution

packages/replayio/src/utils/installation/config.ts
```diff
@@ -18,8 +21,18 @@ const architecture: Architecture = process.arch.startsWith("arm") ? "arm" : "x86
 
 switch (process.platform) {
   case "darwin":
+    const homeDirectory = process.env.HOME || process.env.USERPROFILE;
+    assert(homeDirectory, "HOME or USERPROFILE environment variable must be set");
```
bvaughn, line 25: Why not just use [`os.homedir()`](https://nodejs.org/api/os.html#oshomedir)?

packages/replayio/src/commands/record.ts
```diff
@@ -176,3 +171,57 @@ async function record(url: string = "about:blank") {
...
+
+  const errorLogPath = getReplayPath("recorder-crash.log");
+  writeFileSync(errorLogPath, stderr, "utf8");
+  console.log(dim(`More information can be found in ${errorLogPath}`));
+
+  const accessToken = await checkAuthentication();
+
+  if (!accessToken) {
+    return;
+  }
```
bvaughn, line 186: Some nits:

* I prefer methods like this to be in their own files (like you did with `utils/findMostRecentFile`) because IMO this makes it easier to navigate project source, but this is just a personal preference.
* I think the first few parts of this method (the console logging and `recorder-crash.log` writing) should be in the caller, rather than this method (given this method's name)

packages/replayio/src/commands/record.ts
```diff
@@ -176,3 +171,57 @@ async function record(url: string = "about:blank") {
...
+        Authorization: `Bearer ${accessToken}`,
+        "User-Agent": userAgent,
+      },
+      body: formData,
+    });
+    if (response.status >= 200 && response.status < 300) {
+      console.log(dim(`The crash was reported to the Replay team`));
+      return;
+    }
+  } catch (err) {}
```
bvaughn, line 226: nit: Should we print the error to the `DEBUG` output?

packages/replayio/src/utils/findMostRecentFile.ts
```diff
@@ -0,0 +1,33 @@
...
+  );
+
+  let mostRecent: (typeof candidates)[number] | undefined = undefined;
+
+  for (const candidate of candidates) {
+    if (!candidate || (mostRecent && candidate.mtimeMs < mostRecent.mtimeMs)) {
+      continue;
+    }
+    mostRecent = candidate;
+  }
```
bvaughn, line 30: tiny nit: Seems like we could do this check in the above `map`-filter loop and avoid having to iterate twice

bvaughn (approved):
Nice!

packages/replayio/src/utils/installation/config.ts
Andarist: It's a copy-pasta from `getReplayPath`. Reading through `os.homedir` docs, it looks like it's a better fit - I'll switch both places to it

packages/replayio/src/commands/record.ts
Andarist: ye, good idea - i added this just now

packages/replayio/src/commands/record.ts
Andarist: applied both suggestions

packages/replayio/src/utils/getReplayPath.ts
```diff
@@ -1,15 +1,12 @@
-import assert from "assert";
+import { homedir } from "os";
```
bvaughn, line 1: Nice!

packages/replayio/src/commands/record.ts
```diff
@@ -62,7 +54,12 @@ async function record(url: string = "about:blank") {
     await launchBrowser(url, { processGroupId });
   } catch (error) {
     if (error instanceof ProcessError) {
-      await reportRecorderCrash(error.stderr);
+      const { errorLogPath, uploaded } = await reportBrowserCrash(error.stderr);
+      console.log("\nSomething went wrong while recording. Try again.");
+      console.log(dim(`More information can be found in ${errorLogPath}`));
+      if (uploaded) {
+        console.log(dim(`The crash was reported to the Replay team`));
+      }
```
bvaughn, line 62: Nice!

bvaughn (approved):

bvaughn:
I think if we're good to merge this, we can publish a release.

PR #505, opened 2024-06-04
Show recording title (metadata.title) if one has been set

bvaughn:
Otherwise fall back to showing the host

### Before
Screenshot 2024-06-04 at 9 13 46 AM

### After
Screenshot 2024-06-04 at 9 12 14 AM

### DIFF
--- a/packages/replayio/src/utils/recordings/getRecordings.ts
+++ b/packages/replayio/src/utils/recordings/getRecordings.ts
@@ -26,7 +26,9 @@ export function getRecordings(processGroupIdFilter?: string): LocalRecording[] {
 
           Object.assign(recording.metadata, metadata);
 
-          const { argv, process, processGroupId, uri } = metadata;
+          const { argv, process, processGroupId, title, uri } = metadata;
+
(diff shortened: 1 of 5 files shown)


packages/replayio/src/utils/recordings/getRecordings.ts
```diff
@@ -26,7 +26,9 @@ export function getRecordings(processGroupIdFilter?: string): LocalRecording[] {
 
           Object.assign(recording.metadata, metadata);
 
-          const { argv, process, processGroupId, uri } = metadata;
+          const { argv, process, processGroupId, title, uri } = metadata;
+
+          recording.metadata.title = title;
```
Andarist, line 31: I'm not sure if that's a valid scenario right now but it doesn't seem unreasonable to have multiplied `addMetadata` entries for a single recording. In such a case, if the title would be in the first one then we'd remove it when processing the second one.
```suggestion
          if (title) {
            recording.metadata.title = title;
          }
```

Andarist (approved):

PR #508, opened 2024-06-05
replayio: add trailing newline to recordings.log

hbenl:
[This comment](https://linear.app/replay/issue/PRO-572/playwright-reporter-sometimes-doesnt-write-test-metadata#comment-2c3f1fea) explains how leaving out the trailing newline creates problems.

### DIFF
--- a/packages/replayio/src/utils/recordings/readRecordingLog.ts
+++ b/packages/replayio/src/utils/recordings/readRecordingLog.ts
@@ -15,6 +15,10 @@ export function readRecordingLog() {
       } catch (err) {
         debug(`Error parsing line:\n${line}`);
 
+        // Early versions of `replayio` could remove the trailing \n from recordings.log,
+        // so the next entry would be appended to the last line, creating a line with two
+        // entries. This workaround lets us read these corrupted entries but it should
+        // be removed eventually.
         const replaced = line.replace(/\}\{/g, "}\n{");
 
         if (replaced.length === line.length) {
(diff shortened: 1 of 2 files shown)


.changeset/witty-laws-join.md
```diff
@@ -0,0 +1,5 @@
+---
+"replayio": patch
+---
+
+Fixed a potential data corruption issue related to the `recrdings.log` updates
```
Andarist, line 5: @hbenl feel free to tweak this if you think this is inaccurate
```suggestion
Fixed a potential data corruption issue related to the `recordings.log` updates
```

oops, i made a type 🤐

Andarist (approved):

Andarist:
I think we should add a TODO comment to [this line](https://github.com/replayio/replay-cli/blob/4954cd14547359936842b2d3c59d6f138eab21ba/packages/replayio/src/utils/recordings/readRecordingLog.ts#L18) now. It should explain why we have this logic in place. We can't remove this logic just yet because users might already have corrupted logs but maybe we could set some date or something at which we remove it?

packages/replayio/src/utils/recordings/readRecordingLog.ts
```diff
@@ -15,6 +15,10 @@ export function readRecordingLog() {
       } catch (err) {
         debug(`Error parsing line:\n${line}`);
 
+        // Early versions of `replayio` could remove the trailing \n from recordings.log,
+        // so the next entry would be appended to the last line, creating a line with two
+        // entries. This workaround lets us read these corrupted entries but it should
+        // be removed eventually.
```
Andarist, line 21: ❤️ thank you

PR #509, opened 2024-06-05
replayio: really remove all recordings

hbenl:
While working on PRO-572 I noticed that recordings without a URL are not removed by `replayio remove --all`. The reason is that these are considered to be empty recordings, so `getRecordings()` will filter them out and the current implementation of `replayio remove --all` doesn't remove recordings that were filtered out by `getRecodings()`.

### DIFF
--- a/packages/replayio/src/commands/remove.ts
+++ b/packages/replayio/src/commands/remove.ts
@@ -3,7 +3,7 @@ import { exitProcess } from "../utils/exitProcess";
 import { findRecordingsWithShortIds } from "../utils/recordings/findRecordingsWithShortIds";
 import { getRecordings } from "../utils/recordings/getRecordings";
 import { printRecordings } from "../utils/recordings/printRecordings";
-import { removeFromDisk } from "../utils/recordings/removeFromDisk";
+import { removeAllFromDisk, removeFromDisk } from "../utils/recordings/removeFromDisk";
 import { selectRecordings } from "../utils/recordings/selectRecordings";
 import { LocalRecording } from "../utils/recordings/types";
 import { dim } from "../utils/theme";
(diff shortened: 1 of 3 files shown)


packages/replayio/src/commands/remove.ts
```diff
@@ -15,6 +15,10 @@ registerCommand("remove")
   .action(remove);
 
 async function remove(shortIds: string[], { all = false }: { all?: boolean }) {
+  if (all) {
+    removeFromDisk();
```
Andarist, line 19: nit: it's not particularly clear what this function does when it doesn't receive any argument. I'd suggest splitting its implementation into two functions, one would require the recording id and the other one would be called smth like `removeAllRecordings`

packages/replayio/src/commands/remove.ts
```diff
@@ -15,6 +15,10 @@ registerCommand("remove")
   .action(remove);
 
 async function remove(shortIds: string[], { all = false }: { all?: boolean }) {
+  if (all) {
+    removeAllFromDisk();
+  }
```
Andarist, line 20: This now gets called prematurely in the sense that it leads to a weird experience. It removes everything from the disk, then reads the recordings log and since there is no recording log at this point this function logs "No recordings found". I think this should be fine-tuned.

Previously we were printing a list of everything that was removed. I'm on the fence when it comes to how useful that was. On one hand the user was notified about what was removed, on the other hand... they really couldn't quite act on it anyhow as far as I know. Printing a potentially huge table of redundant data doesn't seem that useful so perhaps a simple confirmation of what just happened would be enough here

Andarist:
please don't forget to add a user-friendly changeset :) You can do it either by invoking `yarn changeset` or by clicking the link in the bot's comment [here](https://github.com/replayio/replay-cli/pull/509#issuecomment-2149426817)

packages/replayio/src/commands/remove.ts
hbenl: > It removes everything from the disk, then reads the recordings log and since there is no recording log at this point this function logs "No recordings found"

D'oh!

bvaughn (approved):

PR #511, opened 2024-06-05
replayio: sort recording log entries

hbenl:
We can't rely on the ordering of recording log entries, the runtimes sometimes write the `createRecording` entry after another entry for the same recording, which confuses `getRecordings()`.


bvaughn (approved):
> the runtimes sometimes write the createRecording entry after another entry for the same recording, which confuses getRecordings().

:(

PR #513, opened 2024-06-06
Use branch name as an explicit run title for local runs

Andarist:


packages/replay/src/metadata/source.ts
Andarist, line 1: no changes here, it's just require->import + a TS type fix

packages/test-utils/src/reporter.ts
```diff
@@ -187,6 +188,28 @@ export class ReporterError extends Error {
   }
 }
 
+function getFallbackRunTitle() {
+  // for CI runs we don't want to set an explicit title
+  // dashboard is meant to use the commit/PR information retrieved on CI
+  if (process.env.CI) {
```
Andarist, line 194: we could use `ci-info` package to improve this check. `CI` should be set by GitHub Actions and a couple of other vendors though

packages/test-utils/src/reporter.ts
```diff
@@ -187,6 +188,28 @@ export class ReporterError extends Error {
...
+    return;
+  }
+
+  let gitChild;
+
+  try {
+    gitChild = spawnSync("git", ["rev-parse", "--abbrev-ref", "HEAD"]);
```
Andarist, line 201: I'm not super fond of doing this in this package. I started adding this to the source metadata but it also didn't feel right. I concluded this isn't really part of what we call source here. That shape can be rechecked here:
https://github.com/replayio/replay-cli/blob/cdb96e8259d27ec3792b5e05478232b985b3f3b1/packages/replay/src/metadata/source.ts#L235-329

Also, since this is configurable using `config.runTitle` it also didn't fit well into how `sourceMetadata` gets initialized. We also had this `this.runTitle` computed synchronously when the source metadata is computed asynchronously

packages/test-utils/src/reporter.ts
```diff
@@ -187,6 +188,28 @@ export class ReporterError extends Error {
...
+  } catch {
+    return;
+  }
+
+  if (gitChild.status !== 0) {
+    return;
+  }
+
+  return `(local) ${gitChild.stdout.toString().trim()}`;
```
bvaughn, line 210:
```suggestion
  return `(local) ${gitChild.stdout.toString().trim()} branch`;
```

bvaughn (approved):

PR #514, opened 2024-06-06
Add an observability package with Grafana 

miriambudayr:


examples/create-react-app-typescript/package.json
```diff
@@ -5,6 +5,7 @@
     "@playwright/test": "^1.40.1",
     "@replayio/playwright": "workspace:^",
     "@replayio/replay": "workspace:^",
+    "@replayio/telemetry-node": "workspace:^",
```
Andarist, line 8: This example doesn't directly depend on this package so this is redundant:
```suggestion
```

packages/test-utils/package.json
```diff
@@ -6,6 +6,7 @@
   "exports": {
     ".": "./dist/index.js",
     "./testId": "./dist/testId.js",
+    "./telemetry": "./dist/telemetry.js",
```
bvaughn, line 9: Can probably revert this change

packages/telemetry-node/package.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "@replayio/telemetry-node",
+  "version": "1.0.0",
+  "description": "Node utilities for telemetry",
+  "main": "./dist/index.js",
+  "exports": {
+    ".": "./dist/index.js",
+    "./package.json": "./package.json",
+    "./telemetry": "./dist/telemetry.js"
```
Andarist, line 9: this `/telemetry` entrypoint is redundant, the main entrypoint already exports everything from it
```suggestion
    ".": "./dist/index.js",
    "./package.json": "./package.json"
```

packages/telemetry-node/src/telemetry.ts
```diff
@@ -0,0 +1,35 @@
...
+
+export function telemetryDebug(message: string, tags?: Tags) {
+  if (process.env.REPLAY_TELEMETRY_DISABLED) {
+    return;
+  }
+
+  telemetryLogger.debug(message, tags);
+}
+
+export { telemetryLogger };
```
bvaughn, line 35: I wonder if we should avoid exporting this directly, and force everything to go through an intermediate API instead (so we could enforce things like checking the opt-out `process.env.REPLAY_TELEMETRY_DISABLED`)

We'd also need to export a method that wrapped `telemetryLogger.close` but we might want to do that anyway to flush pending requests?

packages/telemetry-node/package.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "@replayio/telemetry-node",
+  "version": "1.0.0",
+  "description": "Node utilities for telemetry",
```
Andarist, line 4: Feel free to tweak this anyhow you want, we probably could even just omit the description. Either way is fine with me
```suggestion
  "description": "Node utilities for Replay.io's own telemetry",
```

packages/telemetry-node/telemetry.d.ts
Andarist, line 1: if we remove this extra entrypoint (like suggested [here](https://github.com/replayio/replay-cli/pull/514/files#r1630031275)) then we'll be able to remove this file too

packages/telemetry-node/telemetry.js
Andarist, line 1: if we remove this extra entrypoint (like suggested [here](https://github.com/replayio/replay-cli/pull/514/files#r1630031275)) then we'll be able to remove this file too

packages/test-utils/package.json
Andarist, line 9: do we expect to import things from `@replayio/test-utils/telemetry`?

tsconfig.json
```diff
@@ -9,6 +9,7 @@
     { "path": "./packages/replayio" },
     { "path": "./packages/sourcemap-upload" },
     { "path": "./packages/sourcemap-upload-webpack-plugin" },
-    { "path": "./packages/test-utils" }
+    { "path": "./packages/test-utils" },
+    {"path": "./packages/telemetry-node"}
```
Andarist, line 13: nit:
```suggestion
    { "path": "./packages/telemetry-node" }
```

packages/playwright/src/reporter.ts
```diff
@@ -24,6 +24,7 @@ import { getPlaywrightBrowserPath } from "@replayio/replay";
 import { FixtureStepStart, ParsedErrorFrame, TestIdData } from "./fixture";
 import { StackFrame } from "./playwrightTypes";
 import { getServerPort, startServer } from "./server";
+import { telemetryLogger } from "@replayio/telemetry-node";
```
Andarist: if we end up using it here, we should add a reference to this package in https://github.com/replayio/replay-cli/blob/7cc25e7717dc8a1bb21b4b3df99f46cafeddc192/packages/playwright/tsconfig.json

The time will probably come soon that I'll have to automate generating those references 😅

packages/telemetry-node/src/telemetry.ts
miriambudayr: Yeah, I thought I was only exporting `telemetryDebug` for this reason, so I'll fix this.

Speaking of `telemetryLogger.close`, I'm assuming we will need to figure out the best place to call that for each package. For the playwright package, I was considering putting it here:

https://github.com/replayio/replay-cli/blob/75ee1cb037de7e86f047ffe79dc84b1423c28aaa/packages/playwright/src/reporter.ts#L319

Andarist: Yeah, `onEnd` would likely be the correct place for this - maybe u could even put it in `onEnd` within the `test-utils`.

What does this method? is it mainly for flushing the events that may have not been sent out yet?

bvaughn: I'm thinking ahead to if we pull Mixpanel into this package as well, as we have an explicit "wait a couple hundred ms to flush pending events before exciting" hook in the new CLI.

miriambudayr: I agree @bvaughn and @Andarist .

When we add Mixpanel to this package, I think we should merge both functionalities under the same function.

packages/playwright/src/reporter.ts
miriambudayr: Got it.

packages/telemetry-node/src/telemetry.ts
miriambudayr: > Yeah, `onEnd` would likely be the correct place for this - maybe u could even put it in `onEnd` within the `test-utils`.
> 
> What does this method? is it mainly for flushing the events that may have not been sent out yet?

Yes.

bvaughn (approved):
I don't really know much about this "winston" package, nor about Grafana, so I'm not really looking too closely at the `src/grafana` contents.

Both of these NPM packages are kind of large (at least according to bundlephobia) but maybe that isn't a big deal for a CLI dependency.

I'm probably more interested in follow up work we might do like:
* What will it look like to add a second logger, like Mixpanel?requirements?)
* What will it look like to add the user's access token? (Where does that fit into the flow in terms of registering a user's session and how does that impact message queueing/flushing?)

packages/observability-node/package.json
```diff
@@ -0,0 +1,34 @@
+{
+  "name": "@replayio/observability-node",
+  "version": "1.0.0",
```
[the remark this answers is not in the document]
Andarist: Whatever version is used here will just get published together with the next release. So it's up to us to say what that version should be. The lower the better here, I think - so 0.0.1 works perfectly 👍

Andarist:
> Both of these NPM packages are kind of large (at least according to bundlephobia) but maybe that isn't a big deal for a CLI dependency.

Those will also get used outside of the CLI as we want to add telemetry to the test plugins. I'd love smaller deps but I think it's more important for us to have this than not so overall I'm just no concerned about it.

> I'm probably more interested in follow up work we might do like:

Same. Do we have to make this package's API so specific? The public exports mention Grafana and that feels like an implementation detail. I have imagined that we'd abstract this more here but we can also start "small", I'm fine with that. 

---

_This is just a comment. Don't treat it as blocking or anything. I will be away for the weekend - feel free to land this if needed. It's already approved by Brian :)_

bvaughn:
> Those will also get used outside of the CLI as we want to add telemetry to the test plugins.

That's fair! I think I was really thinking more along the lines of...so long as we aren't adding these bytes to our website, I don't care as much. Developer laptops (and CLI) can afford to load them.

miriambudayr:
> > Both of these NPM packages are kind of large (at least according to bundlephobia) but maybe that isn't a big deal for a CLI dependency.
> 
> Those will also get used outside of the CLI as we want to add telemetry to the test plugins. I'd love smaller deps but I think it's more important for us to have this than not so overall I'm just no concerned about it.
> 
> > I'm probably more interested in follow up work we might do like:
> 
> Same. Do we have to make this package's API so specific? The public exports mention Grafana and that feels like an implementation detail. I have imagined that we'd abstract this more here but we can also start "small", I'm fine with that.
> 
> _This is just a comment. Don't treat it as blocking or anything. I will be away for the weekend - feel free to land this if needed. It's already approved by Brian :)_

Personally I would prefer to start small but I also do understand what you mean and I agree.

PR #516, opened 2024-06-07
Fixed a race condition that could cause some tests not being added correctly to a test run

Andarist:
`CreateTestRunShard` is an async call and it's not guaranteed that it completes before we start working on `AddTestsToShard` (or even `CompleteTestRunShard`).

This fixes that problem by always awaiting the shard id "task" before we continue with other calls that depend on this shard ID.

### DIFF
--- a/packages/test-utils/src/reporter.ts
+++ b/packages/test-utils/src/reporter.ts
@@ -82,7 +82,9 @@ export interface PendingWorkError<K extends PendingWorkType> {
 export interface PendingUploadError extends PendingWorkError<"upload"> {
   recording: RecordingEntry;
 }
-type PendingWorkEntry<K extends PendingWorkType, T = {}> = PendingWorkError<K> | (T & { type: K });
+type PendingWorkEntry<K extends PendingWorkType, T = {}> =
+  | PendingWorkError<K>
+  | (T & { type: K; error?: never });
 type TestRunPendingWork = PendingWorkEntry<
   "test-run",
   {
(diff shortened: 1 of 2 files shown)


hbenl (approved):

packages/test-utils/src/reporter.ts
```diff
@@ -200,6 +202,7 @@ class ReplayReporter<TRecordingMetadata extends UnstructuredMetadata = Unstructu
   upload = false;
   filter?: (r: RecordingEntry<TRecordingMetadata>) => boolean;
   recordingsToUpload: ExternalRecordingEntry<TRecordingMetadata>[] = [];
+  _testRunShardIdPromise: Promise<TestRunPendingWork> | null = null;
```
bvaughn, line 205: Any reason not to just make this private?
```suggestion
  private testRunShardIdPromise: Promise<TestRunPendingWork> | null = null;
```

Andarist: Not making it private is an oversight on my part but I think it's still worth maintaining the `_` convention, even when using `private` keyword.

bvaughn: Huh. Why?

Andarist: `private` doesn't hide it at a runtime level, just at the type level. I find it useful to know what's private and what's not when inspect values in the debugger

bvaughn: That's fair. I guess it depends on what our compile target is. I was assuming TS would get compiled to `#testRunShardIdPromise` but I guess it doesn't?

Andarist: Ye, it doesn't do that - regardless of the target. To use private fields u have to use `#` syntax, not the `private` keyword

bvaughn: TIL

Bad assumption on my part about TS :) Thanks for sharing

PR #517, opened 2024-06-07
Delete postinstall

jasonLaster:
This deletes the other post install scripts that we have which will also make this more consistent with playwright and our docs which suggest you run `replayio install`


bvaughn (commented):
Seems like we should also be deleting the "first-run" files in src?

jasonLaster:
good catch. thought i had

packages/cypress/package.json
```diff
@@ -17,7 +17,6 @@
     "*.d.ts"
   ],
   "scripts": {
-    "postinstall": "node ./first-run.js",
```
Andarist, line 20: this is a breaking change, so it deserves a major bump in the relevant packages - and we might want to take a look if there are any other breaking changes in those packages that we'd like to batch together with this

jasonLaster: nothing comes to mind. will defer to you.

Andarist:
superseded by https://github.com/replayio/replay-cli/pull/522

PR #518, opened 2024-06-07
Update `sourcemap-upload `package to support deletion and correlating sourcemaps by filename

markerikson:
This PR:

- Updates the `source-map` upload package to accept two new options:
  - `deleteAfterUpload`: deletes any uploaded sourcemaps after the upload is done
  - `matchSourcemapsByFilename`: tries to look for a corresponding `.map` file on disk, in case the generated source did not having a sourcemap URL comment (such as if Webpack is set to `devtool: "hidden-source-map"`)
- Adds doc comments to all exported option fields
- Pastes the options type and description into the README

### DIFF
--- a/packages/sourcemap-upload/src/index.ts
+++ b/packages/sourcemap-upload/src/index.ts
@@ -26,6 +26,7 @@ export interface UploadOptions {
   group: string;
   key?: string;
   dryRun?: boolean;
+  matchSourcemapsByFilename?: boolean;
   extensions?: Array<string>;
   ignore?: Array<string>;
   root?: string;
@@ -49,7 +50,10 @@ export async function uploadSourceMaps(opts: UploadOptions): Promise<void> {
   );
   assert(
     typeof opts.dryRun === "boolean" || opts.dryRun === undefined,
-    "'dryRun' must be a string or undefined"
+  assert(
+    typeof opts.matchSourcemapsByFilename === "boolean" ||
+      opts.matchSourcemapsByFilename === undefined,
+    "'matchSourcemapsByFilename' must be a boolean or undefined"
   );
   assert(
     (Array.isArray(opts.extensions) &&
@@ -110,6 +114,7 @@ export async function uploadSourceMaps(opts: UploadOptions): Promise<void> {
     groupName: opts.group,
     apiKey,
     dryRun: !!opts.dryRun,
+    matchSourcemapsByFilename: !!opts.matchSourcemapsByFilename,
     extensions: opts.extensions || [".js", ".map"],
     ignorePatterns: opts.ignore || [],
     rootPath: path.resolve(cwd, opts.root || ""),
@@ -126,6 +131,7 @@ interface NormalizedOptions {
   groupName: string;
   apiKey: string;
   dryRun: boolean;
+  matchSourcemapsByFilename: boolean;
   extensions: Array<string>;
   ignorePatterns: Array<string>;
   rootPath: string;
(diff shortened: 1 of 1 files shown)


packages/sourcemap-upload/README.md
```diff
@@ -28,50 +28,68 @@ and vice-versa.
...
-
   /**
    * Run all of the local processing and searching for maps, but skip uploading them.
    */
   dryRun?: boolean;
-
+  /**
+   * Delete all found sourcemap files after they have been uploaded.
+   */
+  deleteAfterUpload?: boolean;
```
Andarist, line 57: So we'll recommend people switch to this from `webpack-delete-sourcemaps-plugin` if they happen to use that plugin. The goal of this option is to serve as a replacement for that other thing, right?

.changeset/tender-ghosts-obey.md
```diff
@@ -0,0 +1,5 @@
+---
+"@replayio/sourcemap-upload": patch
+---
+
+Update `sourcemap-upload `package to support deletion and correlating sourcemaps by filename
```
Andarist, line 5: This is going to be part of the CHANGELOG of this `@replayio/sourcemap-upload` package. So it should be written from its PoV. Mentioning "sourcema-upload" here is redundant. 

You could also mention what kind of options where added. It's helpful to know what's going to be possible now but some extra references to *how* would be useful too

packages/sourcemap-upload/README.md
markerikson: Yep!  Right now our doc page at https://docs.replay.io/reference/replay-cli/source-maps#next-js has a fictional `deleteSourcemaps()` function in the example. I was going to update that to direct people to use the deletion plugin, ran into issues, and said "I'll just do it myself".  Plus it's simpler this way anyway :)

packages/sourcemap-upload/README.md
```diff
@@ -28,50 +28,68 @@ and vice-versa.
...
+  /**
+   * If sourcemaps can't be matched to generated files by their sourceMappingURL, try matching by filenames on disk
+   */
+  matchSourcemapsByFilename?: boolean;
```
jcmorrow, line 61: Is there any reason not to default this to `true`? It seems like very reasonable default behavior.

markerikson: Can't immediately think of a reason _against_ having a default of `true`.

In practice, this shouldn't matter often, because normally generated bundles _will_ have a `sourceMappingURL` comment.  I just happened to find that they _don't_ when you set `devtool: "hidden-source-map"`, and only tried that because the delete plugin suggested it to avoid 404s.

Andarist: > I just happened to find that they don't when you set devtool: "hidden-source-map"

Isn't this a popular setting though? At least this is what I was always using to avoid my source maps being publicly available.

I'd probably make this a default too if we don't have any concerns around this.

markerikson: I'd actually never even _heard_ of `"hidden-source-map"` until now :)

But yes, went ahead and made this setting default to `true`

.changeset/tender-ghosts-obey.md
```diff
@@ -0,0 +1,5 @@
+---
+"@replayio/sourcemap-upload": patch
```
Andarist, line 2: Even though the other package will already benefit automatically~ from this change, I'd probably call it out explicitly here:
```suggestion
"@replayio/sourcemap-upload": patch
"@replayio/sourcemap-upload-webpack-plugin": patch
```

Andarist (approved):

PR #519, opened 2024-06-10
Implement new `upload` options

Andarist:
This PR implements a more advanced `upload` options. We now allow users to specify `statusThreshold` for the recordings we should upload. We also allow users to `minimizeUploads`.

Those options make common uploading strategies simpler to implement. Users might want to upload only recordings with failed tests. Or at most one recording per failed test, or at most two per flake etc. We control those settings now and thus we minimize the common mistakes surface. 

This also deprecates the `filter` option. From the user's PoV, it's way easier to think in terms of tests than recordings. There is no 1 to 1 correspondence between a test and its recordings.


packages/test-utils/src/reporter.ts
```diff
@@ -41,14 +41,34 @@ export interface TestIdContext {
...
+export type UploadStatusThreshold = "all" | "failed-and-flaky" | "failed";
+
+type UploadStatusThresholdInternal = UploadStatusThreshold | "none";
+
+export type UploadOption =
+  | boolean
+  | {
+      minimizeUploads?: boolean;
+      statusThreshold?: UploadStatusThreshold;
+    };
```
Andarist, line 53: I just realized that Playwright has somewhat similar options for videos/traces/screenshots:
https://github.com/microsoft/playwright/blob/664c4dd44297fc8147d10734b617801a6c464609/packages/playwright/types/test.d.ts#L5130-L5132

Maybe there would be value in aligning our functionality with theirs here

bvaughn: Interesting...so like `uploadMode` rather than `statusThreshold`?

I don't know if this feels like it should be a goal necessarily, because it might end up being difficult to map shared concepts like this (in `test-utils`) to Playwright and Cypress _both_ in a way that lines up with their own naming conventions, and changing it to only match one seems odd (or at least, there's no clear "winner" to me in between the two).

I think I'm in favor of sticking with the names we settled on, but I'm totally open for discussing it and reconsidering.

bvaughn:
Reviewing...

packages/cypress/src/reporter.ts
```diff
@@ -23,7 +23,13 @@ type ReplayCypressRecordingMetadata = {
   test: TestMetadataV2.TestRun;
 };
 
-export interface PluginOptions extends ReplayReporterConfig<ReplayCypressRecordingMetadata> {}
+// `filter` is re-applied here so its deprecated comment gets lost here
```
bvaughn, line 26: I don't understand this comment.

packages/playwright/src/reporter.ts
```diff
@@ -242,16 +242,16 @@ class ReplayPlaywrightReporter implements Reporter {
     // skipped tests won't have a reply so nothing to do here
     if (status === "skipped") return;
 
-    const testMetadata = {
+    const testIdData = {
```
bvaughn, line 245: "metadata" seems more technically correct so I'm curious why this rename?

(No need to change it, just wondering.)

packages/test-utils/src/reporter.ts
```diff
@@ -59,21 +79,11 @@ export interface TestRunner {
...
-  ? TestMetadataV1.TestError
-  : TestMetadataV2.TestError;
-type TestRun = ReplayReporter["schemaVersion"] extends "1.0.0"
-  ? TestMetadataV1.TestRun
-  : TestMetadataV2.TestRun;
+type UserActionEvent = TestMetadataV2.UserActionEvent;
+type Test = TestMetadataV2.Test;
+type TestResult = TestMetadataV2.TestResult;
+type TestError = TestMetadataV2.TestError;
+type TestRun = TestMetadataV2.TestRun;
```
bvaughn, line 86: Nice 👍🏼

packages/test-utils/src/reporter.ts
```diff
@@ -42,14 +42,34 @@ export interface TestIdContext {
...
 export interface ReplayReporterConfig<
   TRecordingMetadata extends UnstructuredMetadata = UnstructuredMetadata
 > {
   runTitle?: string;
   metadata?: Record<string, any> | string;
   metadataKey?: string;
-  upload?: boolean;
+  upload?: UploadOption;
   apiKey?: string;
+  /** @deprecated Use `upload.minimizeUploads` and `upload.statusThreshold` instead */
```
bvaughn, line 72: Nice 👍🏼

packages/test-utils/src/reporter.ts
```diff
@@ -309,32 +322,44 @@ class ReplayReporter<TRecordingMetadata extends UnstructuredMetadata = Unstructu
...
       );
     }
+    this._minimizeUploads = config.upload
+      ? typeof config.upload === "object" && typeof config.upload.minimizeUploads === "boolean"
+        ? config.upload.minimizeUploads
+        : false
+      : false;
+    this._uploadStatusThreshold = config.upload
+      ? (typeof config.upload === "object" && config.upload.statusThreshold) || "all"
+      : "none";
```
bvaughn, line 351: I wonder if we could write this in a way that's a bit less _dense_? I think it's easy to misread the semantics for upload status threshold in particular ("all" vs "none")

I wonder if this would be a little clearer/easier to read? TBH I'm not even sure.

```ts
if (config.upload) {
  if (isAdvancedUploadOption(config.upload)) {
    this._minimizeUploads = config.upload.minimizeUploads ?? false;
    this._uploadStatusThreshold = config.upload.statusThreshold ?? "all";
  } else {
    this._uploadStatusThreshold = "all";
  }
}
```

packages/test-utils/src/reporter.ts
```diff
@@ -42,14 +42,34 @@ export interface TestIdContext {
...
+export type UploadStatusThreshold = "all" | "failed-and-flaky" | "failed";
+
+type UploadStatusThresholdInternal = UploadStatusThreshold | "none";
+
+export type UploadOption =
+  | boolean
+  | {
+      minimizeUploads?: boolean;
+      statusThreshold?: UploadStatusThreshold;
+    };
```
bvaughn, line 54: nit: If we'll have to type check this in a few places, maybe it would be nice to make a type refinement util for it?

```ts
type AdvancedUploadOption = {
  minimizeUploads?: boolean;
  statusThreshold?: UploadStatusThreshold;
};

export type UploadOption =
  | boolean
  | AdvancedUploadOption;

function isAdvancedUploadOption(upload: UploadOption): upload is AdvancedUploadOption {
  return typeof config.upload === "object" && config != null;
}
```

Maybe not worth it if it's only in one spot though. It's just a habit I have.

packages/test-utils/src/reporter.ts
```diff
@@ -873,48 +907,90 @@ class ReplayReporter<TRecordingMetadata extends UnstructuredMetadata = Unstructu
...
+      if (result.result === "passed" || result.attempt >= result.maxAttempts) {
+        this._enqueueUploads(uploadableResults);
+      }
     }
+  }
 
-    if (this.pendingWork.length === 0) {
-      return [];
+  private _enqueueUploads(results: UploadableTestRunResult<TRecordingMetadata>[]) {
+    const latestResult = results[results.length - 1];
```
bvaughn, line 934: nit: I know that this method is only ever called with at least one thing in the array, but it's probably still worth adding an assert to ensure it stays that way?

packages/test-utils/src/reporter.ts
```diff
@@ -873,48 +907,90 @@ class ReplayReporter<TRecordingMetadata extends UnstructuredMetadata = Unstructu
...
+
+    let toUpload: typeof results | undefined;
+
+    switch (this._uploadStatusThreshold) {
+      case "all":
+        toUpload = this._minimizeUploads ? [latestResult] : results;
+        break;
+      case "failed-and-flaky":
+        if (latestResult.result !== "passed" || results.length > 1) {
```
bvaughn, line 943: Maybe this could use an inline comment. My first thought was that `results.length > 1` would be sufficient but maybe retries are disabled.

bvaughn (approved):

packages/cypress/src/reporter.ts
Andarist: Take a look at the behavior here:
[TS playground](https://www.typescriptlang.org/play/?ts=5.4.5#code/JYOwLgpgTgZghgYwgAgEJwM4oN4FgBQyyA9AFSnIACAJhAA5QQJyTXKnEFEzAA2kUAFzIAFAA9hAVxABrEAHsA7iACUyALwA+ZACN583hDggA3AQC+BAqAHwkyAGL6AjMghjIIahjSYcl-FoEXjhGZAR5EAwwZBgXYSd5ZzN8OKSAOh5+aBTrcGg7FESAJjcPCC8fdCxkPEJYvgFhaogAbQAiLIF2gF0UgKCQsIiomLTihP1ilPHMxpySYmQFGIQoeQwsNnlJGNQAVQAVKloGJhYINmi+XmQ4OjojKB9I5AALeQA3aCt8GwLEEV9ABmMqebzIADyAFtgGAADwtAA0yE68yg7W0dW46Oafg6XWgvX6BEGoRQI2isRBk3kwJmILm2SgJkWy3kq3Wm0uyB2MWMbBo9EYzFYyGo8ggGBAAHJ+Q8nryQO8vj98EA)

packages/playwright/src/reporter.ts
Andarist: we have a lot of different metadata types and I'm getting lost in them ;p this object gets passed to `this.getStepsFromFixture` and `this.getFixtureData. In both of them, the relevant parameter is typed using `TestIdData` type so this rename aligne the name with that.

packages/cypress/src/reporter.ts
bvaughn: Gotcha. I think "_its deprecated comment gets lost here_" is maybe still unclear as a comment though. I guess it doesn't much matter since we plan to remove this too soonish :)

packages/playwright/src/reporter.ts
bvaughn: Okay 😅

Andarist:
@bvaughn I'm landing this after addressing all of your comments, thanks for the review!

PR #520, opened 2024-06-10
Remove accidental mutation

bvaughn:
Check out [go/r/5bb1c16d-abc8-4c69-962a-9bef52b0efa9](https://app.replay.io/recording/pro-614--5bb1c16d-abc8-4c69-962a-9bef52b0efa9) for more.

### DIFF
--- a/packages/cypress/src/steps.ts
+++ b/packages/cypress/src/steps.ts
@@ -3,8 +3,8 @@
 import { ReporterError, TestMetadataV2 } from "@replayio/test-utils";
 import Debug from "debug";
 import { AFTER_EACH_HOOK } from "./constants";
-import type { StepEvent } from "./support";
 import { Errors, assertCurrentTest, assertMatchingStep, isStepAssertionError } from "./error";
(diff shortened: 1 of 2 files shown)


Andarist (approved):
This needs a changeset, other than that LGTM :) Nice find.

packages/cypress/src/steps.ts
```diff
@@ -281,10 +281,13 @@ function groupStepsByTest(tests: Test[], steps: StepEvent[]): Test[] {
         "beforeAll",
       ];
       for (const phase of phases) {
-        const stepWithError = t.events[phase].reverse().find(t => t.data.error);
-        if (stepWithError) {
-          t.error = stepWithError.data.error;
-          break;
```
bvaughn, line 287: `Array.prototype.reverse` mutates the events array in place. This is _not_ what we want.

Furthermore it's kind of wasteful to call this method anyway. We could use `Array.prototype.findLast` except I'm not sure what our minimum Node target is so...

PR #522, opened 2024-06-11
Remove `postinstall` script from `@replayio/cypress`

Andarist:


packages/cypress/src/bin.ts
Andarist, line 1: We are still left with a CLI script here which is somewhat odd (and different from `@replayio/playwright` now). This has support for some kind of a flake/internal problem-detector. It was implemented in https://github.com/replayio/replay-cli/pull/151 and we mention it in the docs so I don't want to touch this.

hbenl (approved):

packages/cypress/src/bin.ts
bvaughn: I wish we knew if anyone was actually using it so we maybe _could_ get rid of it.

bvaughn: Looking at the motivation for that script, [SCS-766](https://linear.app/replay/issue/scs-766), I'm surprised to hear that we're referencing it in our docs. Seems like an internal tool that we should have a separate, internal script/package for.

bvaughn: Reading #151 PR description:
> * Adds support for running cypress + replay in one of 4 modes:
>  * `record` - Record everything (the default)
>  * `record-on-retry` - Only record failures
>  * `diagnostics` - Runs cypress a bunch of times (no record, record, record without asserts, and record with each feature disabled). `--level` option allows selection between "basic" (no record, record, record with asserts disabled) and "full" (basic + running with each feature disabled).
>  * `stress` - Is really just `cypress-repeat` with a preset `-n` of 10

Some of this ("diagnostics") sounds like it should be an internal script only used by us and some of it ("record-on-retry") sounds like what we're adding in [PRO-549](https://linear.app/replay/issue/PRO-549). The rest ("stress") is doable with vanilla Cypress I think.

I vote we remove this script as part of the major release.

PR #525, opened 2024-06-12
Removed the CLI script from the `@replayio/cypress` package

Andarist:


hbenl (approved):
Approved assuming that you've checked that the environment variables from `getDiagnosticConfig()` aren't needed anywhere else (?)

Andarist:
I was as cautious as I could be about this 🤞

PR #526, opened 2024-06-12
Add "replayio open" command to launch browser in non-recording mode

bvaughn:
This has been requested so that people can do things like authentication (without Replay recording) before starting recording.

### DIFF
--- a/packages/replayio/src/utils/browser/launchBrowser.ts
+++ b/packages/replayio/src/utils/browser/launchBrowser.ts
@@ -11,10 +11,11 @@ import { getBrowserPath } from "./getBrowserPath";
 export async function launchBrowser(
   url: string,
   options: {
-    processGroupId: string;
+    processGroupId?: string;
+    record?: boolean;
   }
 ) {
-  const { processGroupId } = options;
+  const { processGroupId, record = true } = options;
 
   const profileDir = join(runtimePath, "profiles", runtimeMetadata.runtime);
   ensureDirSync(profileDir);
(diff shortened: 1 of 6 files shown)


packages/replayio/src/utils/browser/launchBrowser.ts
```diff
@@ -28,9 +29,10 @@ export async function launchBrowser(
   ];
   const processOptions = {
     env: {
-      RECORD_ALL_CONTENT: "1",
+      RECORD_ALL_CONTENT: record ? "1" : undefined,
+      RECORD_REPLAY_DONT_RECORD: record ? undefined : "1",
```
Andarist, line 33: TIL, what is this? test plugins dont set this at all (IIRC)

Andarist: sideways: do u know what `RECORD_REPLAY_DRIVER` is for? test plugins use it but we don't use it here

Andarist (approved):

packages/replayio/src/utils/browser/launchBrowser.ts
bvaughn: TIL too. Jason mentioned it in Discord. All I know about it is that if it's set, it tells the runtime not to record anything.

bvaughn: > sideways: do u know what `RECORD_REPLAY_DRIVER` is for? test plugins use it but we don't use it here

No. Maybe we should remove it too? I don't remember why it's here. :)

PR #530, opened 2024-06-13
Fixed `displayError` being incorrectly used for all tests attempts

Andarist:
This PR addresses one of the issues recognized in https://linear.app/replay/issue/PRO-597


packages/cypress/package.json
```diff
@@ -54,6 +54,6 @@
     "ws": "^8.14.2"
   },
   "peerDependencies": {
-    "cypress": ">=5.3.0"
+    "cypress": "^13"
```
Andarist, line 57: Cypress 13 was released almost one year ago (08/29/2023). We don't have any tests for this whole integration, trying to test compatibility manually with an extended range of versions is a nightmare. I don't feel we can quite guarantee compatibility right now with other versions, it's impractical.

If requested by clients we can always loosen up this range after manually confirming that a particular combination works. We'll release a new major of this package soon so I think we should grab the opportunity and bump this here.

Note that supporting more right now would require some extra work here, I changed this because I ran into typing problems (that likely were related to breaking changes in Cypress).

packages/cypress/src/steps.ts
```diff
@@ -96,12 +97,20 @@ function getTestsFromResults(
...
-            message: result.displayError,
-          }
-        : null,
+      // attempt.error is available here:
+      // https://github.com/cypress-io/cypress/blob/61156808413be8b99264026323ce3abfb22c4c26/packages/server/lib/modes/results.ts#L20
+      // but it's lost when creating a public test result:
+      // https://github.com/cypress-io/cypress/blob/61156808413be8b99264026323ce3abfb22c4c26/packages/server/lib/modes/results.ts#L111
+      // `.displayError` represents the error reported by the last attempt
+      // for all of the previous attempts we rely on the search for the last errored step in `groupStepsByTest`
+      // if an error is found there, it will be set on the test, the information set here is not overriden though
```
Andarist, line 106: This logic is a little bit wonky but it makes sense. The `.displayError` is something computed by Cypress so it should have higher accuracy than what we can compute on our own. OTOH, we have to compute this on our own for non-last attempts because Cypress doesn't expose this information to us.

It would be great to refactor this so we don't have to rely on a mutation *elsewhere* but that's out of this PR's scope

packages/cypress/src/support.ts
```diff
@@ -579,7 +579,7 @@ function register() {
           logs: () => [],
           add: () => {},
           get: (key?: any) => (!key ? log : log[key]),
-          toJSON: () => log,
+          toJSON: () => [],
```
Andarist, line 582: the previous version was completely crashing Cypress, the `toJSON`'s type here expects `string[]` so I used an empty array to ensure that it's compatible. In Cypress 10 `log` was typed as `any` and maybe at that time it worked OK, dunno. Today it's an object so we can see how it isn't compatible with the expected `string[]`.

I verified that we can still render `cy.log` just fine with this. It's quite plausible that we just don't rely on the return value of this function.

hbenl (approved):

packages/cypress/package.json
bvaughn: I support this.

packages/cypress/src/steps.ts
```diff
@@ -78,11 +78,12 @@ function getTestsFromResults(
...
     const scope = [...result.title];
     const title = scope.pop()!;
+    const lastAttemptIndex = result.attempts.length - 1;
 
     return result.attempts.map<Test>((a, attemptIndex) => ({
       id: getIdForTest(result) ?? attemptIndex,
-      // Cypress 10.9 types are wrong here ... duration doesn't exist but wallClockDuration does
-      approximateDuration: a.duration || (a as any).wallClockDuration || 0,
+      // those properties don't exist since Cypress 13: https://github.com/cypress-io/cypress/pull/27230
+      approximateDuration: (a as any).duration || (a as any).wallClockDuration || 0,
```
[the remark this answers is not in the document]
bvaughn: Can we add a Linear task + TODO comment to follow up here? Else it will get forgotten about.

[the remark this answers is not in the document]
Andarist: done, PRO-640

Issue #534, opened 2024-06-14
ESM Error when trying to start replay CLI


FleetAdmiralJakob:
I installed replay on my windows machine using `pnpm i -g replayio`

trying to execute: 

`replayio record https://first.replay.io`

getting:

```
D:\Program Files (x86)\pnpm\global\5\.pnpm\replayio@1.3.2_encoding@0.1.13\node_modules\replayio\dist\utils\formatting.js:7
const strip_ansi_1 = __importDefault(require("strip-ansi"));
                                     ^

Error [ERR_REQUIRE_ESM]: require() of ES Module D:\Program Files (x86)\pnpm\global\5\.pnpm\strip-ansi@7.0.1\node_modules\strip-ansi\index.js from D:\Program Files (x86)\pnpm\global\5\.pnpm\replayio@1.3.2_encoding@0.1.13\node_modules\replayio\dist\utils\formatting.js not supported.
Instead change the require of index.js in D:\Program Files (x86)\pnpm\global\5\.pnpm\replayio@1.3.2_encoding@0.1.13\node_modules\replayio\dist\utils\formatting.js to a dynamic import() which is available in all CommonJS modules.
    at Object.<anonymous> (D:\Program Files (x86)\pnpm\global\5\.pnpm\replayio@1.3.2_encoding@0.1.13\node_modules\replayio\dist\utils\formatting.js:7:38) {
  code: 'ERR_REQUIRE_ESM'
}

Node.js v20.10.0
```

Andarist:
The fix is on the way: https://github.com/replayio/replay-cli/pull/535 . Note that, at the moment, we don't have full Windows support.

PR #535, opened 2024-06-14
Add an explicit dependency on `strip-ansi` module

Andarist:
fixes https://github.com/replayio/replay-cli/issues/534

Without an explicit dependency we accidentally relied on node modules hoisting. It could load `strip-ansi@7.x` this way and that's esm-only. I pinned it to 6.x here since it's the one that supports CJS


bvaughn (approved):
Cool thx

PR #536, opened 2024-06-17
Correctly identify test executions by utilizing `repeatEachIndex`

Andarist:


packages/replay/src/metadata/test/v2.ts
```diff
@@ -131,23 +133,23 @@ const v2_1_0 = assign(
...
+  export type Test = Infer<typeof test_v3_0_0>;
   export type TestResult = Infer<typeof testResult>;
-  export type TestRun = Infer<typeof v2_2_0>;
+  export type TestRun = Infer<typeof v3_0_0>;
   export type TestError = Infer<typeof testError>;
 }
 
 export default {
-  "2.2.0": v2_2_0,
+  "3.0.0": v3_0_0,
```
Andarist, line 152: TODO: I discovered that devtools might depend on properties that I wanted to remove. So this version should now be reverted to 2.2.0

packages/test-utils/src/reporter.ts
```diff
@@ -971,7 +995,12 @@ class ReplayReporter<TRecordingMetadata extends UnstructuredMetadata = Unstructu
     this._pendingWork.push(
       ...toUpload
         .flatMap(result => result.recordings)
-        .filter(r => (this._filter ? this._filter(r) : true))
+        .filter(r => {
+          if (this._filter && this._filter(r)) {
+            return !this._uploadedRecordings.has(r);
```
Andarist, line 1000: After consideration, I don't quite like this. It probably would be better to "consume" those stored results and remove them from memory altogether before passing them to `_uploadRecording`. This way we also wouldn't try to upload the same recording twice.

I probably should recheck if such a change won't impact flakiness detection that is done in the switch/cace above

bvaughn: Want me to hold off reviewing then?

Andarist:
@bvaughn @hbenl it's good to be reviewed now

packages/playwright/src/reporter.ts
```diff
@@ -129,23 +133,31 @@ class ReplayPlaywrightReporter implements Reporter {
...
+    return this.fixtureData[id];
   }
 
-  getTestId(test: TestCase) {
-    return test.titlePath().join("-");
+  // Playwright alrady provides a unique test id:
+  // https://github.com/microsoft/playwright/blob/6fb214de2378a9d874b46df6ea99d04da5765cba/packages/playwright/src/common/suiteUtils.ts#L56-L57
+  // this is different because it includes `repeatEachIndex` and `attempt`
+  // TODO: this could be simplified to `${test.testId}-${test.repeatEachIndex}-${test.attempt}`
+  // before doing that all recipients of `TestExecutionIdData` should be rechecked to see if such a change would be safe
```
bvaughn, line 151: Can you file a Linear follow-up issue for this and update the TODO comment to reference it?

packages/test-utils/src/reporter.ts
```diff
@@ -37,31 +41,37 @@ interface TestRunTestInputModel {
...
 export type UploadStatusThreshold = "all" | "failed-and-flaky" | "failed";
 
 type UploadStatusThresholdInternal = UploadStatusThreshold | "none";
 
 export type UploadOption =
   | boolean
   | {
+      /**
+       * It minimizes uploads related to a single test within a shard
+       */
```
bvaughn, line 53: I think it might be useful to be slight more verbose here, e.g.

```suggestion
      /**
       * Minimize the number of recordings uploaded for a test attempt (within a shard).
       * e.g. Only one recording would be uploaded for a failing test attempt, regardless of retries.
       * e.g. Two recordings would be uploaded for a flaky test attempt (the passing test and one of the failures).
       */
```

packages/test-utils/src/reporter.ts
```diff
@@ -920,44 +917,83 @@ class ReplayReporter<TRecordingMetadata extends UnstructuredMetadata = Unstructu
...
+    assert(!!latestExecution, "Expected at least one execution in the list");
 
+    let toUpload: typeof executions | undefined;
     switch (this._uploadStatusThreshold) {
       case "all":
-        toUpload = this._minimizeUploads ? [latestResult] : results;
+        // even when `minimizeUploads` is combined with `repeatEach` it always makes sense to upload the latest result
+        // otherwise, we could upload a single successful attempt without uploading a potential failure of the same test
+        // coming from a different `repeatEachIndex`
+        toUpload = this._minimizeUploads ? [latestExecution] : executions;
```
bvaughn, line 956: This looks wrong to me.

What if the test was flaky? We're only getting the passing recording this way.

bvaughn: I wonder if this should be rewritten so that we have the pass/fail/flaky check on the outside and then the minimize check within?

bvaughn: Something like this?

```js
const status = computeStatus();
switch (status) {
  case "failed": {
    switch (this._uploadStatusThreshold) {
      case "all":
      case "failed":
      case "failed-and-flaky": {
        toUpload = this._minimizeUploads ? [latestExecution] : executions;
        break;
      }
    }
    break;
  }
  case "flaky": {
    switch (this._uploadStatusThreshold) {
      case "all":
      case "failed-and-flaky": {
        if (this._minimizeUploads) {
          toUpload = [executions[0], executions[executions.length - 1]];
        } else {
          toUpload = executions;
        }
        break;
      }
    }
    break;
  }
  case "pass": {
    switch (this._uploadStatusThreshold) {
      case "all": {
        toUpload = this._minimizeUploads ? [latestExecution] : executions;
        break;
      }
    }
    break;
  }
}
```

packages/playwright/src/reporter.ts
Andarist: done

packages/test-utils/src/reporter.ts
```diff
@@ -920,44 +919,69 @@ class ReplayReporter<TRecordingMetadata extends UnstructuredMetadata = Unstructu
...
-          } else if (results.length > 1) {
-            toUpload = [results[0], latestResult];
-          } else {
-            toUpload = [latestResult];
-          }
+        // if there is no failed execution then this test is not flaky or failing
+        if (
+          latestExecution.result !== "passed" ||
+          executions.length > 1 ||
+          (result.uploadedStatuses.failed && !result.uploadedStatuses.passed)
```
bvaughn, line 967: This looks super funky.

packages/test-utils/src/reporter.ts
```diff
@@ -37,31 +41,39 @@ interface TestRunTestInputModel {
...
   recordings: RecordingEntry<TRecordingMetadata>[];
   result: TestRun["result"];
   testId: string;
 }
 
+interface UploadableTestResult<TRecordingMetadata extends UnstructuredMetadata> {
+  executions: Record<string, UploadableTestExecutionResult<TRecordingMetadata>[]>;
+  uploadedStatuses: {
+    passed: boolean;
+    failed: boolean;
```
bvaughn, line 73: I don't think I understand what this is supposed to mean.

packages/test-utils/src/reporter.ts
bvaughn: Like is this really supposed to be in the `||` chain? I don't really understand the `uploadedStatuses` data structure (I think it could use some inline comments at least) but my intuition is that maybe this is more of an `&&` conditional, else it seems like we would get false positives here?

packages/test-utils/src/reporter.ts
Andarist: This function uses executions from a single execution group (that currently is equivalent to `repeatEachIndex`) but the uploads are minimized per test id.

We'd have an easier time if we'd push this whole decision until `onEnd` (that runs after everything gets completed). I decided to keep it in `onTestEnd` to start uploads asap - so we only have to await those upload jobs in `onEnd`. Maybe it's not worth it but it felt nice to have it "paralellized".

We can't determine if a test case is flaky based on a signal execution group because both could have only single executions and each of them could be of a different status. I thought about having an aggregate status on this uploadable result but it felt like a complication at a time.

The version above misses that those flakes can be spread across execution groups. I like your suggestion to reorganize this, I'll try to come up with something better.

packages/test-utils/src/reporter.ts
Andarist: this is meant to track what has been uploaded for flaky tests since those flakes can be in different execution groups

bvaughn: I see. I think the naming is kind of confusing to me. It seemed _prescriptive_ (aka "you should upload recordings with this status") rather than _descriptive_ (aka "I already uploaded recordings with this status"), maybe because I read "upload statuses" instead of "upload**ed** statuses" but I think that might be an easy mistake to make. Maybe "did upload statuses" would be a tiny bit clearer?

packages/test-utils/src/reporter.ts
Andarist: I have rewritten it based on ur suggestion. I think it's much cleaner now but the end result isn't exactly the same as your snippet, could u take another look?

packages/test-utils/src/reporter.ts
```diff
@@ -920,52 +920,106 @@ class ReplayReporter<TRecordingMetadata extends UnstructuredMetadata = Unstructu
...
+      }
+      case "flaky": {
+        if (this._uploadStatusThreshold === "failed") {
+          break;
+        }
+        if (!this._minimizeUploads) {
+          result.didUploadStatuses.failed ||= executions.some(r => r.result !== "passed");
+          result.didUploadStatuses.passed ||= executions.some(r => r.result === "passed");
+
+          // currently previously completed execution groups that could be entirely passed or failed are not retroactively uploaded here
```
Andarist, line 980: I don't think it's particularly important to "fix" this right now but I could certainly change my mind. 

If we have such execution groups, we detect flake when the third group completes:
```
[passed]
[passed]
[failed, passed]
```

We could look for all completed groups and upload them but it doesn't feel overly important to me.

packages/test-utils/src/reporter.ts
```diff
@@ -920,52 +920,108 @@ class ReplayReporter<TRecordingMetadata extends UnstructuredMetadata = Unstructu
...
+        if (!this._minimizeUploads) {
+          result.didUploadStatuses.failed = true;
+          toUpload.push(...executions);
+          break;
+        }
+        if (result.didUploadStatuses.failed) {
+          break;
+        }
+        result.didUploadStatuses.failed = true;
+        toUpload.push(latestExecution);
```
bvaughn, line 969: I think we need to check if status threshold is "none" here before uploading?

```suggestion
        if (this._uploadStatusThreshold !== "none") {
          if (this._minimizeUploads) {
            if (!result.didUploadStatuses.failed) {
              result.didUploadStatuses.failed = true;
              toUpload.push(latestExecution);
            }
          } else {
            result.didUploadStatuses.failed = true;
            toUpload.push(...executions);
          }
        }
```

TBH I guess we could just return early from this function if that's the status because the "flaky" branch below also needs to respect that condition.

packages/test-utils/src/reporter.ts
```diff
@@ -920,52 +920,108 @@ class ReplayReporter<TRecordingMetadata extends UnstructuredMetadata = Unstructu
...
-        if (latestResult.result !== "passed" || results.length > 1) {
-          if (!this._minimizeUploads) {
-            toUpload = results;
-          } else if (results.length > 1) {
-            toUpload = [results[0], latestResult];
-          } else {
-            toUpload = [latestResult];
+      }
+      case "flaky": {
+        if (this._uploadStatusThreshold === "failed") {
```
bvaughn, line 973: Unless we return early from this function I think we also need to check for "none" here too

packages/test-utils/src/reporter.ts
```diff
@@ -920,52 +920,108 @@ class ReplayReporter<TRecordingMetadata extends UnstructuredMetadata = Unstructu
...
+        }
+
+        if (!result.didUploadStatuses.passed) {
+          const passedExecution = Object.values(result.executions)
+            .flatMap(e => e)
+            .find(r => r.result === "passed");
+
+          if (passedExecution) {
+            result.didUploadStatuses.failed = true;
+            toUpload.push(passedExecution);
```
bvaughn, line 1004: The way this code is organized is a little hard for me to read but I think it looks correct.

packages/test-utils/src/reporter.ts
```diff
@@ -920,52 +920,108 @@ class ReplayReporter<TRecordingMetadata extends UnstructuredMetadata = Unstructu
...
+          result.didUploadStatuses.passed = true;
+          toUpload.push(...executions);
+          break;
+        }
+        if (result.didUploadStatuses.passed) {
+          break;
         }
+        result.didUploadStatuses.passed = true;
+        toUpload.push(latestExecution);
         break;
```
bvaughn, line 1023: I find interleaved `break` statements a little harder to read probably because I don't see them used this way often.

bvaughn (approved):
I think this needs to handle the "none" case as mentioned

Andarist:
> I think this needs to handle the "none" case as mentioned

We never get to those functions with this threshold as there is an early return here:
https://github.com/replayio/replay-cli/blob/be5d659af281002aabe05d7c5532f15cda941a89/packages/test-utils/src/reporter.ts#L912-L914

bvaughn:
> > I think this needs to handle the "none" case as mentioned
> 
> We never get to those functions with this threshold as there is an early return here:
> 
> https://github.com/replayio/replay-cli/blob/be5d659af281002aabe05d7c5532f15cda941a89/packages/test-utils/src/reporter.ts#L912-L914

I don't think that's a strong enough guard. This method could be called from other places (even if it isn't currently) so I think it needs its own guard. (This is a lesson I've learned for myself over the years, not to depend on indirect short-circuits for things like this.)

PR #537, opened 2024-06-17
Fix cypress assertions getting lost

hbenl:
After testing this for a while I am fairly confident that this is the right logic (although with Cypress you never know...).

### DIFF
--- a/packages/cypress/src/support.ts
+++ b/packages/cypress/src/support.ts
@@ -574,7 +574,7 @@ function register() {
 
       // if we failed to find an assertion command but this log is for an
       // assert, it's a chai assertion so we'll emit the command:start now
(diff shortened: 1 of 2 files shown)


packages/cypress/src/support.ts
```diff
@@ -574,7 +574,7 @@ function register() {
 
       // if we failed to find an assertion command but this log is for an
       // assert, it's a chai assertion so we'll emit the command:start now
-      if (!maybeCurrentAssertion && log.name === "assert") {
+      if (maybeCurrentAssertion?.get("type") !== "assertion") {
```
hbenl, line 577: - this will also add the assertion if `maybeCurrentAssertion` is set but is not an assertion itself
- we don't need to check `log.name === "assert"` here because we would have `return`ed earlier if it wasn't

PR #540, opened 2024-06-19
Avoid reporting duplicated API errors

Andarist:
**Before**
```
[replay.io]: Encountered unexpected errors while processing replays
[replay.io]:   Error: GraphQL request for CreateTestRunShard failed (Unknown API key)
[replay.io]:   Error: GraphQL request for CreateTestRunShard failed (Unknown API key)
[replay.io]:   Error: GraphQL request for CreateTestRunShard failed (Unknown API key)
```

**After**
```
[replay.io]: 🕑 Completing some outstanding work ...
[replay.io]: 
[replay.io]: ❌ We encountered some unexpected errors creating your tests on replay.io
[replay.io]:    - Unknown API key
```

### DIFF
--- a/packages/test-utils/src/reporter.ts
+++ b/packages/test-utils/src/reporter.ts
@@ -182,9 +182,15 @@ function parseRuntime(runtime?: string) {
   return ["chromium", "gecko", "node"].find(r => runtime?.includes(r));
 }
 
-function throwGraphqlErrors(operation: string, errors: any) {
+function createGraphqlError(operation: string, errors: any) {
   errors.forEach((e: any) => debug("Error from GraphQL operation %s: %o", operation, e));
-  throw new Error(
+  for (const error of errors) {
+    switch (error.extensions?.code) {
+      case "UNAUTHENTICATED":
+        return new Error(error.message);
+    }
+  }
+  return new Error(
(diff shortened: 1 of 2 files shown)


packages/test-utils/src/reporter.ts
```diff
@@ -530,16 +539,15 @@ class ReplayReporter<TRecordingMetadata extends UnstructuredMetadata = Unstructu
...
     let testRunShardId = this._testRunShardId;
     if (!testRunShardId) {
       await this._testRunShardIdPromise;
       testRunShardId = this._testRunShardId;
       if (!testRunShardId) {
-        return {
-          type: "test-run-tests",
-          error: new Error("Unable to add tests to test run: ID not set"),
-        };
+        return;
```
Andarist, line 550: returning `undefined` here helps us to avoid the error propagation and a situation like this:
```
[replay.io]: 🕑 Completing some outstanding work ...
[replay.io]: 
[replay.io]: ❌ We encountered some unexpected errors creating your tests on replay.io
[replay.io]:    - GraphQL request for CreateTestRunShard failed (Unknown API key)
[replay.io]:    - Unable to complete test run: ID not set
```

If the test run shard couldn't be created we only need to report that error. Everything that fails because of that is just its consequence and noise

PR #542, opened 2024-06-20
Setup more robust `@replayio/playwright` tests

Andarist:


packages/playwright/tests/fixtures-app/app/page.tsx
Andarist, line 1: I probably should simplify this example page - it's just taken straight from `create-next-app`

packages/playwright/tests/fixtures-app/app/_pw-utils/network-mock-global-setup.ts
Andarist, line 1: This is the core of the mocking logic. We can continue to refine this further. Right now, the test doesn't have control over WS "responses" (nor over regular requests). So It can't, for example, error any of those.

I'm looking for inspiration right now in Cypress, MSW, and Playwright. I think we could establish some conventions for aliases (like `@recording[1]` and more). We could set up a queue of "route" mocks where each of the mocks would mock the respective request once. I'd really like to find a way to avoid those be sensitive to the order and timing.

packages/playwright/tests/fixtures-app/app/_pw-utils/network-mock-global-setup.ts
```diff
@@ -0,0 +1,121 @@
...
+  beginRecordingUploadParameters,
+  beginRecordingUploadResult,
+  endRecordingUploadResult,
+  existsResult,
+  setAccessTokenResult,
+  setRecordingMetadataResult,
+} from "@replayio/protocol";
+import { WebSocket } from "undici";
+
+async function globalSetup(_config: FullConfig) {
```
Andarist, line 14: I think it will be best to pass a function through this config, any given test could pass a testing function in `.use` within its own `playwright.config.mts` and we could call this function from here with the appropriate context object. 

This way each test could keep its test and mocking body close to itself (colocation, yay!). 

This might require us to stringify the said function but I think it's fine. It's a standard constraint when dealing with `page.evaluate` and similar APIs as those pass "functions" (in a form of string) across runtime boundaries

packages/playwright/tests/index.test.ts
Andarist, line 1: This reads the list of Next.js routes to create a dynamic list of tests. In each page directory, we can create a playwright config, a playwright test, a custom page, and whatever else we might need to execute that test. Each route defined there should be a self-contained "unit". Each one of them gets executed as a separate playwright run (in an e2e~ fashion).

I'd likely have to introduce support for some `pw-test-config.js` or similar. We might want to know **here** some things about any particular test case, such as its expected status.

I'm also thinking about passing some information from a test to this runner here - like maybe some failed assertions? Jest could create better test failure displays based on that than what we can currently get if we'd just throw within failing tests

PR #549, opened 2024-06-21
Setup Rollup-based build system for packages

Andarist:


bvaughn:
@SocketSecurity ignore npm/esbuild@0.21.5

packages/replayio/src/utils/authentication/config.ts
```diff
@@ -1,4 +1,4 @@
-import { getReplayPath } from "../getReplayPath";
+import { getReplayPath } from "@replay-cli/shared/getReplayPath";
```
Andarist, line 1: This is the only thing I already started to inline as part of this PR. You can inspect the `replayio`'s output to see how this looks like. In particular this file gets transpiled to:
```js
'use strict';

var getReplayPath = require('../../shared/dist/getReplayPath.js');

const authClientId = process.env.REPLAY_AUTH_CLIENT_ID || "4FvFnJJW4XlnUyrXQF8zOLw6vNAH1MAo";
const authHost = process.env.REPLAY_AUTH_HOST || "webreplay.us.auth0.com";
const cachedAuthPath = getReplayPath.getReplayPath("profile", "auth.json");

exports.authClientId = authClientId;
exports.authHost = authHost;
exports.cachedAuthPath = cachedAuthPath;
```

.github/workflows/playwright-tests.yml
```diff
@@ -20,7 +20,7 @@ jobs:
           node-version: "18"
           cache: "yarn"
       - run: yarn --immutable
-      - run: yarn run bootstrap
+      - run: yarn run build
```
bvaughn, line 23: All of these "bootstrap" -> "build" changes are fine, but it looks like both just alias `pkg-build` anyway so are the changes necessary?

Andarist:
I published this on npm under `@pr549` tag

scripts/tsconfig/base.json
```diff
@@ -7,14 +7,16 @@
...
     "isolatedModules": true,
     "strict": true,
     "esModuleInterop": true,
     "skipLibCheck": true,
     "forceConsistentCasingInFileNames": true,
     "useDefineForClassFields": true,
     "rootDir": "${configDir}/src",
-    "outDir": "${configDir}/dist"
+    "outDir": "${configDir}/dist",
+    // TS reports TS5074 without this explicit setting
```
bvaughn, line 19: This does not look like valid JSON

scripts/pkg-build/src/bin.ts
```diff
@@ -0,0 +1,224 @@
...
+
+    successes.forEach(({ pkg }) => {
+      console.log(`${statusSuccess("✔")} Built ${pkg}`);
+    });
+
+    if (successes.length !== results.length) {
+      const failures = results.filter(r => r.status === "rejected");
+
+      failures.forEach(({ pkg, reason }) => {
+        console.log(`${statusFailed("✘")} Failed to build ${pkg}:\n`, reason);
```
bvaughn, line 217: Nit?
```suggestion
        console.error(`${statusFailed("✘")} Failed to build ${pkg}:\n`, reason);
```

bvaughn (approved):
The release-pr comments are all 👎🏼 Is this expected?

Andarist:
> The release-pr comments are all 👎🏼 Is this expected?

The first one failed because this PR removed `bootstrap` script (in favor of just `build`). The workflow is executed from `main` so this created a mismatch between the main branch and this PR that prevented the workflow from being executed correctly. 

I added `bootstrap` as an alias here and tried again. It failed this time because this PR wasn't in sync with main and it lacked this fix: https://github.com/replayio/replay-cli/pull/558

Then I "gave up" because `yarn changeset version --snapshot` is still driven by... changesets. This PR has no changesets and I didn't feel like adding any just yet (or temporarily). It would be handy to have `yarn changeset version --snapshot --force` to release *everything* regardless of the changesets state.

This PR doesn't really introduce any user-facing change so I think it might be best to skip adding changesets here altogether. We could add some just to trigger the release but I'm not sure what to mention in the CHANGELOG. Adding "Internal changes" feel silly but that's kinda what this is so maybe that would be the most appropriate?

scripts/tsconfig/base.json
Andarist: Because it's not :trollface: TS is able to parse this fine though (they allow comments and trailing commas but I'm not 100% if they are using custom parsing rules or if they are compliant with json5 or jsonc)

.github/workflows/playwright-tests.yml
Andarist: they are not, I just added `bootstrap` as an alias when trying to release this - I mentioned the reason in the comment [here](https://github.com/replayio/replay-cli/pull/549#issuecomment-2189484418)

bvaughn:
> Then I "gave up"

dr-house-relatable

PR #553, opened 2024-06-24
Use modern `moduleResolution` setting

Andarist:
I want to make use of this in https://github.com/replayio/replay-cli/pull/549


.yarn/patches/superstruct-npm-1.0.4-44d328b887.patch
```diff
@@ -0,0 +1,12 @@
+diff --git a/package.json b/package.json
+index 6f759661ef1f6ad76ed41b150b63a922b00e11b3..9928998719f56bb23d1b0603ee45ba7e737e1604 100644
+--- a/package.json
++++ b/package.json
+@@ -4,7 +4,6 @@
+   "version": "1.0.4",
+   "license": "MIT",
+   "repository": "git://github.com/ianstormtaylor/superstruct.git",
+-  "type": "module",
```
Andarist, line 9: It kinda doesn't matter for this package when it comes to the runtime code. They are using `.cjs` and `.mjs` extensions so this becomes irrelevant. However, `.d.ts` is a type-level `.js` and for that it *does* matter. 

This package doesn't ship its types correctly and this was the easiest way to fix that for us without changing the runtime semantics at all.

bvaughn (approved):

PR #558, opened 2024-06-25
Fixed PR reference in a changeset file

Andarist:


.changeset/afraid-rocks-design.md
```diff
@@ -2,7 +2,7 @@
 "@replayio/test-utils": patch
 ---
 
-pr: #647
```
Andarist, line 5: we didn't yet reach such numbers here ;p not sure where it came from 😉

bvaughn: Prob PRO-647

bvaughn (approved):

PR #560, opened 2024-06-25
Move logger to shared

Andarist:
- Move shared "logger" class from `test-utils` to the new `shared` package
- Move shared GraphQL util methods from `test-utils` to the new `shared` package
- Delete `observability-node` package


bvaughn (approved):
Looks good after a quick skim

PR #562, opened 2024-06-25
Move legacy @replayio/replay Node APIs to new shared package

bvaughn:
- [x] Remove Node entrypoint from `replayio` package
  - This should technically require a new major version but since there was only a single, useless API exported up until this point– I think we can safely assume that nothing was using it
- [x] Update cypress, playwright, puppeteer, and test-utils packages to depend on the new `@replay-cli/shared` package
- [x] Moved some of the code from the legacy `@replayio/replay` package to `shared` or `test-utils` packages
  - At this point the only package still depending on the legacy `@replayio/replay` package (Node APIs) is `test-utils` which will require follow up work because some of the code is harder to extricate

As of this PR the following dependencies have been updated:
* `@replayio/cypress`
  * ✅  `@replay-cli/shared`
  * ❌ ~~`@replayio/replay`~~
* `@replayio/playwright`
  * ✅  `@replay-cli/shared`
  * ❌ ~~`@replayio/replay`~~
* `@replayio/puppeteer`
  * ✅  `@replay-cli/shared`
  * ❌ ~~`@replayio/replay`~~

### Next steps
 
The major remaining work will be to update `@replayio/test-utils` to no longer depend on `@replayio/replay` (so that we can delete that package entirely)


packages/shared/src/runtime/logger.ts
```diff
@@ -0,0 +1,3 @@
+import { Logger } from "@replay-cli/shared/logger/logger";
```
bvaughn, line 1: If this tries to import from "../logger" the build fails:
```ts
import { Logger } from "../logger";
```

Maybe bc of the same file name? Seems like that should work (and it's what Code auto-completes)

bvaughn:
`yarn build` works for me but `yarn install` now hangs with:
```
➜ replay-cli (PRO-720) ✗ yarn install
yarn build➤ YN0000: · Yarn 4.1.1
➤ YN0000: ┌ Resolution step
➤ YN0000: └ Completed
➤ YN0000: ┌ Post-resolution validation
➤ YN0060: │ typescript is listed by your project with version 5.4.5, which doesn't satisfy what react-scripts (p6e159) and other dependencies request (^3.2.1 || ^4.0.0 || ^3.2.1 || ^4.0.0 || ^3.3.0-dev || ^4.0.0 || ^3.4.0-dev || ^4.0.0 || ^3.5.0-dev || ^4.0.0 || ^3.6.0-dev || ^4.0.0 || ^3.6.0-beta || ^4.0.0 || ^3.7.0-dev || ^4.0.0 || ^3.7.0-beta || ^4.0.0).
➤ YN0002: │ @replay-cli/pkg-build@workspace:scripts/pkg-build doesn't provide typescript (pcd618), requested by rollup-plugin-dts.
➤ YN0002: │ create-react-app-typescript@workspace:examples/create-react-app-typescript doesn't provide @testing-library/dom (p1ae55), requested by @testing-library/user-event.
➤ YN0086: │ Some peer dependencies are incorrectly met; run yarn explain peer-requirements <hash> for details, where <hash> is the six-letter p-prefixed code.
➤ YN0000: └ Completed
➤ YN0000: ┌ Fetch step
➤ YN0000: └ Completed in 0s 328ms
➤ YN0000: ┌ Link step
➤ YN0007: │ esbuild@npm:0.21.5 must be built because it never has been before or the last one failed
➤ YN0007: │ cypress@npm:13.11.0 must be built because it never has been before or the last one failed
➤ YN0007: │ core-js-pure@npm:3.36.1 must be built because it never has been before or the last one failed
➤ YN0007: │ protobufjs@npm:7.3.0 must be built because it never has been before or the last one failed
➤ YN0007: │ core-js@npm:3.36.1 must be built because it never has been before or the last one failed
➤ YN0007: │ @replayio/puppeteer@workspace:packages/puppeteer must be built because it never has been before or the last one failed
```

packages/replayio/package.json
```diff
@@ -4,7 +4,6 @@
   "description": "CLI tool for uploading and managing recordings",
   "bin": "./bin.js",
   "exports": {
-    ".": "./dist/index.js",
```
Andarist, line 7: do u plan to release v2 of this package? or do u consider this entrypoint to be somewhat internal and we'll skip that?

packages/cypress/package.json
```diff
@@ -43,14 +44,17 @@
     "typescript": "^5.5.2"
   },
   "dependencies": {
-    "@replayio/replay": "workspace:^",
     "@replayio/test-utils": "workspace:^",
     "chalk": "^4.1.2",
     "debug": "^4.3.4",
+    "fs-extra": "^11.2.0",
```
Andarist, line 50: It's not a new thing here but it would be great to get rid of this dependency. I feel like it adds too much to install size for what we are actually using it for

packages/replayio/package.json
bvaughn: I am prepared to do v2 but I could be convinced that no one is using this API and we can cheat semver a little.

Andarist: Do u remember why this was added? Who could be using this now?

bvaughn: I don't think anyone is using it.

packages/cypress/package.json
bvaughn: Quick audit of the APIs we use from this package:
* `appendFileSync`
* `createReadStream`
* `ensureDirSync`
* `ensureFileSync`
* `existsSync`
* `readdir`
* `readdirSync`
* `readFile`
* `readFileSync`
* `ReadStream`
* `removeSync`
* `renameSync`
* `rmSync`
* `stat`
* `unlinkSync`
* `writeFileSync`

I don't think removing it is worth spending time on. Saving `8kB` (minified) is nothing for developer' machines or CI boxes.

bvaughn:
I think I will land this soon (after chatting with Adi here in a minute) to unblock Adi and Miriam's work. I can continue iterating on the `test-utils` cleanup in a follow-up PR.

callingmedic911 (approved):

PR #564, opened 2024-06-25
Change logger to singleton

callingmedic911:
Closes PRO-722.
This PR moves the logger to singleton, we can keep all tags centrally. Will be needed for the sessionId change I'm doing.


miriambudayr (approved):

packages/shared/src/logger/logger.ts
```diff
@@ -91,4 +91,17 @@ class Logger {
...
 }
 
-export { Logger };
+let logger: Logger;
+
+// This should be called with the name once at the entry point.
+// For example, with the Playwright plugin, it is called in the Reporter interface constructor.
+function initLogger(name: string) {
+  if (logger) {
+    console.warn(`Logger already initialized.`);
```
miriambudayr, line 100: See comment:

https://github.com/replayio/replay-cli/pull/565/files#r1655248548

packages/shared/src/logger/logger.ts
```diff
@@ -71,6 +72,7 @@ class Logger {
       return;
     }
 
+    this.grafana.logger.close();
```
callingmedic911, line 75: @miriambudayr I had to add, otherwise, the process wasn't exiting `packages/puppeteer/src/first-run.ts`.

packages/shared/src/logger/logger.ts
```diff
@@ -37,6 +37,7 @@ class Logger {
       basicAuth: GRAFANA_BASIC_AUTH,
       format: winston.format.json(),
       replaceTimestamp: true,
+      timeout: 5000,
```
callingmedic911, line 40: The default was 30s. I dialed it down a bit. 30s is extremely slow for the user app.

miriambudayr: Totally agree. I must have set it to a really high value during development and didn't add a comment reminding myself to make the default more reasonable.

packages/shared/src/logger/logger.ts
miriambudayr: Huh, interesting. So basically the CLI process would hang forever?

miriambudayr (approved):

bvaughn:
Would be nice to get this landed this morning. [I'd like to make a small change](https://discord.com/channels/779097926135054346/1254952152250450022/1255878165629243445) to the `Logger` API to simplify the rest of the work I'm trying to do with PRO-720.

packages/shared/src/logger/logger.ts
bvaughn: @Andarist also noticed this -> #570

I'm going to merge and resolve conflicts to unblock this branch from being merged :)

bvaughn: Noticed that you're calling `close` before `flush` here though. Is that intentional?

If it's important to call both, it seems like flushing _before_ closing would be the expected order. (That's the order @Andarist chose too.)

Andarist: > Noticed that you're calling close before flush here though. Is that intentional?

I specifically chose to call flush before close so we could await the flush. Close also flushes things but it's using a "detached" promise to do that (it's not returned):
https://github.com/JaniAnttonen/winston-loki/blob/2ed0954e97d85e4be1b81f8d853f1864fa37cf6c/src/batcher.js#L304

bvaughn: Nice! Good clarification. (Maybe we should add an inline comment with a link to the docs?)

PR #565, opened 2024-06-25
Add device and session ID to logger

callingmedic911:
Closes PRO-644.

- Add deviceId that is persistent across different test runs.
- `sessionId` to filter logs based on a single run.


packages/shared/src/logger/logger.ts
```diff
@@ -91,4 +104,15 @@ class Logger {
   }
 }
 
-export { Logger };
+let logger: Logger;
+
+// This should be called with the name once at the entry point.
+// For example, with the Playwright plugin, it is called in the Reporter interface constructor.
+function initLogger(name: string) {
```
bvaughn, line 111: We should probably warn if it's called more than once?

miriambudayr: I would prefer to error loudly rather than warn. Otherwise developers will likely not realize it when they introduce multiple calls to `initLogger`, and logging is a critical piece of our infrastructure for customer support.

Thoughts? @bvaughn

callingmedic911: I chose to use a warning because if we accidentally ship two `inits`, I don’t want to break userland functionality. For example, if the second `init` is in a code path a developer hasn’t tested, we wouldn’t see an error or warning at all. It would break in userland.

miriambudayr: It's fine to keep it as-is for now, in my opinion, until #562.

miriambudayr (approved):

callingmedic911:
@bvaughn does the commit [`4afdf4a` (#565)](https://github.com/replayio/replay-cli/pull/565/commits/4afdf4aae6ff104eef7934e239acaa848560253c) look right to you? I created `getObservabilityCachePath` because it's being used at two places.

bvaughn:
I think that looks ok

PR #567, opened 2024-06-26
Make build script to be executed per pkg and introduce a watch script

Andarist:
This PR introduced turborepo as our task runner. It figures out a dependency graph on its own and is meant to re-run only the tasks that have to be re-executed since it relies on caching heavily*.

This way we can (again) run builds per package (the previous setup had to be executed in the root and had to build everything). We also gain the ability to run those build tasks on watch.

*I can't figure out why just any change busts the cache right now.


packages/replay/package.json
```diff
@@ -20,7 +20,7 @@
   ],
   "scripts": {
     "prepare": "yarn run build",
-    "build": "rm -rf dist/ metadata/ tsconfig.tsbuildinfo && tsc && node ./scripts/mirror-dist-metadata.js",
+    "build": "pkg-build && node ./scripts/mirror-dist-metadata.js",
```
Andarist, line 23: The previous setup forgot to run `node ./scripts/mirror-dist-metadata.js` here 😢 It was the only `build` script that had some extra requirements.

This broke our upload action since it requires files manually instead of relying on node's resolution:
https://github.com/replayio/action-upload/blob/715f00955415fcc6873ce268396489a2ad07dca2/action.yml#L48-L50

bvaughn (approved):
Accept to unblock but didn't review closely

PR #569, opened 2024-06-26
Fixed turborepo caching on `replayio` changes

Andarist:


package.json
```diff
@@ -27,12 +27,9 @@
   "devDependencies": {
     "@changesets/changelog-github": "^0.5.0",
     "@changesets/cli": "^2.27.5",
-    "@replay-cli/pkg-build": "workspace:^",
     "@replayio/protocol": "^0.73.0",
-    "@replayio/replay": "workspace:^",
     "@types/ws": "^8",
     "prettier": "^2.7.1",
-    "replayio": "workspace:^",
```
Andarist, line 35: It turns out that dependencies of the root are treated like global dependencies, they invalidate everything. So before this PR each change to `replayio`'s source code caused *every* other package to be invalidated because the root depended on it.

bvaughn (approved):
INteresting

Andarist:
> CI seems unhappy here though:

I canceled it yesterday manually because it was hanging. I think this was resolved by https://github.com/replayio/replay-cli/pull/570 . Now I synced this with main to benefit from that fix so let's see if that truly resolved that

bvaughn:
Noice

PR #570, opened 2024-06-27
Close the logger properly

Andarist:
Without this, the process can't be closed because `winston-loki` has an infinite interval loop


bvaughn (approved):

bvaughn:
FYI @miriambudayr, @callingmedic911

PR #571, opened 2024-06-27
Moving more code into shared

bvaughn:
Continuation of [PRO-720](https://linear.app/replay/issue/PRO-720); follows up to #562

Moving much of the `replayio` utility-type code into the new `shared` package so that we can use those Node APIs for querying the list of recording and uploading or deleting them.

Most of this was rote but I did make a couple of changes worth calling out explicitly:
- I changed the value returned by `fetchAuthIdsFromGraphQL` from `{ userId: string | null; workspaceId: string | null }` to `{ id: string; type: "user" | "workspace" }` and updated the callers; this was easier to work with (from a types perspective)
  - Grafana still gets `userId` _or_ `workspaceId` though, just not both
- `Logger` is now eagerly instantiated; the `initialize()` method will accept a name and version and pass them along to the Grafana instance. Logging before initialization (or in Jest tests) will be local-only.
  - Note that a version parameter is also supported now (so we can associate Grafana logs with specific releases)
  - Both name and version should probably be imported and passed directly from `package.json` (which is what I am currently doing for Puppeteer and replayio)
  - Note that I've replaced the previous "replayio" name prefix with whatever the passed package name is
- Lots of `debug(...args)` logging has been changed to `logger.debug("...", {...args})`
- I've moved `mixpanel` and `launch-darkly` code over to `shared` too, even though (for now) `replayio` is the only thing using this stuff
- I also tidied up the `getRecordings` method to not create so many throw-away arrays; (I didn't just do this for performance, but TypeScript was complaining about the return array containing recordings _and_ `undefined` with the previous code structure)

## Next steps
- Remove the `test-utils` dependency on `@replayio/replay` (and delete `@replayio/replay`)
- Update `test-utils` to be a peer dependency so that it will get inlined as well (to resolve ambiguous logger ownership)
- Audit `logger.debug` entries and decide which ones should be `info`, `log`, or even `error`
- Address any added `TODO [PRO-720]` comments


bvaughn:
What is going on with CI:
```
 trackEvent › authenticated › should track pending promises until resolved or rejected

               +test |     TypeError: Cannot read properties of undefined (reading 'debug')

               +test |       15 |     eventName = `replayio.${eventName}`;
               +test |       16 |   }
               +test |     > 17 |   logger.logger.debug(`trackEvent: "${eventName}"`, { properties });
               +test |          |                 ^
               +test |       18 |   trackEventImplementation(mixpanelAPI, eventName, properties);
               +test |       19 | }
               +test |       20 | async function trackEventImplementation(mixpanelAPI, eventName, properties) {

```
This error _would make sense_ if the code had `logger.logger.debug` but it does not:

https://github.com/replayio/replay-cli/blob/642e5b993302964790f2a00814d1f9676d02063b/packages/shared/src/mixpanel/trackEvent.ts#L18

packages/shared/package.json
```diff
@@ -3,7 +3,6 @@
   "version": "0.0.0",
   "private": true,
   "exports": {
-    "./logger": "./dist/logger/index.js",
```
bvaughn, line 6: Ooof I didn't know about (or forgot about) this one

callingmedic911 (approved):
lgtm 
image

packages/shared/package.json
callingmedic911: Just checking, anything to do here?

bvaughn: This was intentional. It fixed the weird `logger.logger.debug` change. My comment was to Mateusz; I just meant that I'd forgot it existed so I didn't think to try deleting it.

PR #572, opened 2024-06-28
Add "whoami" command

bvaughn:
- [x] Merge https://github.com/replayio/docs/pull/211 to add the new API keys redirect page

### Screenshots
| Scenario | |
| :--- | :--- |
| Not authenticated | Not authenticated |
| Authenticated by email and password | Authenticated by email and password
| Team API key | Team API key
| Personal API key | Personal API key

### DIFF
--- a/.changeset/pink-lemons-bathe.md
+++ b/.changeset/pink-lemons-bathe.md
@@ -1,0 +1,5 @@
+---
+"replayio": patch
+---
+
+Add "whoami" command to print information about the current user and API key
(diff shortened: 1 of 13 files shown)


callingmedic911 (approved):
Looks great! Do you think it's worth printing email address associated with user account (if it's user authenticated)? This command reminds me fly.io `fly auth whoami`, which I frequently use to see which account I'm authenticated with because I use fly for both work and personal usage.
Screenshot 2024-06-28 at 9 16 14 AM

bvaughn:
Thanks!

I thought about printing the email or team name, but I didn't because I thought it would require an extra GraphQL query. Thinking about it again though, it wouldn't really.

We _could_ just pull down that info along with the `AuthInfo` query couldn't we? Except I don't think the `viewer.user` relationship has an `email` field, only a `name`.

bvaughn:
Okay. That was great suggestion. This command is way better now. Thanks!

PR #573, opened 2024-06-28
Log broken entries when reading the recordings.log file

hbenl:


packages/replay/src/recordingLog.ts
```diff
@@ -53,6 +53,9 @@ export function readRecordings(dir?: string, includeHidden = false) {
       try {
         return JSON.parse(line) as RecordingLogEntry;
       } catch {
+        if (line) {
+          debug('Failed to parse recordings.log entry: "%s"', line);
+        }
```
bvaughn, line 58: I was going to say "we do this already" but I guess that's the new CLI that does this.

Should we port this fix to the legacy CLI code as well?

https://github.com/replayio/replay-cli/blob/33656ea4a263e3c09a81b23576f00fce2456a0f3/packages/shared/src/recording/readRecordingLog.ts#L44-L67

bvaughn (approved):

packages/replay/src/recordingLog.ts
hbenl: > Should we port this fix to the legacy CLI code as well?

I don't understand what you're asking for - this change *is* in the legacy CLI code, no?

bvaughn: I'm asking if we should copy the fix in `replayio`'s read-recordings-log method to the legacy CLI as well. I don't know how likely it is, but it seems like if an older version of `replayio` and Cypress/Playwright were used on the same machine, the recordings log might be broken in a way that the legacy CLI would fail to parse.

hbenl: Got it. I think it's unlikely that we'll run into this problem since it only affected the first recording made after the old version of `replayio` removed the trailing `\n` and `replayio` has been fixed for a while now.

PR #574, opened 2024-06-28
Migrate logs in `playwright` package

callingmedic911:
- Sets up logger initialization a level up in ReplayPlaywrightReporter.
- Left initLogger in `ReplayReporter`. Once we have test-utils inlined we'd remove logger init and close from `ReplayReporter` in test-utils. The side effect right now is we'll have two different session IDs between these packages.

Closes PRO-701.

### DIFF
--- a/packages/playwright/src/reporter.ts
+++ b/packages/playwright/src/reporter.ts
@@ -23,6 +23,7 @@ import { getRuntimePath } from "@replay-cli/shared/runtime/getRuntimePath";
 import { FixtureStepStart, ParsedErrorFrame, TestExecutionIdData } from "./fixture";
 import { StackFrame } from "./playwrightTypes";
 import { getServerPort, startServer } from "./server";
+import { initLogger, logger } from "@replay-cli/shared/logger";
 
 const debug = dbg("replay:playwright:reporter");
 const pluginVersion = require("@replayio/playwright/package.json").version;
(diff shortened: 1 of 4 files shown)


packages/playwright/src/reporter.ts
```diff
@@ -81,6 +82,9 @@ class ReplayPlaywrightReporter implements Reporter {
   private _foundReplayBrowser = false;
 
   constructor(config: ReplayPlaywrightConfig) {
+    const runnerName = "test_adi";
```
bvaughn, line 85: This looks wrong?

bvaughn (commented):
> The side effect right now is we'll have two different session IDs between these packages.

If this is problematic we could change `Logger` to only initialize Grafana once (first in iwns)

packages/playwright/src/reporter.ts
callingmedic911: Whoops yes.

packages/playwright/src/reporter.ts
```diff
@@ -90,7 +94,7 @@ class ReplayPlaywrightReporter implements Reporter {
     this.config = config;
     this.reporter = new ReplayReporter(
       {
-        name: "playwright",
```
bvaughn, line 93: Shouldn't the package name and version be imported from the `package.json` ? That way they'll stay in sync.

That's what we're doing in the other packages:

https://github.com/replayio/replay-cli/blob/db3f422416d9385db543e9d4876752a159adfa79/packages/replayio/src/bin.ts#L4

https://github.com/replayio/replay-cli/blob/db3f422416d9385db543e9d4876752a159adfa79/packages/replayio/src/bin.ts#L20

callingmedic911:
> If this is problematic we could change Logger to only initialize Grafana once (first in iwns)

Not sure what you mean. I was not sure how we can use the same instance across two packages (that are not yet inlined). We could have something in the device but I don't think multiple session IDs will be there for long.

bvaughn:
Oh. Sorry I'm assuming we'll finish the inline effort very soon and I was really only considering that scenario.

bvaughn (approved):

PR #575, opened 2024-06-28
Forked @replayio/replay Node APIs by copying them into test-utils

bvaughn:
To deprecate `@replayio/replay` we need to remove the following dependencies from `test-utils`:
* `listAllRecordings`
* `removeRecording`
* `uploadRecording`

The two options we can explore are:

1. Update `test-utils` to use the new shared variations of these methods (e.g. replace `listAllRecordings` from the legacy CLI with the new shared `getRecordings`).
   - **Cons**: The in-memory types for recordings are different– the metadata and the status field(s)– and it's unclear how much work would be required to bring them into alignment.
   - **Pros**: Once it's done, there is no further clean up work. We can delete the legacy CLI code and be done with it. (Pending the outcome of [PRO-730](https://linear.app/replay/issue/PRO-730/take-care-of-github-actions-after-nuking-the-old-cli) possibly.)
1. Fork those three methods from `@replayio/replay` into `test-utils` and leave the cleanup for later.
   - **Cons**: Those methods refer to a lot of other methods and it would probably end up being a pretty big amount of stuff to fork over. Maybe we could do it as a batch and separate it into its own `test-utils/legacy` directory or something so that it's clear we need to clean it up. (This doesn't really move us any closer toward our longer-term code sharing goals in other words.)
   - **Pros**: It's probably probably faster and a little less risky in the short term, and most importantly it unblocks the observability and funnel tracking work.

After some discussion with Mateusz, this PR implements the second option above.

This PR also updates Cypress, Jest, Playwright, and Puppeteer packages so that `@replayio/test-utils` is a _peer dependency_ (so it will be inlined).


bvaughn:
This PR isn't running GitHub actions for some reason. They ran on the first commit and then stopped.

bvaughn:
```
============================ ❌ FAILURE [2. Build 🔧] ============================
Repeating the output of the command that caused the failure
               +lint *failed* | --> RUN yarn run lint
               +lint *failed* | Checking formatting...
               +lint *failed* | [warn] packages/test-utils/src/legacy-cli/README.md
               +lint *failed* | [warn] Code style issues found in the above file. Forgot to run Prettier?
               +lint *failed* | Command /bin/sh -c 'yarn run lint' failed with exit code 1
               +lint *failed* |                +lint *failed* | ERROR: Command exited with non-zero code: RUN yarn run lint
```
This is an annoying thing for us to fail CI on, considering we don't provide a prettier command, and running `yarn lint` locally gives a very unhelpful output:
```
$ yarn lint
Checking formatting...
[warn] packages/test-utils/src/legacy-cli/README.md
[warn] Code style issues found in the above file. Forgot to run Prettier?
```

callingmedic911 (approved):
:shipit:

PR #576, opened 2024-06-29
Rework the dts generation

Andarist:
fixes https://linear.app/replay/issue/PRO-740


packages/jest/package.json
```diff
@@ -37,6 +37,7 @@
     "@jest/test-result": "^27.5.1",
     "@jest/types": "^27.5.1",
     "@replay-cli/pkg-build": "workspace:^",
+    "@replay-cli/shared": "workspace:^",
```
Andarist, line 40: We get a confusing error if a transitive bundled dependency isn't declared here. While we need this situation to error for regular external dependencies, I think we could compute potential bundled dependencies recursively here:
https://github.com/replayio/replay-cli/blob/229636edcd16a9a3dc11465b447763eff59c34cf/scripts/pkg-build/src/bin.ts#L29-L33

I imagine that this recursive search should include **dependencies** (unlike the top-level one above that looks at `devDependencies`)

packages/test-utils/src/reporter.ts
```diff
@@ -264,7 +262,7 @@ class ReplayReporter<TRecordingMetadata extends UnstructuredMetadata = Unstructu
     schemaVersion: string,
     config?: ReplayReporterConfig<TRecordingMetadata>
   ) {
-    setUserAgent(`${pkgJson.name}/${pkgJson.version}`);
+    // setUserAgent(`${pkgJson.name}/${pkgJson.version}`);
```
Andarist, line 265: @bvaughn https://github.com/replayio/replay-cli/pull/575 moved `test-utils` to a dev dependencies of other packages - so it's now a bundled dependency of them. We should add `setUserAgent` to all of those `test-utils` dependents, right?

bvaughn: Yeah that would make more sense now that test-utils is inlined

bvaughn: IMO we should just delete this line though rather than commenting it out

bvaughn (approved):
Stamping to unblock but didn't really read this.

PR #577, opened 2024-06-30
Add Mixpanel to Playwright

bvaughn:
- [x] Reorganized Mixpanel code to be a bit easier to reason about and added some more unit tests
  - Most of the lines changed in this PR comes from this reorganization
- [x] Add some basic Mixpanel logging to the `playwright` and `test-utils` packages
  - Playwright specific tracking:
    - Invalid reporter config
    - Replay reporter used without Replay browser
  - Test utils tracking (also applies to Cypress)
    - Reporter error
    - Missing API key
    - Recording metadata
    - Test end
    - Test suite begin and end
    - Test suite results (including number of crashed recordings and uploaded recordings)
- [x] Spot test and verify that events are ending up in Mixpanel


packages/replayio/src/utils/initialization/initialize.ts
```diff
@@ -60,10 +60,7 @@ export async function initialize({
     : Promise.resolve();
 
   const mixpanelPromise = raceWithTimeout(
-    initMixpanelForUserSession(accessToken, {
-      packageName,
-      packageVersion,
-    }),
+    mixpanelAPI.initialize({ accessToken, packageName, packageVersion }),
```
Andarist, line 63: not directly related to this PR - but since the logger initialization is grouped together with the mixpanel's initialization in our test plugins it could also make sense to move that into this `[...]/initialize.ts` file here

packages/test-utils/src/reporter.ts
```diff
@@ -448,8 +431,18 @@ class ReplayReporter<TRecordingMetadata extends UnstructuredMetadata = Unstructu
...
+      baseId: this._baseId,
+      runTitle: this._runTitle,
+      upload: this._upload,
+      hasFilter: !!this._filter,
+    });
+
     if (!this._apiKey) {
       logger.info("OnTestSuiteBegin:NoApiKey");
+
+      mixpanelAPI.trackEvent("test-suite.no-api-key");
```
Andarist, line 444: q: couldn't we use a single naming scheme for both of those "loggers"?

Andarist (approved):

packages/test-utils/src/reporter.ts
bvaughn: Yes we could but I don't know if there's value in doing that. They're tracking different things. Sometimes they'll line up (like here) but often they won't. Standardizing on names is work/effort.

PR #581, opened 2024-07-01
Migration logs in `cypress` package

callingmedic911:
Closes PRO-702.

### DIFF
--- a/packages/cypress/src/fixture.ts
+++ b/packages/cypress/src/fixture.ts
@@ -14,11 +14,11 @@ function getFixtureFile() {
 }
 
 export function initFixtureFile() {
-  logger.info("CypressReporter:InitFixtureFile", {
+  logger.info("InitFixtureFile:Started", {
     updateFixture: process.env.REPLAY_CYPRESS_UPDATE_FIXTURE,
   });
   if (process.env.REPLAY_CYPRESS_UPDATE_FIXTURE) {
-    logger.info("CypressReporter:InitFixtureFile", { fixtureFile: getFixtureFile() });
+    logger.info("InitFixtureFile:FixtureFile", { fixtureFile: getFixtureFile() });
     try {
       mkdirSync(path.dirname(getFixtureFile()), { recursive: true });
       writeFileSync(getFixtureFile(), "");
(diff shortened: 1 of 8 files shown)


packages/cypress/src/junit.ts
```diff
@@ -130,7 +131,7 @@ function appendReplayUrlsToFailureNodes(node: INode, replayUrls: string[]) {
       failure.children[0] = escapeForXml(output);
     });
   } catch (e) {
```
bvaughn, line 133: These `e` and `err` names are pet peeves of mine 😆 especially when they cause us to spend more characters renaming them

bvaughn (approved):

packages/cypress/src/fixture.ts
```diff
@@ -16,9 +14,11 @@ function getFixtureFile() {
 }
 
 export function initFixtureFile() {
-  debug("REPLAY_CYPRESS_UPDATE_FIXTURE: %s", process.env.REPLAY_CYPRESS_UPDATE_FIXTURE);
+  logger.info("CypressReporter:InitFixtureFile", {
```
miriambudayr, line 17: Won't it already be clear from passing the package name/version to the `initLogger` call that the logs are coming from the cypress package?

Usually when I format logs, I do something more like:

`InitFixtureFile:Started`

That way, it is clear which function each log comes from, and in most cases there is a verb there so it is easy to read and understand without looking at the code.

miriambudayr: Examples: `FunctionName:Failed` or`FunctionName:Started` or `FunctionName:WillGetSources`, etc.

packages/cypress/src/steps.ts
```diff
@@ -234,7 +230,7 @@ function groupStepsByTest(tests: Test[], steps: StepEvent[]): Test[] {
 
           // TODO [ryanjduffy]: Skipping handling after each events for now
           if (step.test[0] === AFTER_EACH_HOOK) {
-            debug("After each hooks are not currently supported");
+            logger.info("CypressReporter:AfterEachNotSupported");
```
miriambudayr, line 233: Here I'd probably do `ShouldSkipstep:AfterEachNotSupported`

PR #582, opened 2024-07-01
Change the log level and message for `replayio` logs

callingmedic911:
- Registers new exit task to close logger.
- Change log messages to the convention we are following at other places.
- Update log level


packages/shared/src/process/exitTask.ts
```diff
@@ -1,9 +1,10 @@
 import { close as finalizeLaunchDarkly } from "../launch-darkly/close";
 import { close as finalizeMixPanel } from "../mixpanel/close";
+import { logger } from "../logger";
 
 export type ExitTask = () => Promise<void>;
 
-export const exitTasks: ExitTask[] = [finalizeLaunchDarkly, finalizeMixPanel];
+export const exitTasks: ExitTask[] = [finalizeLaunchDarkly, finalizeMixPanel, logger.close];
```
bvaughn, line 7: I don't think this is safe to do without binding the method to the instance.

Either:
```suggestion
export const exitTasks: ExitTask[] = [finalizeLaunchDarkly, finalizeMixPanel, () => logger.close()];
```

Or we'll need to change the `close` method to be defined as an error function on the `Logger` class.

bvaughn:
Once you've fixed that we should generate a pre-release candidate for this branch so that @Domiii can run it and see if it gives us any more insights into why record is failing for him.

packages/shared/src/process/exitTask.ts
callingmedic911: Oh right, JS and `this`.

bvaughn (approved):
I defer to @miriambudayr about the specific log levels but I think this makes sense to me fwiw

callingmedic911:
@Domiii you want to give this ^ PR release a try? We'll see logs in 

https://replayioclient.grafana.net/explore?schemaVersion=1&panes=%7B%22ylr%22:%7B%22datasource%22:%22grafanacloud-logs%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22%7Bapp%3D%5C%22replayio%5C%22%7D%20%7C%3D%20%60%60%20%7C%3D%20%60%60%20%7C%20json%22,%22queryType%22:%22range%22,%22datasource%22:%7B%22type%22:%22loki%22,%22uid%22:%22grafanacloud-logs%22%7D,%22editorMode%22:%22builder%22%7D%5D,%22range%22:%7B%22from%22:%22now-5m%22,%22to%22:%22now%22%7D%7D%7D&orgId=1

PR #583, opened 2024-07-01
migrate log points to the new logger in `packages/test-utils/src/legacy-cli/main.ts`

miriambudayr:
This is one part of a series of migration changes.

I started with `packages/test-utils/src/legacy-cli/main.ts` because `packages/test-utils/src/reporter.ts` directly imports from it. Some of the most consequential functions like `listAllRecordings` and `uploadRecording` are defined in `packages/test-utils/src/legacy-cli/main.ts` as well.

### DIFF
--- a/packages/shared/src/logger.ts
+++ b/packages/shared/src/logger.ts
@@ -119,9 +119,7 @@ class Logger {
       if (value instanceof Error) {
         result = {
           ...result,
-          errorName: value.name,
-          errorMessage: value.message,
-          errorStack: value.stack,
+          [key]: { name: value.name, message: value.message, stack: value.stack },
         };
         continue;
       }
--- a/packages/shared/src/logger.ts
+++ b/packages/shared/src/logger.ts
@@ -74,7 +74,9 @@ class Logger {
   }
 
   private log(message: string, level: LogLevel, tags?: Tags) {
-    this.localDebugger(message, JSON.stringify(tags));
+    const formattedTags = this.formatTags(tags);
+
+    this.localDebugger(message, formattedTags);
 
     if (process.env.REPLAY_TELEMETRY_DISABLED) {
       return;
@@ -83,7 +85,7 @@ class Logger {
     const entry: LogEntry = {
       level,
       message,
-      tags,
+      ...formattedTags,
       deviceId: this.deviceId,
       sessionId: this.sessionId,
     };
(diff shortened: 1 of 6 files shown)


packages/test-utils/src/legacy-cli/main.ts
```diff
@@ -184,10 +194,18 @@ async function setMetadata(
...
+          logger.error("SetMetadata:WillRetry", {
+            recordingId,
+            errorMessage: getErrorMessage(e),
+          });
         }
       );
     } catch (e) {
+      logger.error("SetMetadata:Failed", {
+        recordingId,
+        errorMessage: getErrorMessage(e),
```
callingmedic911, line 206: I don’t think it’s a good idea to strip down to just the error name when logging into Grafana. The stack trace is helpful for debugging the problem.

packages/shared/src/logger.ts
```diff
@@ -106,6 +108,30 @@ class Logger {
...
+    let result: Record<string, unknown> = {};
+
+    for (const [key, value] of Object.entries(tags)) {
+      if (value instanceof Error) {
+        result = {
+          ...result,
+          errorName: value.name,
+          errorMessage: value.message,
+          errorStack: value.stack,
+        };
```
callingmedic911, line 125: This would overwrite if `result` had any `errorName`, `errorMesage`. Also if there multiple error instance (unlikely) but the last one would win. What if we do something like:
```suggestion
          ...result,
          [key]: {
            ...error // if there are any extra properties.
            errorName: value.name,
            errorMessage: value.message,
            errorStack: value.stack,
          }
        };
```
This will put everything inside the same key.

packages/shared/src/logger.ts
```diff
@@ -106,6 +108,28 @@ class Logger {
...
+          [key]: { errorName: value.name, errorMessage: value.message, errorStack: value.stack },
+        };
+        continue;
+      }
+
+      result = { ...result, [key]: value };
+    }
+
+    return result;
+  }
```
callingmedic911, line 131: nit: I avoid destructuring here. Log is used frequently, this adds overhead but probably not that much that's why nit.

Though I'd prefer to spread ...value, if in case there are more properties attached. Like if we extend an Error object and more properties in that.
```suggestion
  private formatTags(tags?: Record<string, unknown>) {
    if (!tags) {
      return;
    }
  
    return Object.entries(tags).reduce((result, [key, value]) => {
      if (value instanceof Error) {
        result[key] = {
          ...(value as Record<string, unknown>), // Intentionally keeping this for any extra properties attached in Error
          errorName: value.name,
          errorMessage: value.message,
          errorStack: value.stack,
        };
      } else {
        result[key] = value;
      }
      return result;
    }, {} as Record<string, unknown>);
  }
```

callingmedic911 (approved):

callingmedic911 (approved):

PR #584, opened 2024-07-02
Change formatting for browser warning

bvaughn:
If you have not installed the Replay browser we will show the following message:
Screenshot 2024-07-01 at 12 23 12 PM

If you have already installed the Replay browser, we will show:
Screenshot 2024-07-01 at 12 25 43 PM

### DIFF
--- a/packages/playwright/src/reporter.ts
+++ b/packages/playwright/src/reporter.ts
@@ -5,26 +5,27 @@ import type {
   TestError,
   TestResult,
 } from "@playwright/test/reporter";
+import { emphasize, highlight, link } from "@replay-cli/shared/theme";
 import {
-  getMetadataFilePath as getMetadataFilePathBase,
-  removeAnsiCodes,
   ReplayReporter,
   ReplayReporterConfig,
   TestMetadataV2,
+  getMetadataFilePath as getMetadataFilePathBase,
+  removeAnsiCodes,
 } from "@replayio/test-utils";
-import { readFileSync } from "fs";
+import { existsSync, readFileSync } from "fs";
 import path from "path";
 import { WebSocketServer } from "ws";
 
 type UserActionEvent = TestMetadataV2.UserActionEvent;
 
+import { initLogger, logger } from "@replay-cli/shared/logger";
 import { getRuntimePath } from "@replay-cli/shared/runtime/getRuntimePath";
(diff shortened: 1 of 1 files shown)


packages/playwright/src/reporter.ts
```diff
@@ -328,9 +329,19 @@ class ReplayPlaywrightReporter implements Reporter {
     try {
       await this.reporter.onEnd();
       if (!this._foundReplayBrowser) {
```
Andarist, line 331: wouldn't this prevent the installation warning from being logged when projects are configured correctly but our browser is not installed?

Andarist (approved):

packages/playwright/src/reporter.ts
bvaughn: I don't understand what you mean.

I didn't change the naming of `_foundReplayBrowser` but the way it was meant to be used was to flag whether or not Replay Chromium was in one of the projects that ran. That's separate from whether or not it's actually installed. (I check for that below using `!existsSync(getRuntimePath())`)

Andarist: That's what I tried to say - both are kinda separate problems but the `existsSync` check is within the `! this._foundReplayBrowser` branch.

So after merging this, if I **did** configure the project to run with `replay-chromium` but forgot to actually install the Replay Chromium I end up with this output:
```
  ✘  1 [replay-chromium] › test/example.spec.ts:3:5 › basic test (8ms)


  1) [replay-chromium] › test/example.spec.ts:3:5 › basic test ─────────────────────────────────────

    Error: browserType.launch: Failed to launch chromium because executable doesn't exist at /Users/mateuszburzynskiprev/.replay/runtimes/Replay-Chromium.app/Contents/MacOS/Chromium

  1 failed
    [replay-chromium] › test/example.spec.ts:3:5 › basic test ──────────────────────────────────────
[replay.io]: 🕑 Completing some outstanding work ...
[replay.io]: 
```

bvaughn: I see what you're saying. I didn't change that behavior in this PR, but it was a scenario we weren't calling out before.

bvaughn: I have filed [PRO-751](https://linear.app/replay/issue/PRO-751) to follow up on this and will tackle it once you land #585

PR #585, opened 2024-07-02
Improve the warning about Replay-less executions

Andarist:


packages/playwright/src/reporter.ts
```diff
@@ -327,9 +340,21 @@ class ReplayPlaywrightReporter implements Reporter {
...
+        return executed && !usingReplay;
+      });
+      if (projectsWithoutReplay.length) {
+        const projectText =
+          projectsWithoutReplay[0] === ROOT_PROJECT_NAME
+            ? "Your project"
+            : `${projectsWithoutReplay.join(", ")} project${
+                projectsWithoutReplay.length > 1 ? "s" : ""
+              }`;
+
```
bvaughn, line 355: I'm not sure this produces a meaningfully better warning message to be worth it. I think just saying "Your tests ran without Replay Chromium" would be sufficient.

What would this actually look like in practice? My experience is pretty limited here, but the way I've typically seen projects named were like "firefox" or "chrome" So in that case this would print a warning like:

> chrome project ran without Replay Chromium. If this wasn't intentional, please recheck your configuration.

or

> firefox, chrome projects ran without Replay Chromium. If this wasn't intentional, please recheck your configuration.

Both of those seem kind of awkward to me, particularly if the project has a 1:1 relationship with a browser. I guess in most cases it would be the root project so we'd just say "Your project ran without Replay Chromium" but in that case, I think that saying "Your tests ran without Replay Chromium" might still read a little better?

packages/playwright/src/reporter.ts
```diff
@@ -327,9 +340,21 @@ class ReplayPlaywrightReporter implements Reporter {
   async onEnd() {
     try {
       await this.reporter.onEnd();
-      if (!this._foundReplayBrowser) {
+
+      const projectsWithoutReplay = Object.keys(this._projects).filter(projectName => {
+        const { executed, usingReplay } = this._projects[projectName];
+        return executed && !usingReplay;
+      });
```
bvaughn, line 347: I'm not sure this is the check we want here. Shouldn't we _actually_ be checking to ensure that at least one project using Replay ran? (It's okay if others ran that weren't using it.) e.g.

```js
const shouldWarn = !Object.keys(this._projects).find(projectName => {
  const { executed, usingReplay } = this._projects[projectName];
  return executed && usingReplay;
});
```

I don't think we'd want to show ourselves this warning in the case when we chose to run with both regular Chromium and Replay Chromium, would we? (Maybe you think we would? Maybe you're worried about a case where different projects have different sets of tests they run? But that still seems pretty intentional to me.)

Andarist: When I initially implemented this warning I specifically didn't want to warn when "mixed" projects would be executed... and I just forgot about this intention when working on this. I think you are right here. This also addresses the other concern that you raised

packages/playwright/src/reporter.ts
```diff
@@ -177,14 +177,24 @@ export default class ReplayPlaywrightReporter implements Reporter {
...
+    }
     this.reporter.setTestRunnerVersion(version);
     this.reporter.onTestSuiteBegin();
   }
 
   onTestBegin(test: TestCase, testResult: TestResult) {
+    const projectName = test.parent.project()?.name;
+
+    // it's important to handle the root project's name here and that's an empty string
+    if (typeof projectName === "string") {
```
bvaughn, line 194: This comment and the check below are not totally aligned. An empty string would still be `typeof` "string"

Do you mean to check `if (projectName)` ?

packages/playwright/src/reporter.ts
```diff
@@ -340,22 +350,37 @@ export default class ReplayPlaywrightReporter implements Reporter {
...
+      });
+
+      if (!executedProjectWithReplay) {
+        mixpanelAPI.trackEvent("playwright.warning.reporter-used-without-replay-project");
         output.push(emphasize("None of the configured projects ran using Replay Chromium."));
-        if (!existsSync(getRuntimePath())) {
+      }
+
+      if (!existsSync(getRuntimePath())) {
+        mixpanelAPI.trackEvent("playwright.warning.reporter-used-without-replay-browser");
```
bvaughn, line 367: I think... putting this warning here doesn't make sense, does it? I was planning to follow up on this in PRO-751 but...wouldn't we error much earlier on when we tried to run the first test but there was no browser to run it with?

packages/playwright/src/reporter.ts
Andarist: no, the opposite - I meant that the root project's name is an empty string and thus I have to check `typeof` instead of relying on coercion

bvaughn: Ah, okay. I misinterpreted the intention of the comment.

packages/playwright/src/reporter.ts
Andarist: > wouldn't we error much earlier on when we tried to run the first test but there was no browser to run it with?

Not quite. We can't check what projects are actually executed in the constructor. So we start tracking this information in `onBegin` but we also can't check what projects are actually executed there... so we only gain that knowledge in `onEnd` as at that time we know about all executed tests and related projects.

You are right that all related tests would just fail in such a scenario. It would be a Playwright-generated error though, smth like:
```
  1) [replay-chromium] › test/example.spec.ts:3:5 › basic test ─────────────────────────────────────

    Error: browserType.launch: Failed to launch chromium because executable doesn't exist at /Users/mateuszburzynskiprev/.replay/runtimes/Replay-Chromium.app/Contents/MacOS/Chromium
```

And when that gets logged we can't add a mixpanel event.

bvaughn: Okay. I was thinking we could perhaps detect and warn earlier (in `onTestBegin`) but maybe it's better to print the error at the end anyway.

That being said, this check doesn't look right because it's warning _unconditionally_ (even if Replay project wasn't used)

Andarist: > That being said, this check doesn't look right because it's warning unconditionally (even if Replay project wasn't used)

Hm, yee - you are right. It makes sense to only warn about this on runs when Replay was meant to be executed.

packages/playwright/src/reporter.ts
```diff
@@ -340,22 +350,37 @@ export default class ReplayPlaywrightReporter implements Reporter {
...
+        return executed && usingReplay;
+      });
+
+      if (!executedProjectWithReplay) {
+        mixpanelAPI.trackEvent("playwright.warning.reporter-used-without-replay-project");
         output.push(emphasize("None of the configured projects ran using Replay Chromium."));
-        if (!existsSync(getRuntimePath())) {
+      }
+
+      if (executedProjectWithReplay && !existsSync(getRuntimePath())) {
```
bvaughn, line 366: Would it make more sense for this check to be...
```suggestion
      if (output.length || (executedProjectWithReplay && !existsSync(getRuntimePath()))) {
```

bvaughn: No that's not quite right either.

bvaughn, line 366: Maybe I lead you wrong here. Maybe we should have been?
```suggestion
      if (!existsSync(getRuntimePath())) {
```

packages/playwright/src/reporter.ts
```diff
@@ -340,22 +350,37 @@ export default class ReplayPlaywrightReporter implements Reporter {
...
+        mixpanelAPI.trackEvent("playwright.warning.reporter-used-without-replay-browser");
+        if (output.length) {
           output.push("");
-          output.push(`Install Replay Chromium by running ${highlight("npx replayio install")}`);
         }
+        output.push(`Install Replay Chromium by running ${highlight("npx replayio install")}`);
```
bvaughn, line 371: Should we add a message before this too so it's less floaty. Something that gives context like "To record tests with Replay, you need to install the Replay browser"?

Andarist: Something like that? I like the idea that we keep this in a single line
```suggestion
        output.push(`To record tests with Replay, you need to install the Replay browser ${highlight("npx replayio install")}`);
```

packages/playwright/src/reporter.ts
```diff
@@ -340,22 +350,37 @@ export default class ReplayPlaywrightReporter implements Reporter {
...
+      });
+
+      if (!executedProjectWithReplay) {
+        mixpanelAPI.trackEvent("playwright.warning.reporter-used-without-replay-project");
         output.push(emphasize("None of the configured projects ran using Replay Chromium."));
-        if (!existsSync(getRuntimePath())) {
+      }
+
+      if (executedProjectWithReplay && !existsSync(getRuntimePath())) {
+        mixpanelAPI.trackEvent("playwright.warning.reporter-used-without-replay-browser");
```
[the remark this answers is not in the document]
Andarist: > Also should this be a warning or an error? I guess warning is okay?

It depends. If no projects with Replay were selected then this could be a warning - otherwise an error. I don't feel like complicating it this way right now. It wouldn't be thrown here anyway.

packages/playwright/src/reporter.ts
```diff
@@ -340,22 +350,37 @@ export default class ReplayPlaywrightReporter implements Reporter {
...
+        if (output.length) {
```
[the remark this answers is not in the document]
Andarist: i became meaningful again with the recent change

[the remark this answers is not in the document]
bvaughn: > i became meaningful again with the recent change

So happy for you 🥺

packages/playwright/src/reporter.ts
```diff
@@ -340,22 +350,41 @@ export default class ReplayPlaywrightReporter implements Reporter {
...
+      });
+
+      if (!executedProjectWithReplay) {
+        mixpanelAPI.trackEvent("playwright.warning.reporter-used-without-replay-project");
         output.push(emphasize("None of the configured projects ran using Replay Chromium."));
-        if (!existsSync(getRuntimePath())) {
+      }
+
+      if (!existsSync(getRuntimePath())) {
+        mixpanelAPI.trackEvent("playwright.warning.replay-browser-not-installed");
```
bvaughn, line 367: I think the only thing I would tweak about this maybe is to conditionally report this error (if `executedProjectWithReplay`) so we don't get a bunch of false positives in Mixpanel?

bvaughn (approved):

packages/playwright/src/reporter.ts
Andarist: good idea, i pushed out this change

packages/playwright/src/reporter.ts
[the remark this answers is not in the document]
Andarist: 😂

PR #587, opened 2024-07-02
Initialize Mixpanel events for Cypress reporter

bvaughn:
This will unblock the logger done by test-utils (which previously would not have been sent anywhere, because it waits until Mixpanel has been initialized).

I also logged a couple of Cypress specific events that might be relevant to onboarding. (TBH I'm not sure how our onboarding funnel is going to work and I'm kind of just taking guesses here.)

### DIFF
--- a/packages/playwright/src/reporter.ts
+++ b/packages/playwright/src/reporter.ts
@@ -93,7 +93,7 @@ export default class ReplayPlaywrightReporter implements Reporter {
     });
 
     if (!config || typeof config !== "object") {
-      mixpanelAPI.trackEvent("playwright.error.invalid-reporter-config", { config });
+      mixpanelAPI.trackEvent("error.invalid-reporter-config", { config });
 
       throw new Error(
         `Expected an object for @replayio/playwright/reporter configuration but received: ${config}`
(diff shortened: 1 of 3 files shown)


packages/playwright/src/reporter.ts
```diff
@@ -373,13 +373,13 @@ export default class ReplayPlaywrightReporter implements Reporter {
...
       if (!didUseReplayBrowser) {
-        mixpanelAPI.trackEvent("playwright.warning.reporter-used-without-replay-project");
+        mixpanelAPI.trackEvent("warning.reporter-used-without-replay-project");
         output.push(emphasize("None of the configured projects ran using Replay Chromium."));
       }
 
       if (!isReplayBrowserInstalled) {
         if (didUseReplayBrowser) {
-          mixpanelAPI.trackEvent("playwright.warning.replay-browser-not-installed");
+          mixpanelAPI.trackEvent("warning.replay-browser-not-installed");
```
bvaughn, line 382: This isn't strictly related to the change but I noticed we're prepending an unnecessary "playwright." here. (The events will all already get prefixed with "@replayio/playwright.")

Andarist (approved):

PR #588, opened 2024-07-02
Add assert for eTags and uploadLinks

callingmedic911:
I  can’t think if this would even be possible, but I am adding this assert so it will surface if it somehow happens

Related PRO-715.


bvaughn (approved):
Should we add a similar assert to `replayio` too just to be safe?

callingmedic911:
repalyio checks if each response has etag, that should be good enough I think.

https://github.com/replayio/replay-cli/blob/33656ea4a263e3c09a81b23576f00fce2456a0f3/packages/shared/src/recording/upload/uploadRecording.ts#L274

bvaughn:
Oh cool

PR #589, opened 2024-07-02
migrate remaining logs in `test-utils/*`

miriambudayr:
The title is self-explanatory!

### DIFF
--- a/packages/test-utils/src/config.ts
+++ b/packages/test-utils/src/config.ts
@@ -1,10 +1,11 @@
 import fetch from "node-fetch";
-import dbg from "debug";
 import { warn } from "./logging";
+import { logger } from "@replay-cli/shared/logger";
+import { getErrorMessage } from "./legacy-cli/error";
 
-const debug = dbg("replay:test-utils:config");
+async function queryGraphqlEndpoint(apiKey: string, name: string, query: string, variables = {}) {
+  logger.info("QueryGraphqlEndpoint:Started", { name });
 
-async function query(apiKey: string, name: string, query: string, variables = {}) {
   const options = {
     method: "POST",
     headers: {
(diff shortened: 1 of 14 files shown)


packages/test-utils/src/config.ts
```diff
@@ -1,10 +1,11 @@
 import fetch from "node-fetch";
-import dbg from "debug";
 import { warn } from "./logging";
+import { logger } from "@replay-cli/shared/logger";
```
miriambudayr, line 3: I think I should clean up the `logging.ts` file in another PR.

packages/test-utils/src/legacy-cli/client.ts
```diff
@@ -84,7 +83,8 @@ class ProtocolClient {
     callback?: (err?: Error) => void
   ) {
     const id = this.nextMessageId++;
-    debug("Sending command %s: %o", method, { id, params, sessionId });
+    logger.info("SendCommand:Started", { id, params, sessionId });
```
bvaughn, line 86: Not sure how much we care about this, but since we explicitly mask the access token in a few places– this call site _could_ contain that token (or possibly other sensitive info?) e.g. we send the access token to this when we call the `Authentication.setAccessToken` API.

Maybe our `formatTags` helper could check for a few keys by name (like `apiKey`, `accessToken`, or `token`) and make sure they're masked?

packages/test-utils/src/legacy-cli/client.ts
```diff
@@ -98,7 +98,12 @@ class ProtocolClient {
           this.socket.send(data, callback);
         } else {
           if (err) {
```
bvaughn, line 100: `err` 😭  lol

bvaughn (approved):

packages/test-utils/src/legacy-cli/client.ts
miriambudayr: Fixed it for you!

PR #592, opened 2024-07-03
Improve GitHub API handling

Andarist:


packages/shared/src/cachedFetch.ts
```diff
@@ -1,62 +1,106 @@
...
-      });
+      if (attempt === maxAttempts) {
+        return storeCachedEntry(url, {
+          json,
+          ok: false,
+          status: resp.status,
+          statusText: resp.statusText,
+        });
+      }
+    } catch (err) {
```
bvaughn, line 93: Pet peeve of mine 😅 
```suggestion
    } catch (error) {
```

packages/shared/src/cachedFetch.ts
```diff
@@ -66,3 +110,11 @@ export function resetCache(url?: string) {
     cache.clear();
   }
 }
+
+function storeCachedEntry(url: string, entry: CacheEntry) {
+  const deferred = cache.get(url)?.deferred;
+  assert(deferred, "Deferred should be defined");
```
bvaughn, line 116: nit: The caller could pass in the `Deferred` instance (it has it already) then you could avoid this check.

packages/shared/src/cachedFetch.ts
```diff
@@ -1,62 +1,106 @@
...
+          return storeCachedEntry(url, {
+            json,
+            ok: false,
             status: resp.status,
             statusText: resp.statusText,
           });
-          break;
+        }
+        if (typeof shouldRetry === "object" && typeof shouldRetry.after === "number") {
+          retryAfter = shouldRetry.after;
```
bvaughn, line 81: This seems like a bit of unnecessary complexity IMO, but maybe you had a reason to make this change? :)

packages/shared/src/cachedFetch.ts
```diff
@@ -1,62 +1,106 @@
...
-      const json = await resp.json();
-      cache.set(url, {
-        json,
-        status: resp.status,
-        statusText: resp.statusText,
-      });
-    } else if (attempt < maxAttempts) {
       if (shouldRetryFn) {
-        const shouldRetry = await shouldRetryFn(resp);
+        const shouldRetry = await shouldRetryFn(resp, json, retryAfter);
```
bvaughn, line 71: nit: Previously we only called this function if we still had remaining retries. Now we call it anyway, then potentially bail out immediately. Seems slightly better before? We could probably remove the duplicate return-error by combining these a bit more

```js
let shouldRetry = attempt < maxAttempts;
if (shouldRetry && shouldRetryFn) {
  shouldRetry = await shouldRetryFn(resp, json, retryAfter);
}

// ...
```

packages/shared/src/recording/metadata/legacy/test/index.ts
```diff
@@ -17,7 +17,7 @@ const versions = {
   ...v2,
 };
 
-export function validate(test: UnstructuredMetadata) {
+export function validate(test?: UnstructuredMetadata) {
   if (!test) {
     throw new Error("Test metadata does not exist");
```
bvaughn, line 22: This seems like an odd change. Why make the type optional and then throw if it's not passed? 🤔

packages/shared/src/recording/metadata/sanitizeMetadata.ts
```diff
@@ -36,9 +43,12 @@ export async function sanitizeMetadata(metadata: UnstructuredMetadata) {
           break;
         }
         default: {
-          logger.debug(
-            `Ignoring metadata key "${key}". Custom metadata blocks must be prefixed by "x-". Try "x-${key}" instead.`
-          );
+          if (opts.verbose) {
+            console.log(
```
bvaughn, line 47: Just double checking that you meant to `console.log` here instead of `logger.log`

bvaughn (approved):
Nice cleanup. I left a few suggestions in `cachedFetch` but otherwise looks good

packages/shared/src/cachedFetch.ts
Andarist: I need to control the delay in case I hit GitHub's rate limit. When hitting the secondary rate limit we need to wait like a minute or two (which is quite long... maybe I need to add some log about it too) and when hitting the primary rate limit the response contains the information about the time I need to wait before making the request again. 

I implemented this first using another optional callback but then I ended up checking the rate limit kind in both of those callbacks and it made sense to combine them into one (so `shouldRetry` could return the desired delay directly)

packages/shared/src/cachedFetch.ts
Andarist: I'll cover this with tests, thanks for noticing

packages/shared/src/cachedFetch.ts
bvaughn: Okay. That makes sense then.

packages/shared/src/cachedFetch.ts
bvaughn: That's fair but it's not even really a problem that a test would root out. I'm just saying that we're doing unnecessary work (and there's a bit of unnecessarily duplicated code) :)

packages/shared/src/recording/metadata/legacy/test/index.ts
Andarist: I'm just matching the previous API here to maintain backwards compatibility with this:
https://github.com/replayio/replay-cli/blob/72867d6c6c3f209c58596de1c6cf36d4d0429dea/packages/replay/src/metadata/test.ts

It was a public export. We might not want to publish the old CLI anymore but it's good to have it in a publishable state.

That said, I can look into hoisting this `throw` into that legacy layer.

packages/shared/src/recording/metadata/sanitizeMetadata.ts
Andarist: probably not, I'll recheck this, thanks!

packages/shared/src/recording/metadata/legacy/test/index.ts
Andarist: I think it makes sense to keep it at this layer. Essentially those `validate` functions should really accept `unknown` but I don't plan to change the parameter annotation here ;p

packages/shared/src/recording/metadata/sanitizeMetadata.ts
Andarist: This is intentional because `opts.verbose` is meant to output things into the console. Our new `logger` doesn't ever do that today

packages/shared/src/recording/metadata/legacy/test/index.ts
bvaughn: > Essentially those validate functions should really accept unknown

Yeah, agreed. I was just a bit surprised to see the halfway change :)

packages/shared/src/recording/metadata/sanitizeMetadata.ts
bvaughn: > Our new logger doesn't ever do that today

Seems like maybe we should figure out a way to make _this_ change rather than adding one or two `console.log` statements?

Andarist: Maybe, although their purpose is somewhat different. It's one thing to log things useful *to us* and log things useful to the consumer. They won't always align. Many of our new logs are in tag/label format like:
```ts
logger.debug("LaunchBrowser:Stdout", { text });
```
From what I understand this is a format that should (eventually) be used by all of those logger calls. But with that I'm not sure what we should do when we want to both use the logger and log smth to the user. Something like this?
```ts
logger.info("SanitizeMetadata:IgnoringKey", {
  key
  verbose: verbose && `Ignoring metadata key "${key}". Custom metadata blocks must be prefixed by "x-". Try "x-${key}" instead.`
});
```

I think it's probably super bikesheddable and it could easily be put in a followup task

bvaughn: Oh yeah totally. Sorry I didn't mean to suggest that was a blocker for merging this :)

PR #593, opened 2024-07-03
Add missing fetch imports

Andarist:


packages/replay/src/auth.ts
```diff
@@ -1,3 +1,4 @@
+import fetch from "node-fetch";
```
Andarist, line 1: This has been "broken"* for months from what I can tell but it seemed like a good idea to fix it now (I run an audit through all of the files in the repo to see if `fetch` is imported by all them). I decided to use `node-fetch` (over `undici`) as that's what is used more by this package (the old CLI)

*didn't always work in node 18 or earlier

packages/shared/src/cachedFetch.ts
```diff
@@ -1,3 +1,5 @@
+import { fetch } from "undici";
```
bvaughn, line 1: This seems _very easy_ to forget to do. Can we add a Lint rule to disallow using the global `fetch`? Else I can see us reintroducing this problem in the future.

Andarist: PRO-755 , it should be a super quick ticket but I'm wrapping up for today and I don't have time to switch branches and test it out. I'm also unsure what's the current state of lining in this repo

bvaughn: 🫡

PR #594, opened 2024-07-03
install `stack-utils` to all packages that inline the shared folder

miriambudayr:
This fixes the bug report here: https://discord.com/channels/779097926135054346/798693728999047203/1258136045619384420


package.json
```diff
@@ -41,5 +42,8 @@
     "@jest/environment@npm:^27.5.1": "patch:@jest/environment@npm%3A28.1.3#~/.yarn/patches/@jest-environment-npm-28.1.3-506a81a227.patch",
     "@jest/environment@npm:^28.1.3": "patch:@jest/environment@npm%3A28.1.3#~/.yarn/patches/@jest-environment-npm-28.1.3-506a81a227.patch",
     "superstruct": "patch:superstruct@npm%3A1.0.4#~/.yarn/patches/superstruct-npm-1.0.4-44d328b887.patch"
+  },
+  "dependencies": {
+    "stack-utils": "^2.0.6"
```
bvaughn, line 47: I don't think we need to (or should?) add this tot he root `package.json`. We probably _should_ add it to the one in `packages/replay` though (even if we don't plan any more releases of this project)

bvaughn (approved):
Approve! but with a change suggested

package.json
miriambudayr: Okay

bvaughn (approved):
I think this looks right. If CI is happy we can test it quickly before publishing.

bvaughn:
Don't forget to add a changeset file to this PR so this change will be picked up by all of the packages you've updated.

Once that's done we should be unblocked to merge this, and then merge the "Versions" PR that will be auto-opened.

packages/cypress/package.json
```diff
@@ -57,6 +58,7 @@
     "p-map": "^4.0.0",
     "semver": "^7.5.2",
     "sha-1": "^1.0.0",
+    "stack-utils": "^2.0.6",
```
Andarist, line 61: this should catch such issues early:
https://github.com/replayio/replay-cli/blob/78f5c72a62a38772cad90cccf1283b80eea49b61/scripts/pkg-build/src/plugins/resolveErrors.ts#L29-L35

I wonder why it didn't let us know about it here... I'll investigate tomorrow

PR #596, opened 2024-07-03
Use import instead of require to load `stack-utils`

Andarist:


packages/shared/src/logger.ts
```diff
@@ -4,9 +4,7 @@ import LokiTransport from "winston-loki";
 import { AuthInfo } from "./graphql/fetchAuthInfoFromGraphQL";
 import { getDeviceId } from "./getDeviceId";
 import { randomUUID } from "crypto";
-
-// Does not work with import. Only works with require().
```
Andarist, line 8: I checked types and the generated runtime code and both look correct with `import`. Could you double-check what was the problem? How to reproduce it?

PR #597, opened 2024-07-04
Fixed validation in the `resolveErrors` plugin

Andarist:


packages/replayio/src/utils/async/logAsyncOperation.ts
```diff
@@ -6,8 +6,8 @@ import {
 } from "@replay-cli/shared/async/createDeferred";
 import { statusFailed, statusPending, statusSuccess } from "@replay-cli/shared/theme";
 import { dots } from "cli-spinners";
-import { logUpdate } from "../../../../shared/src/logUpdate";
```
bvaughn, line 9: Should we add a lint rule to guard against `../shared` being part of an import?

packages/shared/src/logger.ts
```diff
@@ -4,9 +4,7 @@ import LokiTransport from "winston-loki";
 import { AuthInfo } from "./graphql/fetchAuthInfoFromGraphQL";
 import { getDeviceId } from "./getDeviceId";
 import { randomUUID } from "crypto";
-
-// Does not work with import. Only works with require().
-const StackUtils = require("stack-utils");
+import StackUtils from "stack-utils";
```
bvaughn, line 7: Hmm. Did you test this? Looking at the `stack-utils` package ... this won't work. (There is no default export. They set `module.exports = StackUtils` (which is the class)

packages/replayio/src/utils/async/logAsyncOperation.ts
Andarist: The `resolveErrors` fails on those correctly now, see the logic related to this:
https://github.com/replayio/replay-cli/pull/597/files#diff-22552c614a4309d01e45261b51401c60c927eafcf4b6e005245f3a611b141292R64

I rechecked auto-imports and they suggest the correct imports for me - so I'm not entirely sure how this was added here. Maybe by using a different VS Code or TS Server version?

It often doesn't have to be a big issue, even when this one went through it was just "virtualized" by Rollup (put into `dist/_virtual`). It's still a problem when that happens though.

But yes, a lint rule would be great as that would give us feedback earlier.

packages/shared/src/logger.ts
Andarist: I did... but this is a great point. I'd expect something to be veeeery wrong already. I think we lucked out because we don't depend on many packages that we import defaults from. Any package that would contain `module.exports.default` would fail now with the current setup.

We can see here what is the output of this code right now ([rollup repl](https://rollupjs.org/repl/?version=4.18.0&shareable=JTdCJTIyZXhhbXBsZSUyMiUzQW51bGwlMkMlMjJtb2R1bGVzJTIyJTNBJTVCJTdCJTIyY29kZSUyMiUzQSUyMmltcG9ydCUyMFN0YWNrVXRpbHMlMjBmcm9tJTIwJ3N0YWNrLXV0aWxzJyU1Q25jb25zb2xlLmxvZyhTdGFja1V0aWxzKSUyMiUyQyUyMmlzRW50cnklMjIlM0F0cnVlJTJDJTIybmFtZSUyMiUzQSUyMm1haW4uanMlMjIlN0QlNUQlMkMlMjJvcHRpb25zJTIyJTNBJTdCJTIyb3V0cHV0JTIyJTNBJTdCJTIyZm9ybWF0JTIyJTNBJTIyY2pzJTIyJTJDJTIyaW50ZXJvcCUyMiUzQSUyMmRlZmF1bHQlMjIlN0QlMkMlMjJ0cmVlc2hha2UlMjIlM0F0cnVlJTdEJTdE)):
```ts
'use strict';

var StackUtils = require('stack-utils');

console.log(StackUtils);
```

This is clearly wrong for the mentioned cases with `.default`. We should be using `interop: 'auto'` to fix this. With that option the output would look like this ([rollup repl](https://rollupjs.org/repl/?version=4.18.0&shareable=JTdCJTIyZXhhbXBsZSUyMiUzQW51bGwlMkMlMjJtb2R1bGVzJTIyJTNBJTVCJTdCJTIyY29kZSUyMiUzQSUyMmltcG9ydCUyMFN0YWNrVXRpbHMlMjBmcm9tJTIwJ3N0YWNrLXV0aWxzJyU1Q25jb25zb2xlLmxvZyhTdGFja1V0aWxzKSUyMiUyQyUyMmlzRW50cnklMjIlM0F0cnVlJTJDJTIybmFtZSUyMiUzQSUyMm1haW4uanMlMjIlN0QlNUQlMkMlMjJvcHRpb25zJTIyJTNBJTdCJTIyb3V0cHV0JTIyJTNBJTdCJTIyZm9ybWF0JTIyJTNBJTIyY2pzJTIyJTJDJTIyaW50ZXJvcCUyMiUzQSUyMmF1dG8lMjIlN0QlMkMlMjJ0cmVlc2hha2UlMjIlM0F0cnVlJTdEJTdE)):
```ts
'use strict';

var StackUtils = require('stack-utils');

function _interopDefault (e) { return e && e.__esModule ? e : { default: e }; }

var StackUtils__default = /*#__PURE__*/_interopDefault(StackUtils);

console.log(StackUtils__default.default);
```

bvaughn: I had also tried this at one point (seeing if an `import` would trigger our build validation error, where `require` did not). It didn't, but `StackUtils` threw an error when instantiated with `new` because the thing that was imported wasn't a class. I'm...surprised this didn't throw for you.

packages/replayio/src/utils/async/logAsyncOperation.ts
bvaughn: My VSCode still kind of struggles with this project's config. I think the "../../shared" type paths get added when you refactor/move a file around and it tries to auto-update the paths for you. (I've tried to catch them and fix them manually but..)

packages/shared/src/logger.ts
Andarist: Now I got scared that I didn't test it properly. I'll double-check before merging.

The reason why it just couldn't throw our validation error was the require usage here - those are not special in vanilla Rollup. It's an ESM bundler and CJS support is only brought via plugins. So it didn't even see that it was being loaded by anything as only actual imports are resolved by default.

But we also had those other issues in this plugin here that prevented this from doing its job even after switching to `import`.

packages/replayio/src/utils/async/logAsyncOperation.ts
Andarist: Could you recheck if you are using TS 5.5? In the command palette u can find "Select TypeScript version..."

bvaughn: Huh. For some reason VS Code wasn't using the workspace version.

Screenshot 2024-07-04 at 5 38 53 PM

Andarist: It never does by default 😢  - they claim it's for security reasons. "Select TypeScript version..." is basically memory muscle for me at this point whenever I open a new repo

bvaughn: Weird bc the same value that works to "fix" this in a project setting doesn't seem to work when I put it in my user settings. And adding it to the project settings for e.g. Devtools modifies a committed VSCode settings file 🤪

Andarist: > And adding it to the project settings for e.g. Devtools modifies a committed VSCode settings file 🤪

Yep, for the long time I committed this file to repos because I believed that it would "fix" this for other contributors... but even that's not enough. They require explicit user consent each time. I never tried to put this in my user settings though.

packages/shared/src/logger.ts
Andarist: I re-confirmed that it works now.

bvaughn: Okay cool. I wonder why it didn't work when I tried it.

PR #601, opened 2024-07-05
Remove unused bits from the test utils

Andarist:


packages/test-utils/src/legacy-cli/main.ts
```diff
@@ -286,16 +281,15 @@ async function directUploadRecording(
 }
 
 async function doUploadRecording(
-  dir: string,
```
bvaughn, line 289: Thanks for verifying that it was okay for us to remove this. I wanted to with the earlier PR but it was just too big already.

bvaughn (approved):

PR #606, opened 2024-07-08
Root out legacy upload APIs from the test-utils reporter

Andarist:
I noticed that we removed the file related to multipart upload from the files that we ported to test-utils. Luckily, we also forgot to initialize LD... so those code paths were never hit here (cc @callingmedic911 ). 

This sparked a further refactor to get rid of the legacy uploading code from the test-utils and to use the one that we have in `/shared`.


.changeset/silver-geckos-smoke.md
```diff
@@ -0,0 +1,6 @@
+---
+"@replayio/playwright": patch
+"replayio": patch
+---
+
+Stuff
```
Andarist, line 6:
```suggestion
```

packages/playwright/src/reporter.ts
```diff
@@ -85,13 +86,17 @@ export default class ReplayPlaywrightReporter implements Reporter {
...
     initLogger(packageName, packageVersion);
     mixpanelAPI.initialize({
-      accessToken: getAccessToken(config),
+      accessToken,
       packageName,
       packageVersion,
     });
-
+    if (accessToken) {
+      initLaunchDarklyFromAccessToken(accessToken);
```
Andarist, line 98: This is async but we can't even await things here, this is race'y. But what should be done about it - should `getFeatureFlagValue` await the pending identification?

.changeset/silver-geckos-smoke.md
bvaughn: Looks like you beat me to this one 😉

packages/playwright/src/reporter.ts
bvaughn: Hmm... `initLaunchDarklyFromAccessToken` is async because it fetches auth info.

The reporter base class (in `test-utils`) already awaits auth info in order to initialize the Grafana logger. Should we move LD initialization in there?

>  should `getFeatureFlagValue` await the pending identification?

It already does :) That's how I designed it.

packages/replayio/src/utils/recordings/uploadRecordings.ts
```diff
@@ -0,0 +1,218 @@
...
+  link,
+  statusFailed,
+  statusPending,
+  statusSuccess,
+} from "@replay-cli/shared/theme";
+import { dots } from "cli-spinners";
+import assert from "node:assert/strict";
+import strip from "strip-ansi";
+
+async function printDeferredRecordingActions(
```
bvaughn, line 30: I don't understand why these methods got moved _back into_ `replayio` from shared.

.changeset/silver-geckos-smoke.md
Andarist: I had to temporarily add it to use `/release-pr` since the set of packages it publishes is derived from existing changesets. I really need to add some `--include-changed` or `--all` to the relevant Changesets command :v

packages/playwright/src/reporter.ts
Andarist: > Hmm... initLaunchDarklyFromAccessToken is async because it fetches auth info.

The same is true about MixPanel's initialization - both should get the same treatment here. 

> The reporter base class (in test-utils) already awaits auth info in order to initialize the Grafana logger. Should we move LD initialization in there?

I'll look into this.

> It already does :) That's how I designed it.

It waits on the LD's initialization - but is this the same as the user's identification? 🤔

bvaughn: Perhaps I misremembered what it's doing 😆 You're right. I think waiting some small amount (before falling back to the default user profiles) seems reasonable?

Andarist:
> RE: the changes to initLaunchDarklyFromAccessToken – I'm going to re-think this stuff as part of PRO-797

Ye, I reverted the LD-related changes from this PR as it wasn't completely obvious how it should work. I'll lean on PRO-797 to improve this

PR #609, opened 2024-07-08
Use a unified DEBUG namespace

Andarist:
In the past, we could use `DEBUG=replay:*` or `DEBUG=replayio:*`. None of those work now because we ditched "nested namespaces". The new logger configured the `localDebugger` once (or well, twice - once in the constructor + it gives us the ability to override it in the `initialize` method) and it only uses a single namespace without any colons.

It means that now we should use `DEBUG=replayio`. Since all packages configure the `localDebugger` to use their package names as the namespace... we even have to use verbose names like `DEBUG=@replayio/playwright`. I think we should settle on a single namespace - let it be either `replay` or `replayio`

### DIFF
--- a/packages/shared/src/logger.ts
+++ b/packages/shared/src/logger.ts
@@ -94,9 +94,6 @@ class Logger {
   }
 
   private log(message: string, level: LogLevel, tags?: Tags) {
-    if (!this.initialized) {
-      throw new Error("Logger not initialized. Call `logger.initialize` first.");
-    }
     const formattedTags = this.formatTags(tags);
 
     this.localDebugger(message, formattedTags);
--- a/packages/shared/src/logger.ts
+++ b/packages/shared/src/logger.ts
@@ -51,13 +51,14 @@ class Logger {
     this.sessionId = randomUUID();
   }
 
+  // This should be called with the name once at the entry point.
+  // For example, with the Playwright plugin, it is called in the Reporter interface constructor.
   initialize(app: string, version: string | undefined) {
     if (this.initialized) {
       console.warn(`Logger already initialized.`);
     }
 
     this.initialized = true;
-    this.localDebugger = dbg(app);
     this.grafana = this.initGrafana(app, version);
   }
 
@@ -93,6 +94,9 @@ class Logger {
   }
 
   private log(message: string, level: LogLevel, tags?: Tags) {
+    if (!this.initialized) {
+      throw new Error("Logger not initialized. Call `logger.initialize` first.");
+    }
     const formattedTags = this.formatTags(tags);
 
     this.localDebugger(message, formattedTags);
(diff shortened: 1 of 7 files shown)


packages/shared/src/logger.ts
```diff
@@ -93,6 +94,9 @@ class Logger {
   }
 
   private log(message: string, level: LogLevel, tags?: Tags) {
+    if (!this.initialized) {
+      throw new Error("Logger not initialized. Call `logger.initialize` first.");
+    }
```
Andarist, line 99: This won't play nicely with the Playwright fixture. We can't quite initialize the logger there because we have no place to close it in that context (unless we'd be initializing and closing the logger for each test separately). So I'm torn on what should be done about it.

packages/shared/src/logger.ts
```diff
@@ -51,13 +51,14 @@ class Logger {
     this.sessionId = randomUUID();
   }
 
+  // This should be called with the name once at the entry point.
+  // For example, with the Playwright plugin, it is called in the Reporter interface constructor.
   initialize(app: string, version: string | undefined) {
```
Andarist, line 56: this has to be called somewhere in the old CLI too - let's settle on the other details first though

packages/shared/src/logger.ts
bvaughn: Should we do what we do elsewhere and queue such logs up until the logger has been initialized?

Andarist: We don't have a good communication channel between the reporter and the reporter (which has access to the logger). Currently, we use WS to communicate them... but I'm just literally working on replacing that with Playwright's annotations/attachments. I don't think we should be keeping our logs there though.

bvaughn: > We don't have a good communication channel between the reporter and the reporter

🤔 Huh? 😅 lol

callingmedic911: Reporter and fixture?

bvaughn: I see.

So this change means...if anything in the fixture tries to log, it will throw? Does it throw already? (is there already logging in that context?)

Andarist: > So this change means...if anything in the fixture tries to log, it will throw?

Yes, for that reason it's probably better to revert this part of the PR.

> Does it throw already?

No

> is there already logging in that context?

Just local debug-based one under the `replay` namespace (the default one). We don't have grafana logs in this context

bvaughn: Ok 👍🏼

packages/shared/src/logger.ts
```diff
@@ -51,13 +51,14 @@ class Logger {
...
     if (this.initialized) {
       console.warn(`Logger already initialized.`);
     }
 
     this.initialized = true;
-    this.localDebugger = dbg(app);
```
Andarist, line 60: Before landing this PR I just need to know what our DEBUG namespace should be called. Do we settle on `replayio` for everything? or `replay`?

bvaughn: "replayio" is our CLI package name so IMO it would be a little confusing to use that for everything

bvaughn (approved):
I guess if we're standardizing here, let's standardize on "replay"

PR #610, opened 2024-07-08
add Sentry API

miriambudayr:
This adds Sentry to the repository in a similar manner as we have done with the logger and mixpanel.

I instrumented the Playwright reporter as well so we can start getting errors in production.

We will still need to use Sentry in other packages (`test-utils`, `cypress`, etc.) and this lays a foundation for that work.

### DIFF
--- a/packages/cypress/package.json
+++ b/packages/cypress/package.json
@@ -47,6 +47,8 @@
     "typescript": "^5.5.2"
   },
   "dependencies": {
+    "@sentry/node": "^8.13.0",
+    "@sentry/profiling-node": "^8.13.0",
     "chalk": "^4.1.2",
     "debug": "^4.3.4",
     "fs-extra": "^11.2.0",
--- a/packages/playwright/src/reporter.ts
+++ b/packages/playwright/src/reporter.ts
@@ -6,6 +6,7 @@ import type {
   TestResult,
 } from "@playwright/test/reporter";
 import { initLogger, logger } from "@replay-cli/shared/logger";
+import { initSentry, sentry, withSentry, withSentrySync } from "@replay-cli/shared/sentry";
 import { mixpanelAPI } from "@replay-cli/shared/mixpanel/mixpanelAPI";
 import { getRuntimePath } from "@replay-cli/shared/runtime/getRuntimePath";
 import { emphasize, highlight, link } from "@replay-cli/shared/theme";
@@ -86,6 +87,8 @@ export default class ReplayPlaywrightReporter implements Reporter {
     setUserAgent(`${packageName}/${packageVersion}`);
 
     initLogger(packageName, packageVersion);
+    initSentry(packageName, packageVersion);
+
     mixpanelAPI.initialize({
       accessToken: getAccessToken(config),
       packageName,
@@ -143,14 +146,16 @@ export default class ReplayPlaywrightReporter implements Reporter {
   }
 
   getFixtureData(test: TestExecutionIdData) {
-    const id = this._getTestExecutionId(test);
-    this.fixtureData[id] ??= {
-      steps: [],
-      stacks: {},
-      filenames: new Set(),
-    };
+    return withSentrySync("GetFixtureData", () => {
+      const id = this._getTestExecutionId(test);
+      this.fixtureData[id] ??= {
+        steps: [],
+        stacks: {},
+        filenames: new Set(),
+      };
 
-    return this.fixtureData[id];
+      return this.fixtureData[id];
+    });
   }
 
   // Playwright already provides a unique test id:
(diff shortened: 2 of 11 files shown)


miriambudayr:
> 1.14 MB

That's a lot of dependencies. What size increase normally raises red flags for you? @Andarist @callingmedic911 @bvaughn

packages/playwright/src/reporter.ts
```diff
@@ -159,289 +164,311 @@ export default class ReplayPlaywrightReporter implements Reporter {
...
   private _getTestExecutionId(test: TestExecutionIdData) {
-    return [
-      test.filePath,
-      test.projectName ?? "",
-      test.repeatEachIndex,
-      test.attempt,
-      ...test.source.scope,
-      test.source.title,
-    ].join("-");
+    return withSentrySync("GetTestExecutionId", () => {
```
bvaughn, line 167: I don't think this makes sense> There's no function to wrap here, just an inline array.

packages/playwright/src/reporter.ts
```diff
@@ -159,289 +164,311 @@ export default class ReplayPlaywrightReporter implements Reporter {
...
+      ].join("-");
+    });
   }
 
   getSource(test: TestCase) {
-    return {
-      title: test.title,
-      scope: test.titlePath().slice(3, -1),
-    };
+    return withSentrySync("GetSource", () => {
```
bvaughn, line 180: I don't think this makes sense either (same reason as above)

packages/shared/src/logger.ts
```diff
@@ -30,9 +30,29 @@ function anonymizeStackTrace(stack: string): string {
     .join("\n");
 }
 
-type LogLevel = "error" | "warn" | "info" | "debug";
+export function formatTags(tags?: Tags) {
```
bvaughn, line 33: If this is going to be used by multiple things (logger aka Grafana _and_ Sentry) maybe it makes sense to put it in its own standalone module? (So the scope is clearer?)

packages/shared/src/sentry.ts
```diff
@@ -0,0 +1,126 @@
...
+  try {
+    return fn();
+  } catch (error: any) {
+    logger.error(`${methodName}:Failed`, error);
+    sentry.captureException(error);
+    throw error;
+  }
+}
+
+export { initSentry, withSentry, withSentrySync };
```
bvaughn, line 126: Small nit: I prefer exports that are inlined (e.g. `export function foo() {}`) because I think it's more readable. (When I'm looking at a function, I can tell if it's "private" or exported without having to grep elsewhere in the file.

packages/shared/src/sentry.ts
```diff
@@ -0,0 +1,126 @@
...
+  initialize(app: string, version: string | undefined) {
+    if (this.initialized) {
+      console.warn(`Logger already initialized.`);
+    }
+
+    this.initSentry(app, version);
+    this.initialized = true;
+  }
+
+  private initSentry(app: string, version: string | undefined) {
```
bvaughn, line 31: nit: This function could be inlined into `initialize` since it's not doing anything particularly complex.

packages/shared/src/sentry.ts
```diff
@@ -0,0 +1,126 @@
...
+    await Sentry.close();
+  }
+
+  async flush() {
+    if (process.env.REPLAY_TELEMETRY_DISABLED) {
+      return;
+    }
+
+    await Sentry.flush();
+  }
```
bvaughn, line 61: Should we wrap/expose flush? Seems like we only want `Sentry.close` to be used so maybe it's clearer to only expose that.

bvaughn (approved):
Wow, Yarn lock files are so chatty.

packages/shared/src/sentry.ts
miriambudayr: Yeah, we don't need this. I ended up not using it.

packages/shared/src/sentry.ts
miriambudayr: Thanks, that's a good recommendation. I agree that making the scope and visibility of functions, methods, and classes explicit at the point of declaration is more readable.

miriambudayr:
> Wow, Yarn lock files are so chatty.

I agree. It surprised me too.

packages/cypress/package.json
```diff
@@ -47,6 +47,8 @@
     "typescript": "^5.5.2"
   },
   "dependencies": {
+    "@sentry/node": "^8.13.0",
+    "@sentry/profiling-node": "^8.13.0",
```
Andarist, line 51: ideally, we'd be using the same version range everywhere but right now we are using a mix of `^8.13.0` and `^8.15.0` in this PR

packages/shared/src/sentry.ts
```diff
@@ -0,0 +1,114 @@
...
+    });
+
+    this.initialized = true;
+  }
+
+  identify(authInfo: AuthInfo) {
+    this.authInfo = authInfo;
+  }
+
+  async close() {
```
Andarist, line 45: we should add a call to this to `exitTasks` here:
https://github.com/replayio/replay-cli/blob/207cf051bcbc008da3290b23ae2a466d44f56f45/packages/shared/src/process/exitTask.ts#L7

packages/playwright/src/reporter.ts
```diff
@@ -143,14 +146,16 @@ export default class ReplayPlaywrightReporter implements Reporter {
...
   getFixtureData(test: TestExecutionIdData) {
-    const id = this._getTestExecutionId(test);
-    this.fixtureData[id] ??= {
-      steps: [],
-      stacks: {},
-      filenames: new Set(),
-    };
-
-    return this.fixtureData[id];
+    return withSentrySync("GetFixtureData", () => {
```
Andarist, line 149: it could be nice to have some `withSentryTracking` that could wrap a whole class automatically

packages/shared/src/sentry.ts
```diff
@@ -0,0 +1,114 @@
...
+
+    Sentry.setTags({
+      app,
+      version,
+    });
+
+    this.initialized = true;
+  }
+
+  identify(authInfo: AuthInfo) {
```
Andarist, line 41: this method stays unused - should we be using it somewhere?

packages/shared/src/sentry.ts
```diff
@@ -0,0 +1,114 @@
...
+export async function withSentry<T>(
+  methodName: string,
+  fn: () => T | Promise<T>,
+  tags?: Tags
+): Promise<T> {
+  try {
+    const result = await fn();
+    return result;
+  } catch (error: any) {
+    logger.error(`${methodName}Failed`, tags);
```
Andarist, line 100: I feel like this one might be too hidden here. It lacks symmetry with `*:Started` logs and similar. When reading code sites that use this I'd be tempted to add `logger.error` to *them* because I wouldn't realize that this is already auto-handled within this function.

We could add `*:Started` logs here but I think it would still be surprising. It doesn't like `withSentry` should step outside of its core responsibility (integrating with Sentry).

packages/shared/src/sentry.ts
```diff
@@ -0,0 +1,114 @@
...
+  }
+}
+
+export const sentry = new SentryAPI();
+
+// This should be called with the name once at the entry point.
+// For example, with the Playwright plugin, it is called in the Reporter interface constructor.
+export function initSentry(app: string, version: string | undefined) {
+  sentry.initialize(app, version);
+}
```
Andarist, line 89: I feel like this is redundant. All files importing `initSentry` end up using `sentry` too so they end up with both `initSentry` and `sentry` and the first one could simply be replaced at call sites with `sentry.initialize`

packages/shared/src/sentry.ts
```diff
@@ -0,0 +1,114 @@
...
+
+class SentryAPI {
+  private deviceId: string;
+  private sessionId: string;
+  private authInfo?: AuthInfo;
+  private initialized: boolean = false;
+
+  constructor() {
+    this.deviceId = getDeviceId();
+    this.sessionId = randomUUID();
```
Andarist, line 20: It would be great to share this between SentryAPI and the logger, this way we could try to correlate both data sources more easily

packages/shared/src/sentry.ts
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  initialize(app: string, version: string | undefined) {
+    if (this.initialized) {
+      console.warn(`Logger already initialized.`);
```
Andarist, line 25:
```suggestion
      console.warn(`Sentry already initialized.`);
```

packages/shared/src/sentry.ts
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    Sentry.init({
+      dsn: SENTRY_DSN,
+      integrations: [nodeProfilingIntegration()],
+    });
```
Andarist, line 31: I wonder if this alone is enough to send *something* to Sentry. If yes then perhaps we should not even init this module when `process.env.REPLAY_TELEMETRY_DISABLED` is truthy

packages/shared/src/sentry.ts
```diff
@@ -0,0 +1,114 @@
...
+
+// This should be called with the name once at the entry point.
+// For example, with the Playwright plugin, it is called in the Reporter interface constructor.
+export function initSentry(app: string, version: string | undefined) {
+  sentry.initialize(app, version);
+}
+
+export async function withSentry<T>(
+  methodName: string,
+  fn: () => T | Promise<T>,
```
Andarist, line 93: We have the other helper to track sync executions. I think it would make sense to draw a hard line between both of them:
```suggestion
  fn: () => Promise<T>,
```

Andarist: or we could just have a single `withSentry`:
```ts
export function withSentry<T>(fn: () => T) {
  try {
    const result = fn();
    if (
      !!result &&
      typeof result === "object" &&
      "then" in result &&
      typeof result.then === "function"
    ) {
      return result.then(
        (value: any) => value,
        (error: any) => {
          sentry.captureException(error);
          throw error;
        },
      );
    }
    return result;
  } catch (error: any) {
    sentry.captureException(error);
    throw error;
  }
}
```

PR #612, opened 2024-07-09
Rethrow installation errors

Andarist:


packages/replayio/src/utils/installation/installLatestRelease.ts
```diff
@@ -84,10 +84,10 @@ export const installLatestRelease = createAsyncFunctionWithTracking(
       };
     } catch (error) {
       logger.error("InstallLatestRelease:Failed", { error });
-
       progress.setFailed(
         "Something went wrong installing the Replay browser. Please try again later."
       );
+      throw error;
```
Andarist, line 90: This also makes the error to be logged to the console by the outer catch handler. I think it's desirable because it gives us first insights into the problem when the user reports it

bvaughn (approved):

PR #586, opened 2024-07-02
Only save test data for tests run with Replay browser

bvaughn:
Only upload test metadata for tests that were run with the Replay browser. If a given test run has no tests that match this criteria, then we should not even upload any metadata for it.

## Testing
I've tested the following scenarios...

### Upload behavior
<details><summary>Only upload tests run with Replay browser</summary>

```js
// playwright.config.js
reporter: [
  ["dot"] as const,
  replayReporter({
    apiKey: process.env.REPLAY_API_KEY,
    upload: true,
  })
],
projects: [
  {
    name: "chromium",
    testMatch: "**/*login-logout-devtools.spec.ts",
    use: { ...devices["Desktop Chrome"] },
  },
  {
    testMatch: "**/*login-logout-library.spec.ts",
    use: { ...replayDevices["Replay Chromium"] },
  },
],
```

I was testing a few things with this configuration:
- We do upload recordings but only for tests that were run by the Replay browser
- We don't even upload metadata for tests that were run with other browsers

Here is the terminal output from the above test run:

```
[replay.io]: 🕑 Completing some outstanding work ...
[replay.io]: 
[replay.io]: 🚀 Successfully uploaded 1 recordings:
[replay.io]: 
[replay.io]:    ✅ login-logout-library: should login and logout from the library
[replay.io]:       https://app.replay.io/recording/8dc7d529-d7ff-4a0d-8cb4-7177b21b33df
```

And [here](https://app.replay.io/team/dzpkN2MzMDk5YS1mYzI3LTRiZjctYWU1ZC0wN2RhMTVhNzY1MjE=/runs?testRunId=de278eb2-a2e2-4a0f-a9fa-c81c7daee945&testId=) are the results in our dashboard which seem to confirm that it worked as expected:
image

</details> 

<details><summary>Upload all tests if all are run with Replay browser</summary>

```js
// playwright.config.js
reporter: [
  ["dot"] as const,
  replayReporter({
    apiKey: process.env.REPLAY_API_KEY,
    upload: true,
  })
],
projects: [
  {
    name: "chromium",
    use: { ...devices["Desktop Chrome"] },
  },
  {
    name: "replay-chromium",
    use: { ...replayDevices["Replay Chromium"] },
  },
],
```

Terminal output:
```
····
  4 passed (13.7s)

[replay.io]: 🕑 Completing some outstanding work ...
[replay.io]: 
[replay.io]: 🚀 Successfully uploaded 2 recordings:
[replay.io]: 
[...]

PR #613, opened 2024-07-09
Add a trailing log line in case another playwright reporter removes our last line

hbenl:


packages/playwright/src/reporter.ts
```diff
@@ -403,6 +403,8 @@ export default class ReplayPlaywrightReporter implements Reporter {
           console.warn(`[replay.io]: ${line}`);
         });
       }
+
+      console.log("[replay.io]:");
```
Andarist, line 407: This looks really odd on its own. Maybe we could add some text to this, like this?
```suggestion
      console.log("[replay.io]: Completed");
```

Andarist, line 407:
```suggestion
      // we need to output an extra line that is safe to be deleted
      // Playwright's line reporter removes the last line here:
      // https://github.com/microsoft/playwright/blob/0c11d6ed803db582a5508c70f89e55dc9a36c751/packages/playwright/src/reporters/line.ts#L91
      //
      // so if the user configured their reporters like this:
      //
      // reporters: [replayReporter({ upload: true }), ['line']]
      // 
      // that can lead to removing *our* last log line 
      console.log("[replay.io]:");
```

hbenl: I don't think it looks odd since we already output some (almost) empty lines like this. I see this line as padding.

packages/playwright/src/reporter.ts
```diff
@@ -403,6 +403,17 @@ export default class ReplayPlaywrightReporter implements Reporter {
...
+
+      // we need to output an extra line that is safe to be deleted
+      // Playwright's line reporter removes the last line here:
+      // https://github.com/microsoft/playwright/blob/0c11d6ed803db582a5508c70f89e55dc9a36c751/packages/playwright/src/reporters/line.ts#L91
+      //
+      // so if the user configured their reporters like this:
+      //
+      // reporters: [replayReporter({ upload: true }), ['line']]
+      // 
+      // that can lead to removing *our* last log line
```
Andarist, line 415:
```suggestion
      // that can lead to removing *our* last log line 
      //
      // the issue is tracked here: https://github.com/microsoft/playwright/issues/23875
```

packages/playwright/src/reporter.ts
Andarist: This padding always bothered me - I intuitively expected to find some extra text below it. I think Brian had similar sentiment since he removed this recently [here](https://github.com/replayio/replay-cli/pull/586#discussion_r1662981854).

Perhaps it's the fact that this line is not completely empty (it has our prefix) makes is weird to me 🤷‍♂️

Andarist (approved):

packages/playwright/src/reporter.ts
bvaughn: I'd prefer if this were just be an empty line :) but it's at least better this way than for Playwright to trim it off.

PR #617, opened 2024-07-09
Gracefully handle missing recordings.log entries

bvaughn:


packages/shared/src/recording/getRecordings.ts
```diff
@@ -273,3 +326,13 @@ export function getRecordings(processGroupIdFilter?: string): LocalRecording[] {
...
 }
+
+function logNotFoundWarning(recordingId: string, entries: LogEntry[], entry: LogEntry) {
+  logger.error(`Recording with ID "${recordingId}" not found`, {
+    recordingId,
+    entry,
+    entries: JSON.stringify(
+      entries.filter(entry => entry.id === recordingId || entry.recordingId === recordingId)
+    ),
+  });
```
Andarist, line 337: we'd often log this multiple times which can be a little bit wasteful, maybe it would be better to aggregate those orphaned entries and log per recording id groups at the end?

Andarist (approved):

packages/shared/src/recording/getRecordings.ts
bvaughn: Yeah I'd briefly considered that and decided maybe it wasn't worth adding _state_ to improve the debug logging but lemme sit with the idea a bit more.

Andarist: i'm fine with keeping it this way but then perhaps we shouldn't log filtered `entries` here - they would just be scattered (but not duplicated) through those relevant logs

packages/shared/src/recording/getRecordings.ts
```diff
@@ -273,3 +326,13 @@ export function getRecordings(processGroupIdFilter?: string): LocalRecording[] {
       .sort((a, b) => b.date.getTime() - a.date.getTime())
   );
 }
+
+function logNotFoundWarning(recordingId: string, entries: LogEntry[], entry: LogEntry) {
+  logger.error(`Recording with ID "${recordingId}" not found`, {
```
Andarist, line 331: the recently added logs are in a format like this:
```suggestion
  logger.error(`RecordingLog:RecordingNotFound`, {
```

PR #618, opened 2024-07-09
Call `logger.identify` properly

callingmedic911:
- Moves the logic for getting authInfo from `test-utils/reporter` directly inside th elogger.
- Moves `logger.identify` directly intro the entrypoints of Cypress and Playwright.
- Adds `logger.identify` call into the `replayio`'s `initialize` function.

Limitations
- The Cypress and Playwright entrypoints are not async; therefore, this identify call isn't awaited, so first few logs would be missing `workspaceId` / `userId`.
-  There are a few logs before the `identify` call in `replayio` (for example: log inside `getAccessToken`) that would be missing `workspaceId` / `userId`. I'm not sure where else to put this async call.

### DIFF
--- a/packages/replayio/src/utils/initialization/initialize.ts
+++ b/packages/replayio/src/utils/initialization/initialize.ts
@@ -9,6 +9,7 @@ import { checkForRuntimeUpdate } from "./checkForRuntimeUpdate";
 import { promptForAuthentication } from "./promptForAuthentication";
 import { promptForNpmUpdate } from "./promptForNpmUpdate";
 import { promptForRuntimeUpdate } from "./promptForRuntimeUpdate";
+import { logger } from "@replay-cli/shared/logger";
 
 export async function initialize({
   checkForNpmUpdate: shouldCheckForNpmUpdate,
@@ -42,6 +43,8 @@ export async function initialize({
     npmUpdateCheck = { hasUpdate: undefined },
   ] = await promises;
 
+  await logger.identify(accessToken);
+
   if (requireAuthentication && !accessToken) {
     accessToken = await promptForAuthentication();
   }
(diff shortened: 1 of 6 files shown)


packages/replayio/src/utils/initialization/initialize.ts
```diff
@@ -51,6 +54,8 @@ export async function initialize({
   // They also shouldn't block on failure, so we should only wait a couple of seconds before giving up
   const abortController = new AbortController();
 
+  const loggerPromise = raceWithTimeout(logger.identify(accessToken), 2_500, abortController);
```
bvaughn, line 57: I think this presents a subtle race. If the timeout logic is _external_ to the logger, then it might finish after we "gave up" externally, which seems to defeat the purpose.

I think this same problem exists for our Launch Darkly initialization though so it's not like this is a _new_ thing. I'll try to give it some thought as part of PRO-797.

packages/replayio/src/utils/initialization/initialize.ts
```diff
@@ -42,6 +43,8 @@ export async function initialize({
     npmUpdateCheck = { hasUpdate: undefined },
   ] = await promises;
 
+  await logger.identify(accessToken);
+
```
bvaughn, line 47:
```suggestion
```

bvaughn (approved):

packages/replayio/src/utils/initialization/initialize.ts
bvaughn: Once this extra line is removed, I think this PR can be merged and I'll follow up with PRO-797

callingmedic911: Whoops, I left it accidentally.

PR #619, opened 2024-07-10
Recommend installing `xz` if it's missing

Andarist:


packages/replayio/src/utils/installation/installLatestRelease.ts
```diff
@@ -150,18 +149,27 @@ async function downloadReplayFile({ onRetry }: { onRetry?: (attempt: number) =>
...
     logger.error("ExtractBrowserArchive:Failed", {
       downloadFilePath,
+      hasXz,
       stderr: String(tarResult.stderr),
     });
 
-    throw new Error("Unable to extract browser archive");
+    let message = "Unable to extract browser archive";
+
+    if (!hasXz) {
```
Andarist, line 169: it's tempting to suggest some concrete installation instructions for this but it might differ based on the linux distro used and other factors

packages/replayio/src/utils/installation/installLatestRelease.ts
```diff
@@ -150,18 +149,27 @@ async function downloadReplayFile({ onRetry }: { onRetry?: (attempt: number) =>
...
+      message += `. xz is required to decompress it. Please install xz and try again.`;
```
bvaughn, line 170: I wonder if this should say "XZ Utils" or something more than just "xy" ?

bvaughn (approved):
Love it.

packages/replayio/src/utils/installation/installLatestRelease.ts
Andarist: this is tricky because in some distros the proper thing to install is `xz`... and they say that JS dependency management is a mess 🤷‍♂️

PR #621, opened 2024-07-10
Update ts-configs to not reference the deleted "replay" package

bvaughn:
Why didn't #616 catch this?

### DIFF
--- a/packages/cypress/tsconfig.json
+++ b/packages/cypress/tsconfig.json
@@ -8,9 +8,6 @@
   "include": ["src/**/*.ts"],
   "references": [
     {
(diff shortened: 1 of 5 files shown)


bvaughn:
> That said, we don't run tsc at all in the CI... that's a problem. I'll fix this in a moment

Yup! Definitely an oversight.

I simultaneously wish our "_Build, lint and test_" task did more _and_ less (it takes **so long**)

Andarist:
I wish it wouldn't use Earthly 🤣

bvaughn:
Yeah big same

PR #624, opened 2024-07-10
Refactored async initialization logic for Mixpanel, Grafana, and Launch Darkly

bvaughn:
- Consolidate initialization logic for MixPanel, LaunchDarkly, Grafana, and the user agent string.
- More consistent handling of tasks that should await authentication (and be flushed before exit)
- Minor code reorganization to (hopefully) improve readability/clarity

As part of this tidying up, I also changed the login command to show whoami information after a successful sign-in:

Screenshot 2024-07-15 at 11 07 13 AM

Or if you try to log in when you're already authenticated via API key:

Screenshot 2024-07-15 at 11 07 39 AM


bvaughn:
These auto-generated changeset file names.. 🤔 
Screenshot 2024-07-10 at 3 57 50 PM

Andarist:
The computers have spoken.

packages/shared/src/session/initializeSession.ts
```diff
@@ -0,0 +1,27 @@
...
+}: {
+  accessToken: string | undefined;
+  packageName: string;
+  packageVersion: string;
+}) {
+  if (deferredPackageInfo.status === STATUS_RESOLVED) {
+    return;
+  }
+
+  deferredPackageInfo.resolve({ packageName, packageVersion });
```
Andarist, line 19: do we have to make this thing async? This one feels like something that should be available synchronously. But I suppose that:
1. it could complicate the code as we'd have to split `initializeSession` into 2 calls, one that would prefill the package info and the other one that would initialize it with the `accessToken` lazily
2. it might not matter since we need that user agent to make requests... and those usually need to await for the `accessToken` anyway

packages/test-utils/src/getAccessToken.ts
```diff
@@ -1,5 +1,5 @@
 import { ReplayReporterConfig } from "./types";
 
-export function getAccessToken(config: ReplayReporterConfig<any>) {
-  return config.apiKey || process.env.REPLAY_API_KEY || process.env.RECORD_REPLAY_API_KEY;
+export function getAccessToken(config?: ReplayReporterConfig<any>) {
```
Andarist, line 3: It could be good to include here similar logs as we have in `packages/shared/src/authentication/getAccessToken.ts`

packages/shared/src/session/deferred.ts
```diff
@@ -0,0 +1,6 @@
+import { createDeferred } from "../async/createDeferred";
+import { AuthInfo } from "../authentication/types";
+import { PackageInfo } from "./types";
+
+export const deferredAuthInfo = createDeferred<AuthInfo | null>();
```
Andarist, line 5: This is slightly incorrect. It should be this
```suggestion
export const deferredAuthInfo = createDeferred<AuthInfo | undefined>();
```
That's because `packages/shared/src/session/initializeSession.ts` can resolve it with `undefined` and not `null`.

This has accidentally been allowed because `deferred.resolve` has an optional argument. We could make optionally optional (when the `Type` allows `undefined`) but I think it's not worth the added complexity like this:
```ts
type Foo<T> = {
  method: (...args: undefined extends T ? [a?: T] : [a: T]) => void;
};

declare const foo1: Foo<{ bar: string } | null>;
// @ts-expect-error
foo1.method();
// @ts-expect-error
foo1.method(undefined);

declare const foo2: Foo<{ bar: string } | undefined>;
foo2.method();
foo2.method(undefined);
```

So I'd go with a patch like this:
```diff
diff --git a/packages/shared/src/async/createDeferred.ts b/packages/shared/src/async/createDeferred.ts
index 8558bd50..3fffb90e 100644
--- a/packages/shared/src/async/createDeferred.ts
+++ b/packages/shared/src/async/createDeferred.ts
@@ -11,10 +11,10 @@ export interface Deferred<Type, Data = undefined> {
   debugLabel: string | undefined;
   promise: Promise<Type>;
   rejection: Error | undefined;
-  reject(error: Error): void;
-  rejectIfPending(error: Error): void;
-  resolve(value?: Type): void;
-  resolveIfPending(value?: Type): void;
+  reject: (error: Error) => void;
+  rejectIfPending: (error: Error) => void;
+  resolve: (value: Type) => void;
+  resolveIfPending: (value: Type) => void;
   resolution: Type | undefined;
   status: StatusPending | StatusRejected | StatusResolved;
 }
```

packages/shared/src/session/waitForAuthInfoWithTimeout.ts
```diff
@@ -0,0 +1,11 @@
+import { raceWithTimeout } from "../async/raceWithTimeout";
+import { WAIT_FOR_AUTH_INFO_TIMEOUT } from "./config";
+import { deferredAuthInfo } from "./deferred";
+
+const promise = raceWithTimeout(deferredAuthInfo.promise, WAIT_FOR_AUTH_INFO_TIMEOUT);
```
Andarist, line 5: This starts the race at the module initialization time and not when `initializeSession` gets called. This is probably fine in practice today but I feel like it's a little bit less "correct"

packages/cypress/src/index.ts
```diff
@@ -261,11 +261,7 @@ const plugin = (
   config: Cypress.PluginConfigOptions,
   options: PluginOptions = {}
 ) => {
-  setUserAgent(`${packageName}/${packageVersion}`);
-
-  logger.initialize(packageName, packageVersion);
-  logger.identify(getAuthKey(config));
-  mixpanelAPI.initialize({
```
Andarist, line 268: 😍

packages/shared/src/mixpanelClient.test.ts
```diff
@@ -0,0 +1,513 @@
...
+    });
+  });
+
+  describe("trackAsyncEvent", () => {
+    let createDeferred: typeof import("./async/createDeferred").createDeferred;
+
+    beforeEach(async () => {
+      createDeferred = require("./async/createDeferred").createDeferred;
+
+      await await initializeSession({
```
Andarist, line 380:
```suggestion
      await initializeSession({
```

packages/shared/src/session/initializeSession.ts
```diff
@@ -0,0 +1,27 @@
...
+
+  let authInfo: AuthInfo | undefined = undefined;
+  if (accessToken) {
+    authInfo = await getAuthInfo(accessToken);
```
Andarist, line 23: this can throw - what should be the behavior when that happens? a lot of the callers to `initializeSession` create an orphaned promise and it might take the whole pricess down in unexpected ways

packages/puppeteer/src/install.ts
```diff
@@ -1,13 +1,16 @@
...
+import { initializeSession } from "@replay-cli/shared/session/initializeSession";
+import { getAccessToken } from "@replayio/test-utils";
+import { name as packageName, version as packageVersion } from "../package.json";
 
 export default async function install() {
   try {
-    logger.initialize(name, version);
-    const accessToken = await getAccessToken();
-    await logger.identify(accessToken.accessToken);
+    initializeSession({
```
Andarist, line 9:
```suggestion
    await initializeSession({
```

Andarist: or this should be unwrapped from this try/catch

packages/shared/src/session/deferred.ts
Andarist: Hm, we actually often call `deferred.resolve()` so to avoid problems we might want to use the complicated version outlined above 😢

packages/shared/src/async/isPromiseLike.ts
Andarist, line 1: do we need this one? I don't see it used anywhere

packages/shared/src/session/waitForAuthInfoWithTimeout.ts
Andarist: After a second thought... maybe that's intended so we don't wait too long before flushing logger queues. I'm not sure though

packages/shared/src/process/exitProcess.ts
```diff
@@ -1,7 +1,7 @@
-import { exitTasks } from "./exitTask";
+import { waitForExitTasks } from "./waitForExitTasks";
 
 export async function exitProcess(code?: number): Promise<never> {
-  await Promise.all(exitTasks.map(task => task()));
+  await waitForExitTasks;
```
Andarist, line 4:
```suggestion
  await waitForExitTasks();
```

packages/shared/src/process/waitForExitTasks.ts
Andarist, line 1: nit: those 4 modules are so small and related that maybe they could be combined into a single module?

packages/shared/src/launchDarklylient.ts
```diff
@@ -0,0 +1,56 @@
...
+  LDClient,
+  LDContext,
+  LDUser,
+  initialize as initializeLDClient,
+} from "launchdarkly-node-client-sdk";
+import { AuthInfo } from "./authentication/types";
+import { getReplayPath } from "./getReplayPath";
+import { AuthenticatedTaskQueue } from "./session/AuthenticatedTaskQueue";
+
+class LaunchDarklyClient extends AuthenticatedTaskQueue {
```
Andarist, line 11: I know it's pretty subjective so feel free to ignore this. I feel like this structure with subclassing doesn't give us a lot but it makes the control flow a little bit harder to reason about. It's also, in general, less flexible. To use `AuthenticatedTaskQueue` one has to implement the consumer as a class. I probably would refactor this to a class that anything could instantiate and introduce `queue` as a private property to all of the existing consumers. I tinkered with this locally and it felt better to me but YMMV.

Andarist (approved):

packages/shared/src/session/waitForAuthInfoWithTimeout.ts
bvaughn: I considered this and decided this was okay/intended (and also probably not too impactful since we'll likely call the wait-for method very near the start of the session anyway) but I could be convinced to change it.

I also considered if each caller should get their own start time or if there should be a global timeout and I decided the latter was preferable.

packages/shared/src/mixpanelClient.test.ts
bvaughn: Oh gosh. I got a bit out of control there didn't I.

Andarist: Better to be safe than sorry! :P

packages/shared/src/session/initializeSession.ts
bvaughn: The reason I made this async from an external POV is that I didn't want to assume execution order, particularly for things like our loggers/metrics stuff– which initialize themselves during module parsing (e.g. `new client()`)

This likely happens before the packages call `initializeSession` and so would cause problems.

packages/shared/src/session/deferred.ts
bvaughn: I'm not totally sure I understand what you're proposing in the "complicated option" above, but if it's okay I'll fix the incorrect type (`createDeferred<AuthInfo | null>()` -> `createDeferred<AuthInfo | undefined>()`) and we can address the deferred API signatures in a follow up.

packages/shared/src/session/initializeSession.ts
bvaughn: Good catch. I think for now– log an error (but swallow it)

packages/puppeteer/src/install.ts
bvaughn: Ugh. Good catch.

packages/shared/src/launchDarklylient.ts
bvaughn: That's fair. I think I went with a class because 2 of our 3 authenticated _things_ were already classes (following the `logger` pattern, for consistency) but lemme play with this for a few and see.

Andarist: I cam up with this when I played around with it: https://github.com/replayio/replay-cli/commit/a57b91d667c0280dce6adf72902884082af56ee5 . I didn't run tests or anything but types were OK with it IIRC

bvaughn: Thanks for sharing the patch, but I've already been working on a rewrite of this locally so I'm going to just follow through with it. :)

bvaughn: Update: 9c59dbc

Specifically:
https://github.com/replayio/replay-cli/blob/9c59dbce9584b3ce148ce9e44a88047c6453c5ee/packages/shared/src/session/createTaskQueue.ts

Even just looking at the tests, I am convinced that this is a nicer coding mechanism:
https://github.com/replayio/replay-cli/blob/9c59dbce9584b3ce148ce9e44a88047c6453c5ee/packages/shared/src/session/createTaskQueue.test.ts

packages/replayio/src/utils/initialization/initialize.ts
```diff
@@ -47,34 +43,11 @@ export async function initialize({
     accessToken = await promptForAuthentication();
```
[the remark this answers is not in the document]
bvaughn: Agreed. This is still the part of this that I'm thinking about the most.

.yarn/releases/yarn-4.1.1.cjs
Andarist, line 1: i'm surprised by this in the diff - perhaps it got reformatted or smth?

bvaughn: Yeah this is a weird one. I just noticed it too.

packages/replayio/src/utils/initialization/initialize.ts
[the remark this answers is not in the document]
bvaughn: One idea that occurs to me is that this task can pause/resume the timeout logic.

That would mean that events would queue behind a successful or failed auth attempt, which seems okay because they'd be flushed before close (even in the failed auth attempt).

The case that would fall between the cracks, I think, would be a permanently hung auth attempt (neither success nor failure) which we could maybe handle by adding an explicit timeout to the `promptForAuthentication` method too... thoughts?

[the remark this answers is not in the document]
bvaughn: Another idea (that I would need to think about more) is that we could get rid of the timeout part and just queue all logs behind auth _except_ if `flushAndClose` is called (pre-exit) in which case we could send everything then.

That would not cover the case where a user force-quits before auth-ing but that's maybe not super common?

Andarist (approved):

bvaughn:
Eh fuck it.

PR #626, opened 2024-07-11
Add `upload.maximumLimit`

Andarist:


packages/test-utils/src/reporter.ts
```diff
@@ -364,6 +365,12 @@ export default class ReplayReporter<
       if (typeof config.upload === "object") {
         this._minimizeUploads = !!config.upload.minimizeUploads;
         this._uploadStatusThreshold = config.upload.statusThreshold ?? "all";
+        if (typeof config.upload.maximumLimit === "number") {
+          if (!Number.isFinite(config.upload.maximumLimit) || config.upload.maximumLimit <= 0) {
+            throw new Error("`upload.maximumLimit` must be a finite number higher than 0");
```
bvaughn, line 370: nit: "_a finite number higher than 0_" is technically correct but kind of stuffy. Maybe something more like?
```suggestion
            throw new Error("`upload.maximumLimit` must be greater than 0; to disable uploads entirely, set `upload` to false");
```

packages/test-utils/src/reporter.ts
```diff
@@ -763,6 +770,9 @@ export default class ReplayReporter<
       });
       return;
     }
+    if (this._uploadedRecordings.size >= this._uploadMaxLimit) {
```
bvaughn, line 773: I don't see us accounting for the number of shards here. That seems important?

PR #639, opened 2025-07-30
add playwright-ct support

BLamy:


packages/playwright-ct/src/fixture.ts
```diff
@@ -0,0 +1,615 @@
...
+
+  const clientInstrumentation = playwright._instrumentation;
+  clientInstrumentation.addListener(csiListener);
+
+  // Enhanced mount function that tracks component lifecycle
+  const enhancedMount = async (component: any, options?: any) => {
+    console.log(
+      "[replay.io CT]: Enhanced mount called for component:",
+      component?.type?.name || component?.type || "Component"
+    );
```
Andarist: I think console.logs dont belong to this fixture. It should work transparently and shouldn't clobber up console with potentially unwanted console logs

packages/playwright-ct/src/fixture.ts
```diff
@@ -0,0 +1,615 @@
...
+    // Track mount start
+    const mountStepId = uuid();
+    testInfo.attach(`replay:step:start`, {
+      body: JSON.stringify({
+        event: "step:start",
+        id: mountStepId,
+        test: testData,
+        apiName: "mount",
+        category: "pw:api",
+        title: `mount: ${componentName}`,
```
Andarist: this feels like a hack (it's not a `pw:api` per-se ;p) - how do we use those attachments? do we need them?

PR #655, opened 2026-04-15
PRO-2387 forward REPLAY_CLIENT_SOURCE env as request header

michaelfward:
when `REPLAY_CLIENT_SOURCE` is set in the environment, forward it as the `X-Replay-Source` header on outbound graphql and recording upload requests. lets the backend attribute traffic from inside orchestrated environments (e.g. the autobuilder container) to the driving context.

no-op when the env var is unset.

## related

- replayio/backend: reads `x-replay-source` as `replay.client.context` telemetry attribute
- replayio/app-building: sets the env var in the autobuilder container

### DIFF
--- a/.changeset/client-attribution.md
+++ b/.changeset/client-attribution.md
@@ -1,0 +1,9 @@
+---
+"replayio": patch
+"@replayio/playwright": patch
+"@replayio/cypress": patch
+"@replayio/jest": patch
+"@replayio/puppeteer": patch
+---
+
+add client attribution headers for telemetry. sends X-Client-Info on outbound HTTP requests to enable backend source tracking. forwards REPLAY_CLIENT_SOURCE env var as X-Replay-Source header when set by orchestrated environments.
--- a/.changeset/client-attribution.md
+++ b/.changeset/client-attribution.md
@@ -6,4 +6,4 @@
 "@replayio/puppeteer": patch
 ---
 
-add client attribution headers for telemetry. sends X-Client-Info on outbound HTTP requests to enable backend source tracking. forwards REPLAY_CLIENT_SOURCE env var as X-Replay-Source header when set by orchestrated environments.
+internal: add a way to attribute the client origin of the outbound traffic to Replay services
--- a/packages/shared/src/graphql/queryGraphQL.ts
+++ b/packages/shared/src/graphql/queryGraphQL.ts
@@ -23,9 +23,7 @@ export async function queryGraphQL(name: string, query: string, variables = {}, 
     options.headers.Authorization = `Bearer ${apiKey.trim()}`;
   }
 
-  if (process.env.REPLAY_CLIENT_SOURCE) {
-    options.headers["X-Replay-Source"] = process.env.REPLAY_CLIENT_SOURCE;
-  }
+  options.headers["X-Replay-Source"] = process.env.REPLAY_CLIENT_SOURCE || "cli";
 
   logDebug("Querying graphql endpoint", { name, replayApiServer });
   const result = await fetch(`${replayApiServer}/v1/graphql`, options);
--- a/packages/shared/src/graphql/queryGraphQL.ts
+++ b/packages/shared/src/graphql/queryGraphQL.ts
@@ -23,6 +23,10 @@ export async function queryGraphQL(name: string, query: string, variables = {}, 
     options.headers.Authorization = `Bearer ${apiKey.trim()}`;
   }
 
+  if (process.env.REPLAY_CLIENT_SOURCE) {
+    options.headers["X-Replay-Source"] = process.env.REPLAY_CLIENT_SOURCE;
+  }
+
   logDebug("Querying graphql endpoint", { name, replayApiServer });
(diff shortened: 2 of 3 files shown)


packages/shared/src/recording/upload/uploadRecording.ts
```diff
@@ -479,6 +482,9 @@ async function uploadRecordingReadStream(
           "Content-Length": size.toString(),
           "User-Agent": userAgent,
           Connection: "keep-alive",
+          ...(process.env.REPLAY_CLIENT_SOURCE
```
Andarist, line 489: We should double-check if all outbound requests receive this new header. At the very least, it's missing in `uploadRecordingWithoutPresignedUrls`

packages/shared/src/graphql/queryGraphQL.ts
```diff
@@ -23,6 +23,10 @@ export async function queryGraphQL(name: string, query: string, variables = {},
     options.headers.Authorization = `Bearer ${apiKey.trim()}`;
   }
 
+  if (process.env.REPLAY_CLIENT_SOURCE) {
```
Andarist: where the value for this is meant to come from in the context of the CLI and our test reporters?

markerikson (approved):

packages/shared/src/recording/upload/uploadRecording.ts
michaelfward: good catch. added the header to `authHeaders` in `uploadRecordingWithoutPresignedUrls` which covers all 4 fetch calls in that function (create-recording + partial-upload chunks). the presigned URL path (line 480) was already covered.

packages/shared/src/graphql/queryGraphQL.ts
michaelfward: it's set by orchestrated environments that drive the CLI from inside a container - e.g. the autobuilder (`app-building` repo) sets `ENV REPLAY_CLIENT_SOURCE=autobuilder` in its Dockerfile. the backend reads it as `replay.client.context` in telemetry so we can distinguish autobuilder-driven traffic from direct user traffic. defaults to `cli` when the env var is unset, so all replay-cli traffic is always tagged.

packages/shared/src/graphql/queryGraphQL.ts
```diff
@@ -23,6 +23,8 @@ export async function queryGraphQL(name: string, query: string, variables = {},
     options.headers.Authorization = `Bearer ${apiKey.trim()}`;
   }
 
+  options.headers["X-Replay-Source"] = process.env.REPLAY_CLIENT_SOURCE || "cli";
```
Andarist: I'm afraid a default like this isn't enough if we want to attribute this precisely. Those functions are used by different callers - in this repo the default might be a CLI but a test reporter (like `@replayio/playwright`) isn't exactly a CLI, it's a specific thing. I feel that caller should pass down a specific value for this as its own default.

Andarist: We have some old tracking code here: https://github.com/replayio/replay-cli/blob/ef5c1151839ef6ecac91efc3a62c452c15f7570a/packages/shared/src/logger.ts

It kinda serves a different purpose but it looks like u could maybe you could rely on `initializePackageInfo`+`waitForPackageInfo` to get the proper default value for this?

michaelfward: good call. switched to using `waitForPackageInfo()` so the default is the actual package name (`replayio`, `@replayio/playwright`, `@replayio/cypress`, etc.) instead of hardcoded `cli`. env var override still takes precedence when set.

packages/shared/src/graphql/queryGraphQL.ts
```diff
@@ -2,15 +2,18 @@ import { fetch } from "undici";
 import { replayApiServer } from "../config";
 import { logDebug } from "../logger";
 import { getUserAgent } from "../session/getUserAgent";
+import { waitForPackageInfo } from "../session/waitForPackageInfo";
 
 export async function queryGraphQL(name: string, query: string, variables = {}, apiKey?: string) {
   const userAgent = await getUserAgent();
+  const { packageName } = await waitForPackageInfo();
```
Andarist: nit: if u'd inline this then in the case of `REPLAY_CLIENT_SOURCE` we wouldn't create a redundant/throwaway promise

Andarist (approved):

.changeset/client-attribution.md
```diff
@@ -0,0 +1,9 @@
+---
+"replayio": patch
+"@replayio/playwright": patch
+"@replayio/cypress": patch
+"@replayio/jest": patch
+"@replayio/puppeteer": patch
+---
+
+add client attribution headers for telemetry. sends X-Client-Info on outbound HTTP requests to enable backend source tracking. forwards REPLAY_CLIENT_SOURCE env var as X-Replay-Source header when set by orchestrated environments.
```
Andarist:
```suggestion
internal: add a way to attribute the client origin of the outbound traffic to Replay services
```

